PR #3, opened 2017-03-29
Minor c++ improvements

axxel:
For details see commit messages.


core/src/qrcode/QRDecoder.cpp
```diff
@@ -55,7 +55,7 @@ CorrectErrors(ByteArray& codewordBytes, int numDataCodewords)
 	// First read into an array of ints
```
huycn: The lines from 54 to 59 could be simplified as:
```
std::vector<int> codewordsInts(codewordBytes.begin(), codewordBytes.end());
```

axxel: indeed :)

PR #4, opened 2017-04-01
Minor c++ improvements (part 2)

axxel:
A couple more improvements that are still mostly style / code quality changes.

### DIFF
--- a/core/src/BitMatrix.cpp
+++ b/core/src/BitMatrix.cpp
@@ -17,6 +17,7 @@
 
 #include "BitMatrix.h"
 #include "BitArray.h"
+#include "BitHacks.h"
 
 #include <stdexcept>
 
@@ -96,15 +97,7 @@ BitMatrix::setRegion(int left, int top, int width, int height)
 void
 BitMatrix::rotate180()
 {
-	BitArray topRow(_width), bottomRow(_width);
-	for (int i = 0; i < (_height + 1) / 2; i++) {
-		getRow(i, topRow);
-		getRow(_height - 1 - i, bottomRow);
-		topRow.reverse();
-		bottomRow.reverse();
-		setRow(i, bottomRow);
-		setRow(_height - 1 - i, topRow);
-	}
+	BitHacks::Reverse(_bits, _rowSize * 32 - _width);
 }
 
 void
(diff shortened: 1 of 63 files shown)


axxel:
Note: I just updated the pull request after I noticed that I broke the ENCODERS build, which I simply ignored up until now. Now the whole library compiles under clang/osx. I might add a couple more minor improvements suggested by clang-tidy.

core/src/BitMatrix.cpp
```diff
@@ -96,15 +97,7 @@ BitMatrix::setRegion(int left, int top, int width, int height)
...
-	BitArray topRow(_width), bottomRow(_width);
-	for (int i = 0; i < (_height + 1) / 2; i++) {
-		getRow(i, topRow);
-		getRow(_height - 1 - i, bottomRow);
-		topRow.reverse();
-		bottomRow.reverse();
-		setRow(i, bottomRow);
-		setRow(_height - 1 - i, topRow);
-	}
+	BitHacks::Reverse(_bits, _rowSize * 32 - _width);
```
huycn: It won't yield the same result if _rowSize * 32 != _width, will it? Since there is padding in each row, thus the shifting must be done for each of them.

core/src/BitMatrix.cpp
```diff
@@ -23,16 +23,6 @@
 namespace ZXing {
 
 void
-BitMatrix::init(int width, int height)
```
axxel: My reasoning for removal was that a) it is kind of a 'non-standard' interface for a simple vector/buffer class, b) it is not present in the Java code and c) it should be necessary pretty seldom, especially in situations where the function is logically 'creating' the buffer instead of modifying it. I think a functional approach without the `&outParam` pattern is generally more readable if applicable. I was thinking about suggesting something along the lines of a mix between `std::optional` and `DecodeStatus` as a return type to make the interfaces more natural (like the Java upstream) without reintroducing exceptions.

But you are right that my patch caused a lot more mallocs to happen in this particular example. Please see the added patch fixing this.

core/src/BitMatrix.cpp
axxel: It does indeed work. E.g.:

> 1010....0011....

becomes

> ....1100....0101

if reversed and then needs to be shifted by 4 places, becomes:

> 1100....0101....

So the gaps between the rows just need to be shifted as well.

axxel:
I forgot to say: thanks for taking the time and looking into my PR ;).

huycn:
You are welcome. That's us who must thank you for your contribution!

core/src/BitMatrix.cpp
huycn: You are right indeed!

PR #5, opened 2017-04-10
BitArray::Iterator and RowReader::FindPattern

axxel:
Expand the existing BitArray::Iterator idea and its usage across the
OneD Readers. Details see commit message.

Note: while looking at the possibilities to simplify the code by using std algorithms I noticed the fact that the code uses your BitArray::Iterator only halfheartedly, meaning it was using both int indices and iterators in parallel. One thing led to the next and in the end I spend a lot more time on this than I probably should have ;). I hope you find this PR useful. Let me know if you find issues, I’ll gladly look into them.


core/src/oned/ODCode128Reader.cpp
```diff
@@ -51,63 +51,31 @@ static const int CODE_START_B = 104;
...
+			    }
+		    }
+		    // Look for whitespace before start pattern, >= 50% of width of start pattern
+		    return bestVariance < MAX_AVG_VARIANCE && row.hasQuiteZone(begin, -(end - begin) / 2);
+	    });
 }
 
-static DecodeStatus
-DecodeCode(const BitArray& row, std::vector<int>& counters, int rowOffset, int& outCode)
+static bool DecodeCode(std::vector<int>& counters, int* outCode)
```
huycn: Same here, assert() is preferred.

core/src/oned/ODITFReader.cpp
```diff
@@ -70,8 +70,7 @@ static const std::array<std::array<int, 5>, 10> PATTERNS = {
 * @return The decoded digit
 * @throws NotFoundException if digit cannot be decoded
 */
-static DecodeStatus
-DecodeDigit(const std::array<int, 5>& counters, int& outCode)
+static bool DecodeDigit(const std::array<int, 5>& counters, int* outCode)
```
huycn: Here also, assert() please.

core/src/oned/ODITFReader.cpp
```diff
@@ -96,26 +95,22 @@ DecodeDigit(const std::array<int, 5>& counters, int& outCode)
 * @param resultString {@link StringBuilder} to append decoded chars to
 * @throws NotFoundException if decoding could not complete successfully
 */
-static DecodeStatus DecodeMiddle(const BitArray& row, int payloadStart, int payloadEnd, std::string& resultString)
+static bool DecodeMiddle(BitArray::Iterator begin, BitArray::Iterator end, std::string* resultString)
```
huycn: And this last one, assert() please.

axxel:
Glad you like it (hence, the work was not a waste of time :)). I added 3 asserts and changed the string result function to remove the pointer altogether in that case.

PR #7, opened 2017-04-14
Temporary heap allocation cleanup

axxel:
Lower the number of mallocs for temporary storage.

### DIFF
--- a/core/src/GenericGF.h
+++ b/core/src/GenericGF.h
@@ -17,6 +17,8 @@
 */
 
 #include "GenericGFPoly.h"
+
+#include <cassert>
 #include <vector>
 #include <stdexcept>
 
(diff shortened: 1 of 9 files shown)


core/src/GenericGF.h
```diff
@@ -45,33 +47,46 @@ class GenericGF
...
+	int addOrSubtract(int a, int b) const {
 		return a ^ b;
 	}
 
 	/**
 	* @return 2 to the power of a in GF(size)
 	*/
 	int exp(int a) const {
-		return _expTable.at(a);
+		return _expTable[a];
```
huycn: I think it'd be better to leave it checked as we don't check the bounds at calling site.

axxel:
First off: thanks for the positive feedback. (You might remember one of my first e-mails to you where I said something about trying to make sure we have a somewhat 'compatible' approach on developing software and collaboration. I explicitly did this because of the rather bad experience I had with the rawspeed project. Your constructive feedback is pretty much the opposite of what was 'standard' there. :-) )

I'm completely happy with dropping the thread_local optimization (will add a commit). But I'd still like to respond to your statements:

> First, thread_local doesn't have much benefit in multi-thread environment if threads are not reused

That is not true, decoding a qrcode image (in one thread) allocates the 2-3 temporary variables multiple times, as you saw with your own allocation-counter test. If you'd decode another image after that you'd get even more savings and only those would not be gained if the implementation would choose a different thread then.

> Second, in case of unsupported, you substitute it with static which would create race conditions in multi-thread setups.

Yes, that is why I said so in the code comment in ZXConfig.h :).

> Also, since these are only temporary variables, I feel it's wrong to make them live longer than the scope where they are created and used.

Yes, the memory will be 'wasted' until the program dies. But we are only talking about a couple hundred bytes.

> That will be harder for code maintenance.

In what way? (Please only answer if you feel like having a really not important conversation about a technical detail ;)).

> The result is 3222 locations

That is interesting. My numbers (the ca. 1300) are from valgrind. So it seems operator new does not always call `malloc()`, or your STL implementation behaves differently compared to my Apple-clang one.

> (which is awesome!)

Thanks.

core/src/GenericGF.h
axxel: I removed it mostly because it seemed inconsistent to check the `_expTable` but no the `_logTable` access. A quick performance comparison showed that the extra check is irrelevant -> I reverted this line.

huycn:
To answer your question on what make thread_local in this case harder for code maintenance:
Working (especially debugging) with code that don't keep and reuse states is way easier that code that do. Static methods are expected to be stateless (caching is another story), which is not the case here. Granted that the usage here only for optimization and will not change code's behavior, but it's not that obvious for a new person to pick up the code (e.g. for one thing: theses variables are assigned via function arguments, not with operator= ).

PR #9, opened 2017-04-18
Implemented a platform independent test runner

axxel:
Note: it uses imagemagick’s convert command line tool to read png/jpg
files.


huycn:
Thanks! that will be very useful. Later on, I will probably extract the common part and put in common place. But it's good for now.

PR #14, opened 2017-10-05
Minor c++ improvements

axxel:
improve coding style / make it more modern, mostly based on the output of clang-tidy. Also fix 2 copy'n'paste bugs that were real semantic issues (see first an sixths commit).


huycn:
Hi Axel, welcome back :)
Good catch for 3 serious bugs. The rest are also good improvements.
Thanks!

PR #18, opened 2017-10-23
Bit array iterator part 2/N

axxel:
Continued with the work on switching from index based BitArray lookup to iterator based access in EAN/UPC 1D decoders.


huycn:
Adding names for thing like std::array<int..> makes code look cleaner. Also, returning ranges when looking for patterns make code look more natural than before. Much appreciated.

PR #19, opened 2017-10-26
Function call signatures

axxel:
The major goal of these changes was to bring back the function call signatures closer to upstream and make them (arguably) more readable and natural by replacing the pattern
```
ResultType r;
auto status = SomeFunc(..., r);
if (StatusIsError(status))
    handleError();
```
with
```
ResultType r = SomeFunc(...);
if (!r.isValid())
    handleError();
```
Different strategies were used:
 * augment the ReslutType with a `bool isValid()` possible by the presence of an internal 'null state'
 * add a `DecoderState` member to ResultType (like already present in `ZXing::Result`)
 * switch to `bool` if only one error code was possible (like `NotFound`).

Also constructors for types transporting results (like `DecoderResult`, `Result`) were modified to support the most natural pattern
```
return ResultType(std::move(text), std::move(bytes), ...);
```
instead of
```
ResultType r;
r.setText(text);
r.setBytes(bytes);
...
return r;
```
Which also saves on memcopies and better shows that they are supposed to consume all the different pieces of result information and send it back to the caller.

Furthermore: all the ReedSolomonErrors turned out to be all mapped to the ChecksumError, so they have been removed, simplifying the code.

Those simplifications added up to 200+ lines of code removed.

Note: a little error counting during executing the test runner revealed the following numbers:
```
NotFound : 4.165.351
FormatError : 115.698
ChecksumError : 1.507
```
So, reporting 'NotFound' as an exception might really be a bad idea but using exceptions for reporting Format/Checksum errors might actually be not that bad.

### DIFF
--- a/core/src/DecoderResult.h
+++ b/core/src/DecoderResult.h
@@ -17,6 +17,7 @@
 */
 
 #include "ByteArray.h"
+#include "DecodeStatus.h"
 
 #include <memory>
 #include <list>
@@ -35,6 +36,7 @@ class CustomData;
 */
 class DecoderResult
 {
+	DecodeStatus _status = DecodeStatus::NoError;
 	ByteArray _rawBytes;
 	int _numBits = 0;
 	std::wstring _text;
@@ -50,9 +52,15 @@ public:
 	//explicit DecoderResult(DecodeStatus status);
 	//DecoderResult(const ByteArray& rawBytes, const String& text, std::list<ByteArray>& byteSegments, const std::string& ecLevel, int saSequence, int saParity);
 	//DecoderResult(const ByteArray& rawBytes, const String& text, std::list<ByteArray>& byteSegments, const std::string& ecLevel);
-	DecoderResult() {}
+	DecoderResult(DecodeStatus status) : _status(status) {}
+	DecoderResult() = default;
 	DecoderResult(const DecoderResult &) = delete;
 	DecoderResult& operator=(const DecoderResult &) = delete;
+	DecoderResult(DecoderResult&&) = default;
+	DecoderResult& operator=(DecoderResult&&) = default;
+
+	bool isValid() const { return StatusIsOK(_status); }
+	DecodeStatus errorCode() const { return _status; }
 
 	const ByteArray& rawBytes() const { return _rawBytes; }
 	void setRawBytes(const ByteArray& bytes) { _rawBytes = bytes; _numBits = 8 * bytes.length(); }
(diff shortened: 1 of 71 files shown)


core/src/Result.h
```diff
@@ -91,13 +132,13 @@ class Result
...
 private:
-	DecodeStatus _status;
+	DecodeStatus _status = DecodeStatus::NoError;
 	std::wstring _text;
 	ByteArray _rawBytes;
 	int _numBits = 0;
 	std::vector<ResultPoint> _resultPoints;
 	BarcodeFormat _format = BarcodeFormat::FORMAT_COUNT;
-	time_point _timestamp;
+	time_point _timestamp = std::chrono::steady_clock::now();
```
huycn: We may also remove completely this member, since it plays no role here

core/src/Result.h
```diff
@@ -40,9 +41,43 @@ class Result
...
+	Result(std::wstring&& text, ByteArray&& rawBytes, std::vector<ResultPoint>&& resultPoints, BarcodeFormat format)
+	    : _text(std::move(text)),
+	      _rawBytes(std::move(rawBytes)),
+		  _resultPoints(std::move(resultPoints)),
+	      _format(format)
+	{
+		_numBits = static_cast<int>(_rawBytes.size()) * 8;
+	}
+
+	Result(DecoderResult&& decodeResult, std::vector<ResultPoint>&& resultPoints, BarcodeFormat format)
```
huycn: Can we move this one and the other above into .cpp, it the .h will be less messy?

core/src/DecoderResult.h
```diff
@@ -46,44 +48,54 @@ class DecoderResult
...
+	std::wstring&& text() && { return std::move(_text); }
+
+	// Simple macro to set up getter/setter methods that save lots of boilerplate.
+	// It sets up a standard 'const & () const', 2 setters for setting lvalues via
+	// copy and 2 for setting rvalues via move. They are provided each to work
+	// either on lvalues (normal 'void (...)') or on rvalues (returning '*this' as
+	// rvalue). The latter can be used to optionally initialize a temporary in a
+	// return statement, e.g.
+	//    return DecoderResult(bytes, text).setEcLevel(level);
+#define PROPERTY(TYPE, GETTER, SETTER) \
```
huycn: OK, but we can name the macro ZX_PROPERTY instead which will greatly reduce the chance of name clashing.

axxel:
Awesome, thanks. I'll implement the improvements you suggested and also fix the new unittest project, which I broke with those changes :-(. I only noticed now when I tried to build it the first time.

PR #25, opened 2017-11-16
DM Encoder fixes and more readability improvements

axxel:
Found another two bugs in the DMHighLevelEncoder. One introduced during the c++ port, the other one likely still in upstream.

Introduced Inflate/Deflate functions for the common use case of scaling/subsampling a bit matrix (duplicate code removal).

Also removed lots of verbose/duplicated code in the DM bit placement / bit parsing code. It seems those two sides of the same coin (generating the DM bit pattern during encoding and parsing it during decoding) have been implemented independently and shared lots of the same patterns.

### DIFF
--- a/core/src/BitMatrix.h
+++ b/core/src/BitMatrix.h
@@ -22,6 +22,7 @@
 namespace ZXing {
 
 class BitArray;
+class ByteMatrix;
 
 /**
 * <p>Represents a 2D matrix of bits. In function arguments below, and throughout the common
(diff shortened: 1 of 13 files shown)


core/src/BitMatrix.h
```diff
@@ -206,4 +214,26 @@ class BitMatrix
...
 
+/**
+ * @brief Inflate scales a BitMatrix up and adds a quite Zone plus padding
+ * @param matrix input to be expanded
+ * @param width new width in bits (pixel)
+ * @param height new height in bits (pixel)
+ * @param quietZone size of quite zone to add in modules
+ * @return expanded BitMatrix, maybe move(input) if size did not change
+ */
+BitMatrix Inflate(BitMatrix&& input, int width, int height, int quietZone);
```
huycn: We may want to use `const BitMatrix& input` instead for the sake of coherence (see Deflate)? Accept only rvalue makes it look like the method is written to be called from particular site.

axxel: My reasoning was performance: in case the requested size of the matrix is less than that of the input (like the 'I don't care'-value 0) the input can simply be passed on. If that would be a `const &` then it would need to be copied no matter what.

The performance impact is admittedly completely unnoticeable, so it was merely a 'do the right thing' kind of thinking.

PR #27, opened 2017-11-20
Minor improvements for BigInteger and its usage

axxel:


axxel:
@huycn Please don't apply that PR at the moment. I seems I introduced an issue that I just saw with the help of the ASAN tool. I need to have a closer look what went wrong.

axxel:
Fixed.

core/src/pdf417/PDFDecodedBitStreamParser.cpp
```diff
@@ -486,9 +466,20 @@ Remove leading 1 =>  Result is 000213298174000
 */
 static DecodeStatus DecodeBase900toBase10(const std::vector<int>& codewords, int count, std::string& resultString)
 {
+	// Table containing values for the exponent of 900.
+	static const auto EXP900 = []() {
+		std::array<BigInteger, 16> table = {1, 900};
```
huycn: 👍 It's true that EXP900 should has no place in BigInteger

PR #31, opened 2018-02-07
Fix DMDetector deadlock

### LINKED ISSUE #30: DMDetector.traceGaps() endless loop
@axxel Thank you for your improvements to the detection of datamatrix codes!

I've found an example where the detector loops forever in the traceGaps() method.
Probably due to rounding in traceStep()  p doesn't change after an iteration.

frame_00034

axxel:
This mainly fixes #30 plus adds a few minor improvements/cleanups that I still had lying around.


huycn:
Thanks for the quick response and also for the quick fix!

PR #32, opened 2018-02-08
2nd attempt at fixing #30

axxel:
Disclamer: I still did not prove that the new code makes it impossible to end up in an endless loop again.

### DIFF
--- a/core/src/datamatrix/DMDetector.cpp
+++ b/core/src/datamatrix/DMDetector.cpp
@@ -810,10 +810,15 @@ public:
 				if (!finishLine.isValid() && gaps >= 4)
 					return true;
 			}
-			if (line.isValid()) {
-				// if we are drifting towards the inside of the code, pull the current position back out onto the line
-				if (line.signedDistance(p) > 2)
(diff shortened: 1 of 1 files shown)


bommo1:
2nd attempt looks good. thanks!

bommo1:
It took me more this time but here it is:

bug2

bommo1:
@axxel I've added this hack on my side to avoid the rare cases when the problem occurs:

PointI old_next_p(-1, -1);
do {
   ...
   PointI next_p = round(p);

   if (old_next_p.x != -1 && old_next_p == next_p) {
      return false; // it staled; exit to avoid endless loop
   }
   old_next_p = next_p;
   ...

axxel:
Thanks for the feedback. I discarded your approach when first thinking about a fix, because my gut feeling told me, I'd have to check a hole history of points. I considered that to be too complex and expensive. And it'd still be a 'hack', too. But your example helped me fix a corner case that I overlooked before.
After having spend even more time working on this, I would now be surprised if you found yet another example. Please surprise me! ;)

PR #36, opened 2018-02-10
Fix another endless loop bug in the DMDetector

axxel:
The important change is the second commit. This is actually not a '3rd attempt' of fixing #30, it is fixing a different issue (which also resulted in an endless loop, though). Thanks @bommo1 for all the beta-testing ;).

### DIFF
--- a/core/src/datamatrix/DMDetector.cpp
+++ b/core/src/datamatrix/DMDetector.cpp
@@ -502,6 +502,11 @@ template <typename T> double distance(PointT<T> a, PointT<T> b)
 using PointI = PointT<int>;
 using PointF = PointT<double>;
 
+template <typename T> PointF normalized(PointT<T> a)
+{
+	return PointF(a) / distance(a, {});
+}
+
 PointI round(PointF p)
 {
 	return PointI(std::lround(p.x), std::lround(p.y));
@@ -515,7 +520,7 @@ class RegressionLine
 
 	friend PointF intersect(const RegressionLine& l1, const RegressionLine& l2);
 
-	void evaluate(const std::vector<PointI> ps)
+	bool evaluate(const std::vector<PointI> ps)
 	{
 		auto mean = std::accumulate(ps.begin(), ps.end(), PointF()) / ps.size();
 		double sumXX = 0, sumYY = 0, sumXY = 0;
@@ -536,6 +541,7 @@ class RegressionLine
 			b = -b;
 		}
 		c = normal() * mean; // (a*mean.x + b*mean.y);
+		return _directionInward * normal() > 0.5; // angle between original and new direction is at most 60 degree
 	}
 
 	template <typename Container, typename Filter>
@@ -555,7 +561,7 @@ public:
 	const std::vector<PointI>& points() const { return _points; }
 	int length() const { return _points.size() >= 2 ? int(distance(_points.front(), _points.back())) : 0; }
 	bool isValid() const { return !std::isnan(a); }
-	PointF normal() const { return PointF(a, b); }
+	PointF normal() const { return {a, b}; }
 	double signedDistance(PointI p) const { return normal() * p - c; }
 	PointF project(PointI p) const { return p - signedDistance(p) * normal(); }
 
@@ -563,12 +569,12 @@ public:
 
 	void add(PointI p) { _points.push_back(p); }
 
-	void setDirectionInward(PointF d) { _directionInward = d; }
+	void setDirectionInward(PointF d) { _directionInward = normalized(d); }
 
-	void evaluate(bool clean = false)
+	bool evaluate(bool clean = false)
 	{
 		auto ps = _points;
-		evaluate(ps);
+		bool ret = evaluate(ps);
 		if (clean) {
 			size_t old_points_size;
 			while (true) {
@@ -581,9 +587,10 @@ public:
 #ifdef PRINT_DEBUG
 				printf("removed %zu points\n", old_points_size - _points.size());
 #endif
-				evaluate(_points);
+				ret = evaluate(_points);
 			}
 		}
+		return ret;
 	}
 
 	double modules(PointF beg, PointF end) const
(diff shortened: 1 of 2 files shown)


bommo1:
There you go: 
tmp

axxel:
@huycn Please wait with merging that one as there is obviously more work to be done, still...

bommo1:
@axxel Maybe this info might help into further improving the detector:
Sometimes perfectly aligned dm-codes cannot be read, but after applying a small rotation they are.

02_unrot

02_rot

 I can provide more examples like this when necessary.

axxel:
Interesting. Could you do me a favor and remove the 3 calls to `moveHalfAPixel` from `SampleGrid` and send see if you find more of those examples that still fail for only one of two orientations? Thanks.

bommo1:
Removing the calls to `moveHalfAPixel` helped in 3 out of 13 cases.

With the latest commit I couldn't reproduce until now the endless loop. I'll try some more.

axxel:
If you have 13 pairs ready for testing, could you sent them to me somehow? Dropbox/Wetransfer/etc... ?

bommo1:
https://wetransfer.com/downloads/333c2eee1cd0b3b249c43edada16840920180212144403/939ef2002891cabbaceab0de5688b02d20180212144403/69e326

bommo1:
Your latest commit seams to solve the endless loop problem but has a side effect: more barcodes (around 3% in my testset) need now to be rotated in order to be read.

axxel:
Can you give me one of those that need to be rotated to be detected after the latest commit? Or tell me the filename if it should be in the above zip? (thanks for that).

bommo1:
9_orig

or in zip/with_moveHalfAPixel

11_orig.png
13_orig.png
17_orig.png
19_orig.png
1_orig.png
21_orig.png
23_orig.png
37_orig.png
3_orig.png
41_orig.png
47_orig.png
49_orig.png
51_orig.png
53_orig.png
57_orig.png
61_orig.png
63_orig.png
73_orig.png
75_orig.png
7_orig.png
9_orig.png

bommo1:
Actually the I get the best results **without rotations** using the code stand before the "endless loop" fixes.

https://github.com/nu-book/zxing-cpp/commit/ad994b3ae416ad149a80e82ad5e78f750472ca32

axxel:
@huycn please merge: all known deadlocking examples work now.

@bommo1 the latest version is a slight improvement in detection quality (it detects one more of my own sample set) but still fails for about the same number of examples from your set. I spent quite some time looking at differences and double checking the sub-pixel precision placement of the 4 corner points determining the sample grid (like the `moveHalfAPixel` code) and am convinced that the current version is pretty much the best you can do.

Below is a set of images where one works and the other does not. You can see, the placing of the grid sample points is pretty much the same and about as good as it can get (and that is all the DMDetector is doing). The main differences come from the binarization, which is susceptible to slight  changes  in brightness coming from the interpolation during your rotation step.

The ones still not working are simple too blurry to be detected reliably, I'm afraid. If one really wanted to improve the situation, one would have to discard the hard binarization and use the grey-level info from the original inside the reed-solomon decoder. A simple scheme would be to randomly change those grid-points that are close to the black-white threshold. That would increase the risk of decoding something that was not really there, though. The basic approach of the whole library would need to change to explore that idea. I'm not going to go there at the moment.

I'f you have other ideas how to improve the detector, please let me know.

3_orig (works)
3_rot (fails)

bommo1:
@axxel thank you for the information!

Unfortunately I can still reproduce the endless loop:



Overall I still get better results with https://github.com/nu-book/zxing-cpp/commit/ad994b3ae416ad149a80e82ad5e78f750472ca32.

bommo1:
@axxel thank you again for your patience! (you probably want to shoot me by now :)

axxel:
@huycn hit the pause button once again, please :-/

@bommo1 thanks for your patience and your constructive input. I want to make that code work, not shoot people :). As long as you come up with endless loop examples, I will fix them.

axxel:
@bommo1 Latest endless-loop issue fixed + minor improvements that will not generally help with your blurred samples. I combined your two subfolders from the zip-file which now contains 86 files. 21 of them currently still fail. With the old version you mentioned, you will certainly get different results, maybe the total number is a bit smaller but it will generally be just other angles that fail.

I'll wait 24 hours before asking huycn to merge it, in case you find more endless loop samples ;).

bommo1:
There you go:
bug

bommo1:
One example where the image looks perfectly fine but the code cannot be read:

axxel:
That looks indeed pretty fine at a first glance. It fails because the lower black line is a little too thick towards the bottom and also a little bent. Still, I'd like my detector to be able to cope with that and I have an idea what to do. But right now I have to stop with the fun stuff and do some work for a change ;). Expect this to work in the coming days. If you have 1 or 2 more samples like this one, that might help.

bommo1:
It looks good now. I couldn't generate any "endless loop" image.

axxel:
Nice.

@huycn please unpause the merge button :)

bommo1:
Here are 2 more examples. I'll also try to naively un-distort (without prior calibration) the rectangle.

axxel:
That is an impressive datamatrix ;). May I ask what application domain you have for those kind of input?

The first one might work with my idea, the second definitely not. The library currently works with 4 corner points and generates a regular sampling grid between those. It works with bended input as long as the module size is sufficient that you can still draw a straight line without crossing from one column to the next. The second example crosses like 3 columns on the right side.

bommo1:
Medical. The big one represents a "[Medikationsplan](https://www.abda.de/uploads/tx_news/BMP_Anlage3_Unterschriftenverfahren.pdf)".

huycn:
So it's stable now ;) I will merge it since anyway, it works better than the head. If there is still problem, we just fire another issue.
Many thanks to @bommo1 for all the testings and samples and of course to @axxel for working on the issue. It's very much appreciated.

axxel:
@bommo1 Standardisierte Medikationspläne? Interessant. Noch nie davon gehört. Mal telefonieren? -> PM an awagger@gmail.com

PR #41, opened 2018-07-23
fix build with gcc 4.9 on android

swex:


axxel:
May I ask two questions (as someone with no experience developing on android):
1) Would not the 'upstream' Java library be the best option to get this functionality into an android app?
2) Is gcc 4.9 really the 'current' compiler to use for android development? I heard the NDK dropped GCC support altogether over 2 years ago.

swex:
1. Yes 'upstream' java library must be the first choice for android apps
2. It's still present but will be "finally" dropped in NDK r18

But old "dropped" GCC and zxing-cpp still relevant for Qt  based android applications. As of now official Qt android builds still built with GCC (and I'm not sure whether it change anytime soon).

So I did [small Qt wrapper](https://github.com/swex/QZXingNu) around your work. 
(for some reason [well known Qt zxing wrapper](https://github.com/ftylitak/qzxing)  failed to decode some QR codes)

huycn:
Thank swex for your contribution.

axxel:
@swex: Thanks for your detailed answer to my questions.

PR #43, opened 2018-10-03
Add the ability to install the library and include files

vkrause:
This is needed when using zxing as an external dependency e.g. in Linux
distributions.


huycn:
Thank you, Volker, for your contribution!

PR #52, opened 2019-01-06
Wasm support

huycn:
Add support for WebAssembly using Emscripten


axxel:
_nice_ ! You would not happen to have your compiled sample lying around somewhere publicly accessible to give it a try on my browser? (And maybe even add the link to the readme for others to find it?).

huycn:
I put a demo here: https://nu-book.github.io/zxing-cpp/demo.html

PR #54, opened 2019-01-27
Limit Android workarounds to GCC

vkrause:
Clang also defines __GNUC__ and thus enables these workarounds too, which
break the build there (ambiguity on std::stoi).

### DIFF
--- a/core/src/ZXStrConvWorkaround.h
+++ b/core/src/ZXStrConvWorkaround.h
@@ -14,7 +14,7 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-#if defined(__ANDROID__) && defined(__GNUC__)
+#if defined(__ANDROID__) && defined(__GNUC__) && !defined(__clang__)
 
 #include <sstream>
 #include <cstdlib>


axxel:
Thanks!

Question: @huycn and I were discussing the possibility to increase the c++ version to 14, maybe even a "widely supported subset of c++17" (think VisualStudio 2017, clang 5, gcc 7). I'm not an android developer, what clang/gcc version would you consider to be "widely available" for targeting android?

vkrause:
GCC is no longer supported on Android, and not even contained in the current NDK anymore.
"clang --version" gives me "Android (4751641 based on r328903) clang version 7.0.2 (https://android.googlesource.com/toolchain/clang 003100370607242ddd5815e4a043907ea9004281) (https://android.googlesource.com/toolchain/llvm 1d739ffb0366421d383e04ff80ec2ee591315116) (based on LLVM 7.0.2svn)" with NDK r18.
We are using C++14 on Android, so that certainly works, haven't tried C++17 features there yet though.

axxel:
clang-7... yummy :). Thanks for the feedback.

PR #66, opened 2019-05-02
Fix #63 and add a black box test for structured append qrcode

### LINKED ISSUE #63: Structured Append missing "N"
The "structured append" paging header has three parts: M, N and parity. M is the index of the current code, N is the number of codes in the set, and parity is the checksum to identify them all as belonging together. M and N are both 4 bits, but

https://github.com/nu-book/zxing-cpp/blob/e1d7de9cba1d912a3a13bd5e4c2e6deb80643d41/core/src/qrcode/QRDecoder.cpp#L340-L348

pulls both M and N into the same 8 bit variable. They _should_ be separated and exposed as separate values. The `DecoderResult` misleadingly calls this 8-bit (M,N) tuple the "sequence number":

https://github.com/nu-book/zxing-cpp/blob/e1d7de9cba1d912a3a13bd5e4c2e6deb80643d41/core/src/qrcode/QRDecoder.cpp#L414

when in fact only the first four bits are that.

huycn:
Not sure how to check parity but anyway, we don't need to check in this test


core/src/Result.cpp
```diff
@@ -30,8 +30,8 @@ Result::Result(std::wstring&& text, ByteArray&& rawBytes, std::vector<ResultPoin
 
 Result::Result(DecoderResult&& decodeResult, std::vector<ResultPoint>&& resultPoints, BarcodeFormat format)
     : _status(decodeResult.errorCode()),
-      _text(std::move(decodeResult).text()),
-      _rawBytes(std::move(decodeResult).rawBytes()),
+      _text(std::move(decodeResult.text())),
+      _rawBytes(std::move(decodeResult.rawBytes())),
```
huycn: Note also this unnoticeable issue/bug that I saw just by chance.

axxel: There was actually no issue before and now you introduced one: your change introduces a string/array copy where none was present before. Note the rvalue reference getters in DecoderResult used to steal the contests of `text/rawBytes`. For them to be selected, the object they are called on need to be an rvalue reference. Since `decodeResult` has a 'name' in this context, it is an lvalue reference and needs to be cast back with `std::move`. You are calling `std::move` on a `const &`, which makes no sense and will result in calling the copy constructor.

test/common/QRCodeStructuredAppendReader.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+TestReader::ReadResult
+QRCodeStructuredAppendReader::readMultiple(const std::vector<std::wstring>& filenames, int rotation) const
+{
+	TestReader::ReadResult result;
+	DecodeHints hints;
+	QRCode::Reader reader(hints);
+	std::list<Result> allResults;
+	int prevParity = 0;
```
axxel: I have no idea how that parity feature works either, but since it seems to behave like an 8-bit checksum that needs to be equal in all code symbols (judged by looking at your code), I suggest to change the 'uninitialized value' to `-1` to make sure it can not be a value that is actually returned from `getBits()`.

test/common/QRCodeStructuredAppendReader.cpp
```diff
@@ -0,0 +1,73 @@
...
+	for (const auto& imagePath : filenames) {
+		auto image = _imageLoader->load(imagePath);
+		ZXing::HybridBinarizer binarizer(image, false);
+		auto r = reader.decode(*binarizer.rotated(rotation));
+		if (r.metadata().getInt(ResultMetadata::STRUCTURED_APPEND_CODE_COUNT, 0) != filenames.size()) {
+			return TestReader::ReadResult();
+		}
+		auto parity = r.metadata().getInt(ResultMetadata::STRUCTURED_APPEND_PARITY, 0);
+		if (prevParity != 0 && prevParity != parity) {
```
axxel: Those 2 lines would need the `-1` then, as well...

test/common/QRCodeStructuredAppendReader.cpp
```diff
@@ -0,0 +1,73 @@
...
+		auto s1 = r1.metadata().getInt(ResultMetadata::STRUCTURED_APPEND_SEQUENCE, -1);
+		auto s2 = r2.metadata().getInt(ResultMetadata::STRUCTURED_APPEND_SEQUENCE, -1);
+		return s1 < s2;
+	});
+
+	if (!allResults.empty()) {
+		result.format = "QR_CODE";
+		for (const auto& r : allResults) {
+			auto txt = r.text();
+			result.text.append(txt);
```
axxel: slight performance improvement suggestion: remove the explicit `txt` variable and call `r.text()` directly:
```suggestion
			result.text.append(r.text());
```
This would remove an extra string copy. Or use `auto& txt = ...` to the same effect.

core/src/Result.cpp
huycn: You are totally right, there were no copies before, and my changes create copies. My confusion came from my assumption (without checking, my bad) that text() and rawBytes() return `const &` and `&` for const and non-const object respectively, but in fact, they return  `const &` and `&&`.

huycn:
Changes committed. You may want to squash before merge. Thanks.

PR #68, opened 2019-05-15
Fixed compilation for VS2019:

benys:
I had problems with VS2019 compilation. I made little fixes. Also I tested compilation at Linux.

* They remove sstream from string - added sstream header
* They remove no_init_all - added define


huycn:
Nice, thanks!

PR #72, opened 2019-05-19
Fixes #70 and fixes #71; shared library with version

### LINKED ISSUE #70: Hardcoded "lib" in CMakeLists.txt for installation of the library
Some systems might have a different libdir suffix (e.g. `lib64`). The correct solution would be to allow to use LIB_SUFFIX.

### LINKED ISSUE #71: Shared library has no soversion
It's only installed as a single `.so` instead of `so.$soversion`. A soversion is important at least for coinstallability (and important to get this library packaged in distributions).

huycn:
Also:
- Version is now defined in CMakeLists.txt, include/ZXing/XZVersion.h is generated with corresponding numbers.
- Project definition is moved from core/CMakeLists.txt to ./CMakeLists as it's more logical for it to be there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ set (ZXING_VERSION_PATCH 5)
 
 project (ZXingCpp VERSION ${ZXING_VERSION_MAJOR}.${ZXING_VERSION_MINOR}.${ZXING_VERSION_PATCH})
 
-set (ENABLE_ENCODERS OFF CACHE BOOL "Check to include encoders")
+set (ENABLE_ENCODERS ON CACHE BOOL "Check to include encoders")
 set (ENABLE_DECODERS ON CACHE BOOL "Check to include decoders")
 set (LINK_CPP_STATICALLY OFF CACHE BOOL "MSVC only, check to link statically standard library (/MT and /MTd)")
 set (BUILD_SHARED_LIBRARY OFF CACHE BOOL "Check to build ZXingCore as shared library")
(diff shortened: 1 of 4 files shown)


huycn:
@vkrause When you have few minutes, can you please have a look to make sure this PR makes sense. As I have no experience making package for linux distribution, I'm not too sure basic conventions are followed. The result of "make install" looks good though.

Note also that version numbers are no longer defined in XZConfig.h but ZXVersion.h instead. This later is generated by cmake from constants defined in CMakeLists.txt.

CMakeLists.txt
```diff
@@ -1,2 +1,77 @@
...
+	set (CMAKE_BUILD_TYPE "${DEFAULT_BUILD_TYPE}" CACHE STRING "Choose the type of build." FORCE)
+    set_property (CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+endif()
+
+add_subdirectory (core)
+
+set_target_properties (ZXingCore PROPERTIES VERSION ${PROJECT_VERSION})
+set_target_properties (ZXingCore PROPERTIES SOVERSION ${ZXING_VERSION_MAJOR})
+
+set (CMAKECONFIG_INSTALL_DIR "lib/cmake/ZXing")
```
vkrause: set (CMAKECONFIG_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/ZXing")

huycn: 👍  thanks!

PR #73, opened 2019-05-29
Python wrapper

timrae:
Hi, thanks for the great work on this library, it's a great fork!
I've made some experimental python bindings using [`pybind11`](https://github.com/pybind/pybind11), which I hope someone finds useful.

So far I just exposed the things that I need, but I'm open to suggestions for other classes / fields to be exposed.

### DIFF
--- a/wrappers/python/zxing.cpp
+++ b/wrappers/python/zxing.cpp
@@ -10,72 +10,67 @@
 #include "Result.h"
 #include "TextUtfEncoding.h"
 
+using namespace ZXing;
+
 #if 0
-using Binarizer = ZXing::GlobalHistogramBinarizer;
+using Binarizer = GlobalHistogramBinarizer;
 #else
-using Binarizer = ZXing::HybridBinarizer;
+using Binarizer = HybridBinarizer;
 #endif
 
 namespace py = pybind11;
 
+
 // Note: Image is a Numpy array in BGR format (the default opencv format)
 using Image = py::array_t<uint8_t, py::array::c_style>;
 
-std::string decode(const Image& image, bool fastMode, bool tryRotate, ZXing::BarcodeFormat format) {
-	ZXing::DecodeHints hints;
+Result decode(const Image& image, bool fastMode, bool tryRotate, BarcodeFormat format) {
+	DecodeHints hints;
 	hints.setShouldTryHarder(!fastMode);
 	hints.setShouldTryRotate(tryRotate);
-	if (format != ZXing::BarcodeFormat::FORMAT_COUNT) {
+	if (format != BarcodeFormat::FORMAT_COUNT) {
 		hints.setPossibleFormats({ format });
 	}
-	ZXing::MultiFormatReader reader(hints);
+	MultiFormatReader reader(hints);
 	const auto height = image.shape(0);
 	const auto width = image.shape(1);
 	const auto channels = image.shape(2);
 	const auto bytes = image.data();
-	ZXing::GenericLuminanceSource source(width, height, bytes, width*channels, channels, 2, 1, 0);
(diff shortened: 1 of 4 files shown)


timrae:
Tested working well on OSX and Ubuntu 16.04. Right now I guess it won't work on Windows, due to wget and tar commands in the cmake. Proposed solution would be to handle those two things in `setup.py` instead of cmake. Alternative solution would be to add pybind11 as a git submodule.

Let me know what you think, and we can go from there.

huycn:
This is very great! I'm pretty sure that many people will find this useful, especially when Python's popularity keeps rising.

I used to know only boost's python binding but the issue was always that you almost need to pull in almost everything of boost. pybind11 seems to be a very good alternative.

I will take care to make it work on Windows.

Thank you for your contribution :)

wrappers/python/zxing.cpp
```diff
@@ -0,0 +1,85 @@
...
+		.export_values();
+	py::class_<ResultPoint>(m, "ResultPoint")
+		.def_property_readonly("x", &ResultPoint::x)
+		.def_property_readonly("y", &ResultPoint::y);
+	py::class_<Result>(m, "Result")
+		.def_property_readonly("valid", &Result::isValid)
+		.def_property_readonly("text", &Result::text)
+		.def_property_readonly("format", &Result::format)
+		.def_property_readonly("points", &Result::resultPoints);
+	m.def("decode", &decode, "Decode a barcode from a numpy BGR image array",
```
timrae: Oops I forgot to update the documentation to say that it can accept either monochrome or BGR

timrae:
Yes no problem to squash it, I just kept it separated in commits in case you asked me to roll anything back. Could you please update the doc string as well as indicated in my comment?

wrappers/python/zxing.cpp
huycn: No problem, just add your commit.

bohrax:
Great project, and the Python binding is a big bonus! I tried to add the `rawBytes` attribute from the `Result` class into the Python binding, but my naive attempt below:

```
        .def_property_readonly("raw", &Result::rawBytes)
```
yields this runtime error:
```
Traceback (most recent call last):
  File "/home/bohre/bin/decode", line 13, in <module>
    print res.raw
TypeError: Unable to convert function return value to a Python type! The signature was
	(arg0: zxing.Result) -> ZXing::ByteArray
```

Would you be interested in adding this, or could you point me in the right direction?

timrae:
As the error says, presumably you just need to also add a wrapper for the `ZXing::ByteArray` type

bohrax:
I managed to get the raw data out by adding a new method in the `Result` class that was automatically converted by pybind. 

```
	const std::vector<unsigned char>& raw() const {
	    return _rawBytes;
	}
```
I imagine it would be possible to do in the zxing.cpp wrapper definition only, but I failed to find the correct incantation. Would my best be to read up on Classes or Type conversions in pybind?

huycn:
I'm not expert in pybind11, but I guess the proper ways would be using custom type caster (https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html) or as @timrae mentioned, exposing ByteArray to python (which would be cumbersome implementation-wise but would have better performance since no copies involved).

There is another, faster way:
```
typedef const std::vector<uint8_t>& (Result::*GetRawVectorFunc)() const;
...
.def_property_readonly("raw", (GetRawVectorFunc)&Result::rawBytes)
```
This will break in hard way if `ByteArray` no longer inherits `std::vector` or if `ByteArray` become polymorphic (virtual method added). However, both conditions are unlikely to happen.

bohrax:
Thanks! That was exactly what I was looking for!

PR #83, opened 2019-07-12
Fix 77: Export all symbols when building shared lib on Windows

huycn:


core/src/oned/ODCode39Writer.cpp
```diff
@@ -151,7 +153,7 @@ Code39Writer::encode(const std::wstring& contents, int width, int height) const
 	}
 	
 	if (extendedContent.empty()) {
-		extendedContent.append(contents.begin(), contents.end());
+		extendedContent = TextEncoder::FromUnicode(contents, CharacterSet::ISO8859_1);
```
huycn: This is just to fix some warning on conversion `wchar_t` to `char`

PR #88, opened 2019-07-17
get qr location point data

devilsen:
When i use zxing in android, i need to zoom in the camera automatic when i was far from the qrcode image, but i found If i get the points data ,i got the qrcode text too, so i add the points location data method, even the decoder did'nt get the result text ,but i can control my camera by this points.

### DIFF
--- a/core/src/MultiFormatReader.cpp
+++ b/core/src/MultiFormatReader.cpp
@@ -101,8 +101,11 @@ MultiFormatReader::read(const BinaryBitmap& image) const
 {
 	for (const auto& reader : _readers) {
 		Result r = reader->decode(image);
-  		if (r.isValid())
+  		if (r.isValid()){
 			return r;
+  		}else if (r.isBlurry()){
+			return r;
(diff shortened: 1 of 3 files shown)


axxel:
To make sure I understood: you have a situation where the `QRDetector` seemingly successfully found the code (the 3 finder pattern corners) but the decoder failed to decode the symbol (maybe because of some checksum error). Now you'd like to get access to the position of finder pattern to focus your camera on that area?

I can see how that could be useful but your proposed `isBlurry()` function is not the way to go in my opinion. If you could use the `QRCode::Reader` instead of the `MultiFormatReader`, then I see an easy fix to support your use case without disrupting the `MultiFormatReader`.

axxel:
Now, if you preselect `QR_CODE` when constructing your `MultiFormatReader` the current code should return what need.

devilsen:
hi @axxel  i seen your change just now,  that is batter than my commit, thanks

PR #91, opened 2019-07-26
fix thread-safety regression in RSS* decoders

axxel:
Introduce little ThreadLocal template helper to implement some kind of 'thread local storage' for the RSS-type readers that support 'stacked' 1D codes (i.e. 2D codes actually) and need to store some state between decoding different lines.

This is a fix for the thread-safety regression introduced in 17284ea7775d9a4da034a3f817320fbb10813de0

### DIFF
--- a/core/src/oned/ODRSS14Reader.cpp
+++ b/core/src/oned/ODRSS14Reader.cpp
@@ -403,19 +403,20 @@ ConstructResult(const RSS::Pair& leftPair, const RSS::Pair& rightPair)
 Result
 RSS14Reader::decodeRow(int rowNumber, const BitArray& row_) const
 {
+	auto& possiblePairs = state();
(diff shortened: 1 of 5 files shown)


axxel:
That is true: coming up with new threads would potentially grow that map larger and larger. Preventing that would require to hold some cleanup-token on the stack and then I'd be almost exactly where I started.

I agree that the original has the advantage of cleaning that all up during stack-unwinding and is less involved. I'll revert it but still do away with the dynamic_cast as there is only ever exactly one type at the end of the pointer.

The real issue, I think, is that those RSS-decoders are not 1D but 2D decoders and simply don't fit properly in this `RowReader` abstraction. Also the fact that they get fed the lines in a 'random' order makes no real sense.

PR #93, opened 2019-07-31
fix to_string and stoll bug for android

seed93:


axxel:
Thanks! (looks like you are the first to build the encoders on the affected android setup :))

Issue #99, opened 2019-09-25
Issue with MSVC iterator guard in DEBUG


enzo-simone:
Hello,
this is gonna be a debatable one as it deals with Microsoft specific environment, but let me expose.

Method `ZXing::BitArray::isRange` is defined as
`
	bool isRange(int start, int end, bool value) const {
		return std::all_of(&_bits[start], &_bits[end], [value](uint8_t v) { return v == static_cast<int>(value); });
	}
`
but accessing `_bits[end]`, i.e. past the end, triggers an assertion just during debugging **though is perfectly legitimate to get its address since &_bits[end] == &_bits[0] + end**.

A solution could be 
`return std::all_of(std::begin(bits) + start, std::begin(bits) + end, [value](uint8_t v) { return v == static_cast<int>(value); });`

Thank you.

axxel:
I'm not a MSVC user but it seems to be a valid point to me. Care to prepare a pull request?

PR #100, opened 2019-09-26
issue #99 - avoid referencing past the end element of a vector

enzo-simone:
issue #99 - avoid referencing past the end element of a vector; when parameter `end` equals `_bits.size()` then `_bits[end]` references a past-the-end element which triggers a debug assertion in MSVC, even if `&_bits[end]` is absolutely legal and safe

### DIFF
--- a/core/src/BitArray.h
+++ b/core/src/BitArray.h
@@ -299,9 +299,7 @@ public:
 	*/
 #ifdef ZX_FAST_BIT_STORAGE
 	bool isRange(int start, int end, bool value) const {
-        auto start_it = std::cbegin(_bits) + start;
-        auto end_it = std::cbegin(_bits) + end;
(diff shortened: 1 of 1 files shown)


enzo-simone:
> 
> 
> As you can see, this does not compile on the travis ci system. You would need to qualify `cbegin` and `advance` with `std::`. But even then, why not simply use the version you already posted in your issue ticket? Introducing an extra variable and using the comma operator in this manner is not to my taste ;).

Yes, I did it to minimize creation of iterators instances, so I created an iterator and reused it, but I agree with you code is less maintainable.

I provided a new implementation.

enzo-simone:
No problem at all, I prefer spaces over tabs since code looks the same on all machines, but 'When in Rome, do as the Romans do' ;)
Thank you for all your efforts, I'm going to extensively use the library and will be happy to help.
Question: am I wrong or CMake file is missing for tests? Unfortunately I don't know how to create it, I just can barely use cmake-gui, sorry.

Issue #86, opened 2019-07-16
CMake restructuring


axxel:
I'd like to propose a change in the CMake setup. Instead of having lots of individual cmake projects in each subdirectory that all build the whole library themselfs, we should have one top-level project with a bunch of options for building with tests/examples/wrappers/etc. This should result in less confusion on what to build if you just want to have the lib with an example binary and we could also disable some wrappers based on the platform (see #85).

It would also reduce the amount of build directories to deal with during development just to make sure everything still compiles.

@huycn are you somehow dependent on the current 'multi-project' setup?

huycn:
@axxel That is a good idea. It started like that as it was my personal preferences to make each component as independent as possible, including build details. But as you mentioned, it's somehow confusing.

I think a good approach would be to have [CMakeLists.txt](../blob/master/CMakeLists.txt) as umbrella and forward build details to other CMakeLists.txt files in their own folders, the same way as currently, this file forwards the build of ZXingCore to [core/CMakeLists.txt](../blob/master/core/CMakeLists.txt).

At Nu-book, we don't depend on current structure of project as we use binary distributions from nuget.org and internal repo only. Thus as long it can build, we are fine with it.

axxel:
I spent about 3 full days on this performance optimization sprint, which I started just as a little fun exercise with a new profiler I discovered. That was way more than I planed on investing ;-). Meaning: I have to go back to other stuff now.

Also, since I don't have either a windows nor an android setup available, I could not really test the build setup properly. Meaning: if you have some time to spare to reorganize the cmake scripts, please feel free to do so. Otherwise, I might come back to this at a later point in time.

axxel:
I thought a little more about this and I think I would start with adding the examples and the tests to the top-level project. The wrappers could then come later (I can't built any of those right now).

I believe from a package maintainer and 'unixy' user perspective, the default built option should be a shared library and dynamically linked executables. The unittests will have to stay statically linked because as far as I recall, they rely on extra exported symbols at the moment.

I was wondering whether the name of the library should be changed from `ZXingCore` to maybe `libZXing` or `libZXingCpp`? or a lowercase version of those? The original port seems to use the name `libzxing` for the library and `zxing` for an executable. We should probably play nice and not conflict with those?

[...]

PR #106, opened 2019-10-07
Remove warnings and add some optimizations

enzo-simone:
Explicit narrow casting are useful to check if narrow was intended/tolerated.
Likewise for local variables having the same name as outer scope variables.

### DIFF
--- a/core/src/BitArray.h
+++ b/core/src/BitArray.h
@@ -285,7 +285,7 @@ public:
 	* Clears all bits (sets to false).
 	*/
 	void clearBits() {
-		std::fill(_bits.begin(), _bits.end(), uint8_t{});
+		std::fill(_bits.begin(), _bits.end(), 0);
 	}
 
 	/**
--- a/core/src/BitArray.h
+++ b/core/src/BitArray.h
@@ -285,7 +285,7 @@ public:
 	* Clears all bits (sets to false).
 	*/
 	void clearBits() {
-		std::fill(_bits.begin(), _bits.end(), 0);
+        std::fill(_bits.begin(), _bits.end(), uint8_t{});
 	}
 
 	/**
--- a/core/src/BitArray.h
+++ b/core/src/BitArray.h
@@ -285,7 +285,7 @@ public:
 	* Clears all bits (sets to false).
 	*/
 	void clearBits() {
-        std::fill(_bits.begin(), _bits.end(), uint8_t{});
+		std::fill(_bits.begin(), _bits.end(), uint8_t{});
 	}
 
 	/**
--- a/core/src/ByteMatrix.h
+++ b/core/src/ByteMatrix.h
@@ -34,7 +34,7 @@ class ByteMatrix
 
 public:
 	ByteMatrix() = default;
-	ByteMatrix(int width, int height, int val = 0) : _width(width), _height(height), _data(_width * _height, val) { }
+	ByteMatrix(int width, int height, int val = 0) : _width(width), _height(height), _data(_width * _height, static_cast<uint8_t>(val)) { }
 
 	ByteMatrix(ByteMatrix&&) = default;
 	ByteMatrix& operator=(ByteMatrix&&) = default;
--- a/core/src/LuminanceSource.cpp
+++ b/core/src/LuminanceSource.cpp
@@ -41,16 +41,16 @@ public:
 	const uint8_t* getRow(int y, ByteArray& outBytes, bool) const override
 	{
 		_src->getRow(y, outBytes, true);
-        for (auto& b : outBytes)
-            b = 255 - b;
+		for (auto& b : outBytes)
+			b = 255 - b;
 		return outBytes.data();
 	}
 
 	const uint8_t* getMatrix(ByteArray& outBytes, int& outRowBytes, bool) const override
 	{
 		_src->getMatrix(outBytes, outRowBytes, true);
-        for (auto& b : outBytes)
-            b = 255 - b;
+		for (auto& b : outBytes)
+			b = 255 - b;
 		return outBytes.data();
 	}
 
--- a/core/src/LuminanceSource.cpp
+++ b/core/src/LuminanceSource.cpp
@@ -41,14 +41,16 @@ public:
 	const uint8_t* getRow(int y, ByteArray& outBytes, bool) const override
 	{
 		_src->getRow(y, outBytes, true);
-		std::transform(outBytes.begin(), outBytes.end(), outBytes.begin(), [](uint8_t b) { return 255 - b; });
+        for (auto& b : outBytes)
+            b = 255 - b;
 		return outBytes.data();
 	}
 
 	const uint8_t* getMatrix(ByteArray& outBytes, int& outRowBytes, bool) const override
 	{
 		_src->getMatrix(outBytes, outRowBytes, true);
-		std::transform(outBytes.begin(), outBytes.end(), outBytes.begin(), [](uint8_t b) { return 255 - b; });
+        for (auto& b : outBytes)
+            b = 255 - b;
 		return outBytes.data();
 	}
 
--- a/core/src/WhiteRectDetector.cpp
+++ b/core/src/WhiteRectDetector.cpp
@@ -276,29 +276,29 @@ bool WhiteRectDetector::Detect(const BitMatrix& image, int initSize, int x, int 
 			return false;
 		}
 
-		ResultPoint x;
+		ResultPoint x_;
 		found = false;
 		//go down left
 		for (int i = 1; !found && i < maxSize; i++) {
-			found = GetBlackPointOnSegment(image, right, up + i, right - i, up, x);
+			found = GetBlackPointOnSegment(image, right, up + i, right - i, up, x_);
 		}
 
 		if (!found) {
 			return false;
 		}
 
-		ResultPoint y;
+		ResultPoint y_;
 		found = false;
 		//go up left
 		for (int i = 1; !found && i < maxSize; i++) {
-			found = GetBlackPointOnSegment(image, right, down - i, right - i, down, y);
+			found = GetBlackPointOnSegment(image, right, down - i, right - i, down, y_);
 		}
 
 		if (!found) {
 			return false;
 		}
 
-		CenterEdges(y, z, x, t, width, p0, p1, p2, p3);
+		CenterEdges(y_, z, x_, t, width, p0, p1, p2, p3);
 		return true;
 	}
 	else {
--- a/core/src/datamatrix/DMDecoder.cpp
+++ b/core/src/datamatrix/DMDecoder.cpp
@@ -573,7 +573,7 @@ CorrectErrors(ByteArray& codewordBytes, int numDataCodewords)
 
 	// Copy back into array of bytes -- only need to worry about the bytes that were data
 	// We don't care about errors in the error-correction codewords
-    std::transform(codewordsInts.begin(), codewordsInts.begin() + numDataCodewords, codewordBytes.begin(), [](int const& cw) { return static_cast<uint8_t>(cw); });
+	std::transform(codewordsInts.begin(), codewordsInts.begin() + numDataCodewords, codewordBytes.begin(), [](int const& cw) { return static_cast<uint8_t>(cw); });
 
 	return true;
 }
--- a/core/src/datamatrix/DMDecoder.cpp
+++ b/core/src/datamatrix/DMDecoder.cpp
@@ -573,7 +573,7 @@ CorrectErrors(ByteArray& codewordBytes, int numDataCodewords)
 
 	// Copy back into array of bytes -- only need to worry about the bytes that were data
 	// We don't care about errors in the error-correction codewords
-	std::transform(codewordsInts.begin(), codewordsInts.begin() + numDataCodewords, codewordBytes.begin(), [](int const& cw) { return static_cast<uint8_t>(cw); });
+	std::copy_n(codewordsInts.begin(), numDataCodewords, codewordBytes.begin());
 
 	return true;
 }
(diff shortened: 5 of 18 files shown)


core/src/LuminanceSource.cpp
```diff
@@ -41,14 +41,16 @@ class InvertedLuminanceSource : public LuminanceSource
 	const uint8_t* getRow(int y, ByteArray& outBytes, bool) const override
 	{
 		_src->getRow(y, outBytes, true);
-		std::transform(outBytes.begin(), outBytes.end(), outBytes.begin(), [](uint8_t b) { return 255 - b; });
+		for (auto& b : outBytes)
+			b = 255 - b;
 		return outBytes.data();
```
axxel: You could argue that the for loop is more readable but we generally went with the "Sean Parent" approach of avoiding raw loops if possible to increase reasonability/readability. Based on your comment, you seem to think the for loop is faster? To believe that I would have to be convinced with some assembler output or measurements. :)

core/src/WhiteRectDetector.cpp
```diff
@@ -276,29 +276,29 @@ bool WhiteRectDetector::Detect(const BitMatrix& image, int initSize, int x, int
 			return false;
 		}
 
-		ResultPoint x;
+		ResultPoint x_;
 		found = false;
```
axxel: I think this sort of change reduces the readability of the code (at least if applied no matter what). I personally don't buy the argument that shadowing a variable is inherently bad. Scoping rules are clear and I would personally recommend to disable that warning.

core/src/datamatrix/DMECEncoder.cpp
```diff
@@ -121,8 +121,8 @@ static void CreateECCBlock(ByteArray& data, int codeOffset, int codeLength, int
 	for (int i = 0; i < codeLength; ++i) {
 		const auto m = ecc.back() ^ data[codeOffset + i * stride];
 		for (size_t k = ecc.size() - 1; k > 0; k--)
-			ecc[k] = ecc[k - 1] ^ mult(m, poly[k]);
-		ecc[0] = mult(m, poly[0]);
+			ecc[k] = ecc[k - 1] ^ mult(static_cast<uint8_t const>(m), poly[k]);
+		ecc[0] = mult(static_cast<uint8_t const>(m), poly[0]);
 	}
```
axxel: east vs. west const is certainly an interesting debate. But worse than one or the other is mixing both. Please stick with the precedent set by the rest of the code.

core/src/BitArray.h
```diff
@@ -285,7 +285,7 @@ class BitArray
 	* Clears all bits (sets to false).
 	*/
 	void clearBits() {
-		std::fill(_bits.begin(), _bits.end(), 0);
+        std::fill(_bits.begin(), _bits.end(), uint8_t{});
 	}
```
axxel: Could you attempt to convince me why this is an improvement (apart from the ws fix)?

core/src/ByteMatrix.h
```diff
@@ -34,7 +34,7 @@ class ByteMatrix
 
 public:
 	ByteMatrix() = default;
-	ByteMatrix(int width, int height, int val = 0) : _width(width), _height(height), _data(_width * _height, val) { }
+	ByteMatrix(int width, int height, int val = 0) : _width(width), _height(height), _data(_width * _height, static_cast<uint8_t>(val)) { }
```
axxel: I believe the right fix here would be change the type of val to uint8_t.

I see how unintended narrowing conversions can potentially lead to issues but I believe the strategy of simply making warnings go away by adding explicit casts is counter productive. As can be seen in this first example of those changes. So before adding any of those, I would suggest to look hard if the issue is not actually somewhere else. And as long as the cast is not there one can still ask the compiler for a warning if something is wrong. With the cast the reader and the compiler would have to believe this is truly what the author intended and for this line (and maybe others, too) I believe it is not.

core/src/datamatrix/DMDecoder.cpp
```diff
@@ -573,7 +573,7 @@ CorrectErrors(ByteArray& codewordBytes, int numDataCodewords)
 
 	// Copy back into array of bytes -- only need to worry about the bytes that were data
 	// We don't care about errors in the error-correction codewords
-	std::copy_n(codewordsInts.begin(), numDataCodewords, codewordBytes.begin());
+	std::transform(codewordsInts.begin(), codewordsInts.begin() + numDataCodewords, codewordBytes.begin(), [](int const& cw) { return static_cast<uint8_t>(cw); });
```
axxel: E.g. here the comment explicitly states that we want to copy ints into bytes and the transform with the cast is obviously less readable (see comment above about the casting).

core/src/textcodec/GBTextEncoder.cpp
```diff
@@ -4000,13 +4000,13 @@ void GBTextEncoder::EncodeGB18030(const std::wstring& str, std::string& bytes)
 					bytes[index++] = replacement;
 					++invalid;
 				}
-				high = -1;
+				high = static_cast<unsigned>(-1);
```
huycn: I believe that was a bug that `high` is `unsigned`.

core/src/LuminanceSource.cpp
enzo-simone: Well, it's a range-based loop (a modern C++ citizen), which just expresses the intention to access and possibly modify (the `auto&` stuff) each element.
I went with the rule of the thumb to prefer references over the copies (i.e. the argument of the lambda), even if compilers would certainly optimize this on themselves.
Anyway, all my proposal are... well, just proposals! :)

core/src/datamatrix/DMECEncoder.cpp
enzo-simone: I agree, sorry not to have looked for const convention in use.

core/src/BitArray.h
enzo-simone: Just matching contained type and type of the filling value, not faster, more consistent.

core/src/ByteMatrix.h
enzo-simone: Absolutely, my intention is not to force my modifications into the code base but rather to raise a discussion about what was the original intention.

core/src/datamatrix/DMDecoder.cpp
enzo-simone: But in reality we are transforming `int`s to fit in an 8bit location. Again, just my point of view.

core/src/textcodec/GBTextEncoder.cpp
enzo-simone: I do not have enough knowledge about the overall design.
But this is a point in favor of resolving warnings :)

core/src/WhiteRectDetector.cpp
enzo-simone: I agree that shadowing is not inherently bad, but not shadowing is not bad for sure :)
Maybe something more talking like `ResultPoint local_x` would be preferable.
Just my 2cents.

core/src/textcodec/GBTextEncoder.cpp
huycn: Yep, definitively a bug as seen from [Qt code](https://github.com/qt/qtbase/blob/dev/src/corelib/codecs/qgb18030codec.cpp#L88-L144). The fix it either leave `high` declared as `unsigned` and change here and below to `high = 0`, or change the declaration of high `int high = -1`. In either case, no cast is needed.

enzo-simone: Ok @huycn, feel free to resolve this conversation.

huycn: I've pushed a fix. You can resolve if you find it's ok.

core/src/LuminanceSource.cpp
axxel: My point was: 1) performance is the exact same and 2) your option is actually more readable but less consistent with the rest. I'm trying to get comfortable with std algorithms, so would rather leave it as is. @huycn: we need a third opinion ;).

core/src/datamatrix/DMECEncoder.cpp
enzo-simone: Added commit c1b4ea1d5a11f5cd1a093d689dbea722d0fa29a3, but I will be still waiting opinions whether the whole `static_cast` thing is worth or not.

core/src/BitArray.h
enzo-simone: Added 73ed55c99166d2f3ace630717bc3c41665cca110 to fix the indentation issue. But still ready to withdraw the proposed change.

core/src/WhiteRectDetector.cpp
axxel: I had a look at the surrounding code and if one really wanted to change the names here, I'd change x->l, z->b, y->r and also do that in the `CenterEdges` function. This way, they would actually convey some meaning (as in left, bottom, top, right).

core/src/ByteMatrix.h
axxel: OK. Then I'd like to suggest to fix this and the buggy `high` variable and possibly check all other instances where you suggested the static_cast to see if there are more of those (who ever is interested in doing that) but leave rest as is for now.

core/src/LuminanceSource.cpp
huycn: Zut... I was trying to avoid the topic ;) 

To me, both are valid, of question of styles. `std::transform()` is a little bit harder to read at first, but it's not an issue once you get used to it. Personally, I still find the `for` loop is more readable, especially when once switches often between different languages at work (foreach exists in almost all languages I known). At the same time, the `std::transform` has an advantage of functional programming and more consistent with the rest.

So it will not be clean cut here and in such case, I normally go with "don't fix if it ain't broke" principle.

core/src/datamatrix/DMDecoder.cpp
axxel: I agree with your assessment. But I (personally) don't see the real benefit. If we are sure the implicit narrowing is without issues (there is no value outside of 0..255) then this only reduces readability. If we are not sure about that, then this does nothing more than the implicit narrowing except it suggest, that we actually want it to mask out higher bits (which we don't). In that second case, we should actually check that all values are valid. One would have to inspect the code and see if that narrowing can or can not fail (then the check would be unnecessary and misleading as well).

core/src/BitArray.h
axxel: Thanks. I'm not convinced though: The literal constant 0 is definitively more readable and more obvious and the narrowing conversion is 100% correct, always. I would argue that this is actually a false positive warning.

core/src/LuminanceSource.cpp
enzo-simone: OK, no problem, also where I come from any modification requires at least 50%+1 of the consensus.

Questions:
- am I allowed to add pragma warning disable directives to library code?
- am I allowed to wrap them in MSVC specific enabler and hope someone else would take care of introduce other compiler enablers (if ever needed)?

axxel: I believe the correct location to add diagnostic modifying flags is the CMakeLists.txt. See there for other examples. And if you wander why there are currently duplicates of those kind of settings, see also #86. So my answer to your 3 questions would be no, not applicable and no.

core/src/WhiteRectDetector.cpp
enzo-simone: I just proposed the trailing addition because had no knowledge on what those variables meant.
If renaming is decided maybe is worth renaming them in a more meaningful way.
No problem for me to follow your suggestions.

core/src/datamatrix/DMDecoder.cpp
enzo-simone: OK, withdrawing on this.

core/src/LuminanceSource.cpp
enzo-simone: Done in 6dc5997d37b4ae7d691a7326be4eb1ad2f97510a, anyway I'll wait for a feedback.

core/src/datamatrix/DMDecoder.cpp
enzo-simone: Added eb9a47632cdf3ee0a317e2992332f59947f16094.

core/src/BitArray.h
enzo-simone: OK, reverted in commit 9d2cbaf32f3cfaf4113e27e64740be27e9cc3cf1

axxel:
I went through the collected discussions again and I think the current bottom-line is:
 * fix `high`
 * change `val` parameter of `ByteMatrix` to `uint8_t`
 * MSVC silencing via build option
 * no static_cast<> sprinkling just to silence the compiler
 * no random renames just to silence the compiler
 * no algorithm -> for-loop conversion

Maybe it would be easiest to simply prepare a new PR with those 3 simple changes instead of reverting most of the changed lines back?

enzo-simone:
> 
> 
> I went through the collected discussions again and I think the current bottom-line is:
> 
>     * fix `high`
> 
>     * change `val` parameter of `ByteMatrix` to `uint8_t`
> 
>     * MSVC silencing via build option
> 
>     * no static_cast<> sprinkling just to silence the compiler
> 
>     * no random renames just to silence the compiler
> 
>     * no algorithm -> for-loop conversion
> 
> 
> Maybe it would be easiest to simply prepare a new PR with those 3 simple changes instead of reverting most of the changed lines back?

OK, I'll take care of it.

enzo-simone:
I've just realized that the default warning level for project ZXingCore is 3 (at least for MSVC), this means that all warnings I was ranting for are not enabled.
Though I prefer maximum level when building, I would stay this way until there will be a large consensus in increasing it.

So I'm starting a PR just for
* fix high
* change val parameter of ByteMatrix to uint8_t

Thank you for your guidance.

PR #110, opened 2019-10-21
ReadBarcode: update doc

dabrain34:
Naive documentation update. 

Would be nice to define the documentation, used in the future and keep a consistent way to document the API. 

Any idea for the framework ?


huycn:
That is a good idea for sure.
I only know Doxygen which is quite flexible in formats. However, JavaDoc format seems to be a logical choice here, thus I vote for it.

PR #117, opened 2020-01-08
Mark a few static as static const

tsdgeos:


tsdgeos:
@huycn ping?

huycn:
@tsdgeos pong ;)
Sorry, it has passed unnoticed

PR #121, opened 2020-03-06
Added python bindings for barcode encoding

bimusiek:
Hey, 
I've added python binding for encoding barcodes.

### DIFF
--- a/wrappers/python/zxing.cpp
+++ b/wrappers/python/zxing.cpp
@@ -9,6 +9,10 @@
 #include "GenericLuminanceSource.h"
 #include "HybridBinarizer.h"
 #include "Result.h"
+#include "TextUtfEncoding.h"
+#include "BitMatrix.h"
+#include "ByteMatrix.h"
+#include "MultiFormatWriter.h"
 
 using namespace ZXing;
 namespace py = pybind11;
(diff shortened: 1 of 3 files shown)


wrappers/python/zxing.cpp
```diff
@@ -47,6 +51,30 @@ Result decode(const Image& image, BarcodeFormat format, bool fastMode, bool tryR
...
 }
 
+Image encode(std::string text, int width, int height, BarcodeFormat format, int margin, int eccLevel) {
+	MultiFormatWriter writer(format);
+	if (margin >= 0)
+		writer.setMargin(margin);
+	if (eccLevel >= 0)
+		writer.setEccLevel(eccLevel);
+
+	auto bitmap = writer.encode(TextUtfEncoding::FromUtf8(text), width, height).toByteMatrix();
```
axxel: If you are going to do a pixel-by-pixel initialization of your memory below, then there is no need call toByteMatrix() here effectively doing that twice.

wrappers/python/zxing.cpp
```diff
@@ -47,6 +51,30 @@ Result decode(const Image& image, BarcodeFormat format, bool fastMode, bool tryR
...
+
+	auto bitmapWidth = bitmap.width();
+	auto bitmapHeight = bitmap.height();
+
+	py::array_t<uint8_t> result = py::array_t<uint8_t>(bitmapWidth * bitmapHeight * sizeof(uint8_t));
```
axxel: * Please use the `Image` type alias here to be consistent with the return type and the rest of the code.
 * Also, I believe there is a more suitable constructor that takes the final shape of the array (`{bitmap.width(), bitmap.height()}`)
 * That `* sizeof(uint8_t)` makes no sens to me here. Since it is equal to 1 it is no bug but misleading.

wrappers/python/zxing.cpp
```diff
@@ -47,6 +51,30 @@ Result decode(const Image& image, BarcodeFormat format, bool fastMode, bool tryR
...
+	auto buf3 = result.request();
+	int8_t *ptr3 = (int8_t *) buf3.ptr;
```
axxel: Since your array is explicitly containing `uint8_t`s, why use a `int8_t` here? Also, according to the pybind11 doc, it seems a more canonical way would be something like:
```
    auto r = x.mutable_unchecked<2>();
    for (ssize_t i = 0; i < r.shape(0); i++)
        for (ssize_t j = 0; j < r.shape(1); j++)
                r(i, j) = ...;
```

axxel:
Since there has not been any reaction to my change requests for 3 months and since I was working on the python module anyway, I implemented this. Will push soon after finally deciding on the naming of the function.

I'm strongly considering using the name `read_barcode` instead of `decode` to be consistent with the other interfaces and the internals of the library. Note: the term `decode` is already used inside the library for the second half of the `read` operation (the first being `detect`). In case the python module should later allow a more fine-grained access to the internals, we'd get into a naming clash.

I'd be interested if you have a strong argument against renaming that function. Any thoughts from @timrae or @urishab?

timrae:
I don't have any objections to that

axxel:
I did the rename and added a write_barcode function plus some handy string to format-enum helper when dealing with text input from the user.

PR #125, opened 2020-05-26
add missing <stdexcept> headers for msvc 14.26

DuckSoft:
compilation breaks when upgrading msvc from 14.25 to 14.26.
this pull request fixes it.

note: please squash


axxel:
Thanks! I'm looking into using iwuy to hopefully completely get rid of this kind of 'future' issue.

PR #133, opened 2020-06-20
Fix Android build

vkrause:
The recent changes in Point.h made the unqualified round() call ambiguous
here, but since Android nowadays does have std::round we can fix this and
simplify the code at the same time.

### DIFF
--- a/core/src/ZXNumeric.h
+++ b/core/src/ZXNumeric.h
@@ -83,11 +83,7 @@ template <typename T>
 typename std::enable_if<std::is_floating_point<T>::value, int>::type
 RoundToNearest(T x)
 {
-#if defined(__ANDROID__) && defined(__GNUC__)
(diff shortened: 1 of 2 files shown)


axxel:
Ah. my bad. But very nice to hear about progress on the android sdk side. Questions:
1. Is there a 'standard' / widely available version of the SDK that we could use as a base? (I should add that to the CI system)
2. Does that mean the `ZXStrCovWorkaround.h` can also go away?

vkrause:
(1) I think the general rule there is "latest", as Google tends to be quite aggressive in deprecating the older ones. In this case I suspect the SDK change that fixed this is the switch to a clang-based toolchain in 2018, the previous GCC toolchain was stuck on some older version for quite some time, which is what often caused these kind of issues.

(2) Indeed, commenting out everything in that file does not seem to break the build on Android here.

PR #137, opened 2020-06-24
Support for CMake package versioning

tanius:
This provides basic CMake package versioning support so that `find_package(ZXing 1.0.8 REQUIRED)` becomes possible. Follows the official example from the CMake manual ([cmake-packages(7) section 'Creating Packages'](https://cmake.org/cmake/help/v3.6/manual/cmake-packages.7.html#creating-packages)).

I [have read](https://github.com/swex/QZXingNu/issues/6#issuecomment-648652033) you'd welcome a PR for this :slightly_smiling_face: Tried to figure it out and it seems to work. I don't know much CMake though, so take with a grain of salt …

## Test results

All tests done with ZXing-C++ 1.0.8 @ d66c849 (2020-06-23).

### Without the patch

* **In a client project using `find_package(ZXing REQUIRED)`:** `cmake ..` runs successfully.

* **In a client project using `find_package(ZXing 1.0.8 REQUIRED)`:** `cmake ..` fails with:

     ```
    CMake Warning at CMakeLists.txt:171 (find_package): 
       Could not find a configuration file for package "ZXing" that is compatible 
       with requested version "1.0.8". 
       The following configuration files were considered but not accepted:
         /usr/local/lib/cmake/ZXing/ZXingConfig.cmake, version: unknown
    ```

### With the patch

* **In a client project using `find_package(ZXing REQUIRED)`:** `cmake ..` runs successfully.

* **In a client project using `find_package(ZXing 1.0.7 REQUIRED)`:** `cmake ..` runs successfully.

* **In a client project using `find_package(ZXing 1.0.8 REQUIRED)`:** `cmake ..` runs successfully.

* **In a client project using `find_package(ZXing 1.0.9 REQUIRED)`:** `cmake ..` fails with:

    ```
    CMake Error at CMakeLists.txt:172 (find_package): 
      Could not find a configuration file for package "ZXing" that is compatible 
      with requested version "1.0.9". 
      The following configuration files were considered but not accepted:
        /usr/local/lib/cmake/ZXing/ZXingConfig.cmake, version: 1.0.8
    ```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required (VERSION 3.10)
 
-project (ZXing VERSION "1.0.8" LANGUAGES CXX)
+set(ZXing_VERSION 1.0.8)
+project (ZXing VERSION ${ZXing_VERSION} LANGUAGES CXX)
 
 option (BUILD_WRITERS "Build with writer support (encoders)" ON)
 option (BUILD_READERS "Build with reader support (decoders)" ON)
@@ -106,6 +107,12 @@ ENDIF()
 
 include (CMakePackageConfigHelpers)
 
+write_basic_package_version_file(
+    "${CMAKE_CURRENT_BINARY_DIR}/ZXingConfigVersion.cmake"
+    VERSION ${ZXing_VERSION}
+    COMPATIBILITY AnyNewerVersion
+)
+
 configure_package_config_file (
     core/ZXingConfig.cmake.in
     ZXingConfig.cmake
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -106,13 +107,21 @@ ENDIF()
 
 include (CMakePackageConfigHelpers)
 
+write_basic_package_version_file(
+    "${CMAKE_CURRENT_BINARY_DIR}/ZXingConfigVersion.cmake"
+    VERSION ${ZXing_VERSION}
```
axxel, line 112: Could you please use `${PROJECT_VERSION}` instead. No need to introduce a new variable then.

tanius: Done as requested. (And learned about `${PROJECT_VERSION}` …)

The generated `ZXingConfigVersion.cmake` file is exactly the same after this change and the tests results are as in the original PR.

axxel:
Thanks!

PR #140, opened 2020-07-09
Fix include syntax

gbsf:
Recent changes broke the build downstream.
```
In file included from […]/src/kitinerary/src/qimagepurebinarizer.cpp:10:
[…]/include/ZXing/BitArray.h:30:10: error: 'ZXContainerAlgorithms.h' file not found with <angled> include; use "quotes" instead
#include <ZXContainerAlgorithms.h>
         ^
```

### DIFF
--- a/core/src/BitArray.h
+++ b/core/src/BitArray.h
@@ -18,6 +18,7 @@
 */
 
 #include "ZXConfig.h"
+#include "ZXContainerAlgorithms.h"
 #ifndef ZX_FAST_BIT_STORAGE
 #include "BitHacks.h"
 #endif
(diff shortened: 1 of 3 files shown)


axxel:
Ups, my bad. Do you know any 'trick' to make a standard local trick to complain about the same issue? I assume the issue arises if the project including the headers is not a sub-project of the master CMakeLists.txt?

gbsf:
I don't know about "standard", but you could have a standalone CMake project to test an imported target as you say.

However, I think that the root cause is that you install the headers from `core/src` to `$PREFIX/include/ZXing`, but export `$PREFIX/include` [from CMake](https://github.com/nu-book/zxing-cpp/blob/master/core/CMakeLists.txt#L449) instead.

The cleanest solution for me would be to separate the exported headers into a `include/ZXing` directory, and standardize into using `<ZXing/header.h>` everywhere.

Issue #127, opened 2020-06-01
Retire travis CI builds?


axxel:
I just added github actions based CI support to get automatic build testing for macOS and VisualStudio (which, uncovered 4 more issues related to missing/wrong includes, yay :-)).

From my perspective, github actions are a superset of what travis ci brings us. And running the travis scripts seems like a waste of energy and someone elses money.

axxel:
Travis now runs ubuntu 20.04 and therefore can build + test the whole project again. I still believe it is a waste of resources.

axxel:
As of #161 the travis CI script now provides us with a test on the POWER architecture which github actions apparently does not provide at the moment. So this stays for the time being.

PR #151, opened 2020-08-16
Windows builds

huycn:


wrappers/gdiplus/CMakeLists.txt
```diff
@@ -16,18 +16,19 @@ if (${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
...
 
-if (NOT DEFINED BUILD_READERS)
-    set (BUILD_READERS ON)
-endif()
+    if (NOT DEFINED BUILD_READERS)
+        set (BUILD_READERS ON)
+    endif()
 
-add_subdirectory (${CMAKE_CURRENT_SOURCE_DIR}/../../core ${CMAKE_BINARY_DIR}/ZXing)
+    add_subdirectory (${CMAKE_CURRENT_SOURCE_DIR}/../../core ${CMAKE_BINARY_DIR}/ZXing)
```
axxel, line 28: This would also need be to changed to link against the core lib instead of including it as source here. Similar to the python wrapper. That would be the 'right' way to do it, in my opinion ;).

CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ option (BUILD_EXAMPLES "Build the example barcode reader/writer applicatons" ON)
 option (BUILD_BLACKBOX_TESTS "Build the black box reader/writer tests" ON)
 option (BUILD_UNIT_TESTS "Build the unit tests (don't enable for production builds)" OFF)
 option (BUILD_PYTHON_MODULE "Build the python module" OFF)
+option (BUILD_GDIPLUS_LIB "Build the GDI+ wrapper as static library" OFF)
```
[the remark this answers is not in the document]
huycn: Good idea

wrappers/gdiplus/CMakeLists.txt
huycn: Agreed, you are right.

huycn: Oups, that doesn't work. I got: 
CMake Error at core/CMakeLists.txt:438 (add_library):
  add_library cannot create target "ZXing" because another target with the
  same name already exists.  The existing target is a static library created
  in source directory "D:/Repo/zxing-cpp/core".  See documentation for
  policy CMP0002 for more details.

which seems logical since the master CMakeLists.txt already include the 'core' module.

axxel: I think you read my comment as I intended it but my comment was quite the opposite of 'right'. I only saw the diff and did not realize it was inside this "top level project"-guard. So with the top-level build option for the wrapper inside a WIN32 block and defaulted to `ON` this compiles now on windows? -> merge?

huycn:
@axxel I made the adjustment to build gdiplus library by default on Windows. I also added a workflow for building WinRT package as separated from CI. I'm going to make it build on release or manual rather than any push into master.

wrappers/winrt/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ add_definitions (-DUNICODE -D_UNICODE)
 
 if (MSVC)
     add_definitions (
-        -DNOMINMAX
+        -DNOMINMAX -wd4996
```
axxel, line 22: That is hopefully not there to simply silence the deprecation warnings that I recently introduced? ;)

axxel (approved):
Thanks for asking :).

wrappers/winrt/CMakeLists.txt
huycn: Yes it is. It's necessary since the project is configured to flag warnings as build errors.

axxel: Well, I'd suggest to rather change that setting then to allow for builds with warnings. But this win-rt stuff is definitively not my area of expertise :).

huycn:
Thanks for your helpful comments :)

.appveyor.yml
```diff
@@ -42,7 +42,8 @@ build_script:
     %CMAKE_EXE% --build . --config Release
 
     cd %BASE_DIR%
-    nuget pack -Version %APPVEYOR_REPO_BRANCH:~1% wrappers\winrt\nuget\ZXingWinRT.nuspec
+    for /f "tokens=1,2,3 delims=. " %%a in ("%APPVEYOR_BUILD_VERSION%") do set major=%%a&set minor=%%b&set patch=%%c&set
+    nuget pack -Version %major%.%minor%.%patch% wrappers\winrt\nuget\ZXingWinRT.nuspec
```
axxel, line 46: Just for my understanding: this AppVeyor setup is now somewhat obsolete with the Actions artifacts building you added below? See also #127?

huycn: Yes, it's not obsolete. Same for travis.yml. I will remove them later on.

wrappers/winrt/CMakeLists.txt
huycn: It's ok if we got few warnings to remind us things to be done. But the real issue with leaving warnings is that it's easy get ignored and get accumulated with the time. And when it becomes too many, it becomes serious problem but no one wants to fix ;)

PR #154, opened 2020-09-04
fix: transpose (width, height) in python wrapper

dimatura:
The dimensions for the output matrix in the python barcode writer were transposed, causing segfaults in non-square matrices.
This PR transposes the dimensions and adds a unit test to check output dimensions.


axxel:
Nice. Thanks!

PR #161, opened 2020-10-07
Update .travis.yml

srinivas32:
Added power support for the travis.yml file with ppc64le. This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing. For more info tag @gerrith3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+arch:
+ - amd64
+ - ppc64le
 language: cpp
 dist: focal
 os:


axxel:
Thanks. Please note #127. If you had something similar based on github actions, that would be nice.

gerrith3:
@axxel  Unfortunately there is not currently Power hardware in Azure.  We have been working on some integration with Actions and using Power hardware in our cloud but we haven't had the bandwidth to get this up and running as yet.  You can run GHA for Intel and Travis for Power/MacOS currently without any real issue, other than the annoyance of two build script models.

axxel:
@srinivas32 @gerrith3 I only now noticed by chance that travis-ci.org shut down last year. Any update on the ppc64le front inside GHA? Or another alternative to keep this architecture tested? (I'm about to release 1.3.0 and thought a test run on ppc64le would be really nice before I do that...)

PR #163, opened 2020-10-09
Fix compilation errors with -Werror=undef and add it to the flags.

dfaure-kdab:
This allows to detect #if/#ifdef confusions.
I have it always on, which is how I stumbled upon this.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -25,7 +25,7 @@ if (MSVC)
     )
 else()
     set (ZXING_CORE_LOCAL_DEFINES ${ZXING_CORE_LOCAL_DEFINES}
-        -Wall -Wextra -Wno-missing-braces)
+        -Wall -Wextra -Wno-missing-braces -Werror=undef)
 endif()
 
 
--- a/core/src/GenericLuminanceSource.cpp
+++ b/core/src/GenericLuminanceSource.cpp
@@ -15,7 +15,7 @@
 * limitations under the License.
 */
 
-#if (_MSC_VER >= 1915)
+#if defined(_MSC_VER) && (_MSC_VER >= 1915)
 #pragma warning(disable : 4996)
 #endif
 
--- a/core/src/datamatrix/DMSymbolInfo.cpp
+++ b/core/src/datamatrix/DMSymbolInfo.cpp
@@ -65,7 +65,7 @@ static const SymbolInfo PROD_SYMBOLS[] = {
 static const SymbolInfo* s_symbols = PROD_SYMBOLS;
 static ZXING_IF_NOT_TEST(const) size_t s_symbolCount = Size(PROD_SYMBOLS);
 
-#if ZXING_BUILD_FOR_TEST
+#ifdef ZXING_BUILD_FOR_TEST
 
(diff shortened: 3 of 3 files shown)


CMakeLists.txt
```diff
@@ -27,6 +27,8 @@ if (MSVC)
         set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} /MTd")
         set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /MTd")
     endif()
+ else()
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror=undef")
```
axxel, line 31: Thanks for the PR. This line breaks the build, though. As you can see, gtest seems riddled with these. Also manipulating CXX_FLAGS is generally considered bad practice. Could you please add this to the `ZXING_CORE_LOCAL_DEFINES` in core/CMakeLists.txt and we see how far we get then?

axxel (changes requested):

CMakeLists.txt
dfaure-kdab: Ah, I hadn't noticed because unittests aren't built by default, and the name of the cmake option for building them is non-standard. It's `BUILD_UNIT_TESTS` while CMake standardizes the name `BUILD_TESTING` (see `Modules/CTest.cmake`).
Would it be OK for me to make a separate MR about changing that option name to use the standard one?
Then my build script would have enabled the tests automatically and I would have noticed this :)
But of course this would require updating existing scripts and CI on your side...

dfaure-kdab: Well, on the other hand, if I do that, I'll then hit the `-Werror=undef` in gtest itself, lol.... ;)

axxel: The unit test option is not enabled by default because there are symbols exported to enable the unit testing which are not meant to be visible in a production build. (Might be a questionable approach but that is what it looks right now.)

And `BUILD_TESTING` is not a canonical name either, since we have two test sets: 'black box' tests that operate on example images and 'white box' tests, which are the gtest based 'unit tests' in this case.

axxel:
Thanks!

axxel:
Disclamer: I just noticed that this `-Werror` flag also breaks my local build because of some SIMD library I've been experimenting with. So this is likely going to be removed again, at some point in time... ;)

CMakeLists.txt
dfaure-kdab: I completely agree that unittests shouldn't be enabled by default in production code.
Some projects enable them by default when selecting Debug mode. But no problem.
Exporting more symbols when unittests are enabled makes sense as well, e.g. Qt does that too.

Blackbox tests seem to be enabled by default, so maybe `BUILD_TESTING` could be used for the unittests...

dfaure-kdab:
> Disclamer: I just noticed that this `-Werror` flag also breaks my local build because of some SIMD library I've been experimenting with. So this is likely going to be removed again, at some point in time... ;)

Or you could submit an upstream patch to fix -Wundef errors, just like I did here :-)

PR #164, opened 2020-10-09
Expose position in wasm

Summon528:


wrappers/wasm/BarcodeReader.cpp
```diff
@@ -96,8 +97,21 @@ EMSCRIPTEN_BINDINGS(BarcodeReader)
...
+            .field("position", &ReadResult::position)
 	        ;
 
+    value_object<ZXing::PointI>("Point")
+	        .field("x", &ZXing::PointI::x)
+	        .field("y", &ZXing::PointI::y)
+	        ;
+
+    value_object<ZXing::Position>("Position")
+            .field("topLeft", emscripten::index<0>())
```
Summon528, line 109: emscripten complains if I set this to `.field("topLeft", ZXing::Position::topLeft)`
```
error: pointer to function type 'emscripten::internal::GenericBindingType<ZXing::PointT<int> () const noexcept>::ActualT' (aka 'ZXing::PointT<int> () const noexcept') cannot have 'const' qualifier
            typedef ActualT* WireType;
                           ^
```

wrappers/wasm/CMakeLists.txt
```diff
@@ -11,13 +11,11 @@ if (NOT CMAKE_BUILD_TYPE)
     set (CMAKE_BUILD_TYPE "MinSizeRel" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
-add_definitions ("-s DISABLE_EXCEPTION_CATCHING=0")
```
Summon528, line 14: This seems like a duplicate?

wrappers/wasm/CMakeLists.txt
```diff
@@ -11,13 +11,11 @@ if (NOT CMAKE_BUILD_TYPE)
...
-
 add_subdirectory (${CMAKE_CURRENT_SOURCE_DIR}/../../core ${CMAKE_BINARY_DIR}/ZXing)
 
 include_directories ("../../thirdparty/stb")
 
-set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} --bind -s ENVIRONMENT=web -s DISABLE_EXCEPTION_CATCHING=0 -s FILESYSTEM=0 -s MODULARIZE=1 -s EXPORT_NAME=ZXing")
+set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} --bind -s ENVIRONMENT=web -s DISABLE_EXCEPTION_CATCHING=0 -s FILESYSTEM=0 -s MODULARIZE=1 -s EXPORT_NAME=ZXing -s ALLOW_MEMORY_GROWTH")
```
Summon528, line 18: WASM throws OOM error after the above changes, I wonder if I did something wrong or did we just barely have enough memory?

Summon528: The error msg:

> Uncaught RuntimeError: abort(Cannot enlarge memory arrays to size 18034688 bytes (OOM). Either (1) compile with  -s INITIAL_MEMORY=X  with X higher than the current value 16777216, (2) compile with  -s ALLOW_MEMORY_GROWTH=1  which allows increasing the size at runtime, or (3) if you want malloc to return NULL (0) instead of this abort, compile with  -s ABORTING_MALLOC=0 ) at Error

wrappers/wasm/BarcodeReader.cpp
axxel: `ZXing::Position::topLeft` is a member function. I guess you need to specify that differently.

wrappers/wasm/BarcodeReader.cpp
```diff
@@ -96,8 +97,21 @@ EMSCRIPTEN_BINDINGS(BarcodeReader)
...
+	        .field("topRight", emscripten::index<1>())
+	        .field("bottomLeft", emscripten::index<2>())
+            .field("bottomRight", emscripten::index<3>())
+            ;
+
```
axxel, line 114: You can do this this way (I guess) but be careful of the order: the points are stored in clockwise order. Also, please make sure the whitespace style is consistent with the rest (use tabs with indent level 4 as leading whitespace).

wrappers/wasm/CMakeLists.txt
axxel: No, this affecting compilation, the other setting only linking.

wrappers/wasm/CMakeLists.txt
axxel: If you get an error after the change, why did you make the change in the first place?

axxel (changes requested):

wrappers/wasm/CMakeLists.txt
Summon528: I mean that it errors after I expose `position` and adding `ALLOW_MEMORY_GROWTH` fixes it

axxel: I see. That is indeed weird. You get the error even before you try to decode an image? If it happens during image decoding, then it certainly has to do with how large your image is and you might indeed just have hit the threshold just by adding those few lines? Unlikely things happen, I guess :).

axxel:
General question: do you have some simple code that actually uses this new property that you could add to the `demo_reader.html`? That would be nice.

Summon528:
Updated, example screenshot:
image

axxel:
Awesome. Thank you!

PR #166, opened 2020-10-19
Fix Android compilation failure: no need for external threads lib

tanius:
The line `find_package(Threads REQUIRED)` fails when cross-compiling for Android. In earlier versions where the Android build succeeded, this line had been `find_package(Threads)`, which only caused a non-fatal error.

Under Android, the libc standard library (called 'bionic') includes built-in support for pthreads ([see](https://stackoverflow.com/a/64432754)), so no external package has to be found and no threads library linked to when compiling for Android. So to solve the build issue, the PR excludes these actions when building for Android.

With the PR, builds under Android (requiring API level 16+ here) succeed again, and my application [foodrescue-app](https://github.com/fairdirect/foodrescue-app) that uses ZXing under Android works fine with the resulting library. (I did not try the official ZXing-C++ example applications under Android though, due to an unrelated build problem that is independent of the thread lib issue here and was present in earlier versions of ZXing.)

pthreads support is not complete in Android. But the pthreads functions that exist are POSIX compliant, and any function implemented in a non-POSIX-compliant way has a `_np` suffix in the name ([details](https://stackoverflow.com/a/58348085)). Since building ZXing succeeds, it means that all required pthreads functions are found in a POSIX-compliant implementation in the Android libc, and ZXing usage under Android is expected to be unaffected by the state of the pthreads implementation under Android.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -429,7 +429,9 @@ source_group (Sources\\textcodec FILES ${TEXT_CODEC_FILES})
 
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 set(THREADS_PREFER_PTHREAD_FLAG TRUE)
-find_package(Threads REQUIRED)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Android")
+    find_package(Threads REQUIRED) # Not needed on Android, where pthread.h is part of the standard library.
+endif()
 
 add_library (ZXing
     ${COMMON_FILES}
@@ -465,7 +467,9 @@ endif()
 # the lib needs a c++-17 compiler but can be used with a c++-11 compiler (see examples)
 target_compile_features(ZXing PRIVATE cxx_std_17 INTERFACE cxx_std_11)
 
-target_link_libraries (ZXing PRIVATE Threads::Threads)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Android")
+    target_link_libraries (ZXing PRIVATE Threads::Threads) # On Android, this is in the standard lib.
+endif()
 
 add_library(ZXing::ZXing ALIAS ZXing)
 # add the old alias as well, to keep old clients compiling


axxel:
Thanks for the PR and the detailed explanation. I have a question, though: would it compile if we just dropped the `REQUIRED` keyword?

The thing is, that this seems to me to be either a misuse of cmake by me (adding `REQUIRED`) or a bug in the cmake `threads` module. I have not seen the idiom "if not android then require threads" used much in my quick google search, so I wander whether that is best practice?

Any input from @huycn  or @vkrause (the only two other active android users known to me)?

tanius:
Only dropping the `REQUIRED` keyword from the current code does not yet let it build under Android, because then the linker step fails:

```
find_package(Threads)
target_link_libraries (ZXing PRIVATE Threads::Threads)
```

What did work for building under Android was the version before 229d335, where the linker was effectively given an empty variable under Android:

```
find_package(Threads)
target_link_libraries (ZXing PUBLIC ${CMAKE_THREAD_LIBS_INIT})
```

About CMake best practices I can't comment (still learning …). The `find_package(Threads REQUIRED)` seems logically correct to me though, because under all OSs except Android a threads library is indeed a non-optional dependency.

vkrause:
We hit this as well, but ended up working around it on our side, as other components were showing the same issue: https://invent.kde.org/frameworks/extra-cmake-modules/-/merge_requests/31

axxel:
Thanks for chiming in. Your workaround is pretty much what I would have expected from cmake directly (as in 'just make it work, please'). What I currently don't see is whether this PR would break your build again? You seem to have come to the conclusion that you need `-pthread` during compilation and linking on android for libs depending on `Threads::Threads`. But this PR would remove that dependency for zxing again.

vkrause:
I've now tested the changes from this PR here, seems to cause no problems in our setup.

tanius:
I'm also using ECM, so the fix mentioned by vkrause above would work for me.

The (only?) established way to compile for Android using CMake _without_ ECM is with Android Studio. So if their CMake toolchain file would also include a solution equivalent to [ECM PR #31](https://invent.kde.org/frameworks/extra-cmake-modules/-/merge_requests/31), there would be no need to change anything here in ZXing. (But I doubt that's the case, since ECM depends on the Android NDK CMake toolchain file and still had to fix this issue.)

axxel:
@tanius Does that mean your build issue is in-fact fixed via the ECM patch right now?

tanius:
Yes, indeed my build issue is fixed via the ECM patch from their [PR #31](https://invent.kde.org/frameworks/extra-cmake-modules/-/merge_requests/31). Just tested it all, and the build succeeds when using ECM v5.75.0, which is the first release that includes this patch. Welcome to close this PR, then. :slightly_smiling_face: 

I see that the [current Android build instructions](https://github.com/nu-book/zxing-cpp#android-ndk) seem to not use CMake, which means that the KDE people and me use a rather non-standard build process here anyway. But if you want to add documentation for our CMake-and-ECM Android build process as well, in short this is what I do right now:

1. Build and install CMake 3.18 or newer. This is needed by ECM 5.75.0 because it uses [`list(SORT … COMPARE NATURAL …)`](https://cmake.org/cmake/help/v3.18/command/list.html#ordering). Sadly, CMake 3.18 is not yet available in any Debian / Ubuntu distro, so in that case has to be compiled from source. I used [these instructions](https://askubuntu.com/a/865294).

2. Build and install [ECM 5.75.0](https://invent.kde.org/frameworks/extra-cmake-modules/-/tags/v5.75.0) or newer.

3. Install Android NDK 18.1.5063045 and the corresponding Android build tools, Android platform, and platform tools packages. The process may also work with NDK ≥19, but not yet confirmed.

4. Run CMake for the Android build of ZXing-C++, with variables and options adapted according to your system:

    ```
    mkdir -p build/android_armv7 && cd build/android_armv7
    
    export ANDROID_SDK_ROOT=/opt/android-sdk
    export ANDROID_NDK=/opt/android-sdk/ndk/18.1.5063045
    export ANDROID_ARCH_ABI=armeabi-v7a

    cmake ../.. \
      -DCMAKE_TOOLCHAIN_FILE=/usr/local/share/ECM/toolchain/Android.cmake \
      -DANDROID_SDK_BUILD_TOOLS_REVISION=28.0.2 \
      -DBUILD_EXAMPLES=OFF \
      -DBUILD_BLACKBOX_TESTS=OFF \
      -DCMAKE_INSTALL_PREFIX=$(readlink -f ../../relative/path/to/your/install/dir)
    ```

5. Build and install:

    ```
    make
    make install
    ```

tanius:
Closing my PR, as it is no longer relevant for an (ECM based) Android build process when using the latest ECM version, as explained above. I still think it would be good to document that version requirement and build process, along the lines of what I wrote in the comment just above. If there's interest in integrating that kind of documentation, I can also make it a PR for the README.

Issue #153, opened 2020-08-24
Android anyone?


axxel:
I looked into adding an android build to the CI system and noticed two things:

1. I seem to have broken the format specification feature of the android wrapper a while ago but no-one complained. Is there anyone actively using the android wrapper?

2. Nowadays the way to build external C++ libraries for android seems to be cmake with toolchain files instead of via ndk-build `*.mk` files. I see a benefit in removing the duplicated information in and currently outdated `*.mk` files. Anyone relying on those?

@vkrause?

vkrause:
We are using the normal C++ API and the CMake build system on Android as well, ie. exactly the same as we are using this on all other platforms. So no objection from our side regarding removing any of those.

axxel:
@vkrause thanks for the prompt reply.

Question: what do you mean you use the "normal C++ API"? You have your own JNI wrapper? Or is there another trick to get access to zxing from a JVM-based app?

vkrause:
The part of our application using zxing is written in C++, so we don't even need to access this from Java at all.

huycn:
It seems that WinRT wrapper is also broken with the change in format. I will fix both of them.
Build with CMake is a good idea. I remember I've tried quickly to do that but I was rather complicated than I thought with the structure of CMakeLists.txt that we had. As now it has changed, I will try again.

axxel:
I see you updated the ndk-build files instead of removing them. Did the "switch to cmake" attempt fail?

huycn:
Yes, my first attempt failed but I will try again. In the mean time, I pushed the fix for BarcodeFormat first.

axxel:
@huycn I've been working on the android plugin lately and I have a `cmake` + AndroidStudio/gradle based setup that builds the jni library and a new Kotlin wrapper and also includes a demo app that can be used to actually run the code on a phone and play with it a little (like setting the tryRotate flag or cropping the viewport, etc.) and also includes a button to switch over to the upstream Java code for quick runtime and detection quality comparisons. I intend to use that to replace the current *.mk build setup but wanted to know if relying on AndroidStudio/gradle would be a hassle for you? It would still be possible to call `cmake` manually but I would need to find out how and document it.

huycn:
@axxel I will be no problems for us, and I don't think there are people out there develop for Android without Android SDK :)
So your changes will be welcome.

axxel:
[...]

PR #169, opened 2020-11-12
Fix android build - Android.mk

ph4r05:
Fixes Android `ndk-build` by adding missing files and removing non-existent ones from Android.mk

### DIFF
--- a/core/Android.mk
+++ b/core/Android.mk
@@ -18,6 +18,7 @@ COMMON_FILES :=	\
 	src/BitMatrix.cpp \
 	src/BitSource.cpp \
 	src/CharacterSetECI.cpp \
+	src/ConcentricFinder.cpp \
 	src/DecodeHints.cpp \
 	src/DecodeStatus.cpp \
 	src/GenericGF.cpp \
@@ -70,7 +71,7 @@ ONED_FILES := \
 	src/oned/ODMultiUPCEANReader.cpp \
 	src/oned/ODReader.cpp \
 	src/oned/ODRowReader.cpp \
-    src/oned/ODUPCEANCommon.cpp
+	src/oned/ODUPCEANCommon.cpp \
 	src/oned/rss/ODRSSExpandedBinaryDecoder.cpp \
 	src/oned/rss/ODRSSFieldParser.cpp \
 	src/oned/rss/ODRSSGenericAppIdDecoder.cpp
@@ -89,17 +90,12 @@ PDF417_FILES := \
 	src/pdf417/PDFScanningDecoder.cpp
 
 QRCODE_FILES := \
-	src/qrcode/QRAlignmentPattern.cpp \
-	src/qrcode/QRAlignmentPatternFinder.cpp \
 	src/qrcode/QRBitMatrixParser.cpp \
 	src/qrcode/QRCodecMode.cpp \
 	src/qrcode/QRDataBlock.cpp \
-	src/qrcode/QRDataMask.cpp \
 	src/qrcode/QRDecoder.cpp \
 	src/qrcode/QRDetector.cpp \
 	src/qrcode/QRErrorCorrectionLevel.cpp \
-	src/qrcode/QRFinderPattern.cpp \
-	src/qrcode/QRFinderPatternFinder.cpp \
 	src/qrcode/QRFormatInformation.cpp \
 	src/qrcode/QRReader.cpp \
 	src/qrcode/QRVersion.cpp
--- a/wrappers/android/java/com/zxing/BarcodeReader.java
+++ b/wrappers/android/java/com/zxing/BarcodeReader.java
@@ -89,7 +89,7 @@ public class BarcodeReader
 	private static native int readBarcode(long objPtr, Bitmap bitmap, int left, int top, int width, int height, Object[] result);
 
 	static {
-		System.loadLibrary("zxing-android");
+		System.loadLibrary("zxing_android");
 	}
 
 }


axxel:
Thanks for the effort. I tried to keep that in sync but miserably failed. I have two comments:
1) you added the writer-part for the QRCode. do you need that? That change would be inconsistent, because for the other symbologies the writing part is also not included in the android ndk build.
2) @huycn was looking into replacing this ndk build approach with a cmake approach, exactly to spare us keeping those build definitions in sync. See #153. I assume he did not find the time, yet. Maybe you could give that a go?

ph4r05:
@axxel thanks for feedback! You are right, I don't need the Writer (fixed).  

ad cmake approach - I check that issue also and I was wondering how this went. I am a bit busy now but I may look into it later as I would prefer cmake also. 

I've also fixed `System.loadLibrary("zxing_android");`, the module is now with the underscore.

axxel:
* About the writer part: would you please also remove the addition of the 3 files `QREncoder.cpp, QRMaskUtil.cpp, QRMatrixUtil.cpp`? Thanks in advance.

* ndk-build test: very good idea. at least until we have a working cmake based build, then I'd prefer to drop the 'mk' file

* speed: thanks for the positive feedback! I was spending a considerable amount of my time on performance improvements, so it is very nice to hear someone noticed/takes a advantage of it :). Question: if you are only interested in QRCode, have you set the `formats` property to only include QRCodes? That will potentially affect your frame rate significantly.

ph4r05:
@axxel aah I didn't notice the other 3 files, removed now. 

- speed: I tried to specify QRcode only initially, so the results are with this setting. Also, the scanning robustness is quite impressive! Variable distance, rotations, skews, and so on. Just for comparison:
  - The second fastest libs were *BoofCV* and *QUIRC* with 40ms with the same QR code (you are 2x faster! wow), but QUIRC was quite sensitive to distance/picture quality, rotations. 
  - *ZBar* was 90ms, also sensitive
  - *OpenCV* 100-200ms, crashes with larger QRcodes. 

Thanks again for a great work!

axxel:
Thanks! Especially for the detailed speed comparison. Much appreciated. :). I have some SIMD code experiments lying around. They should result in even better numbers, if you can use the NEON instruction set on your phone. I'll try to remember get back to you once I have that.

axxel:
Two more questions came to my mind:
1) your tests were done with current head, not with the v1.1.1 release, right?
2) you would not happen to have a speed comparison with the upstream Java version of the library? ;)

ph4r05:
sure no problem @axxel:

1. I've tested master of this lib, 0145f517db4a04d5d70beb0c34baeceb071877a1
2. I've added also the Java lib test, `com.google.zxing:core:3.4.1`:

Here are the new testing results. I've added moving exp. average with alpha=0.01:

- QUIRC https://github.com/dlbeer/quirc/commit/7e7ab596e4d0988faf1c12ae89c354b114181c40: 30.621 ms,
- ZXing-cpp 0145f517db4a04d5d70beb0c34baeceb071877a1: 15.828 ms
- com.google.zxing:core:3.4.1: 12.122 ms
- BoofCV 0.36.1: 29.624 ms
- ZBar 0.2.0: 84.280 ms (quite old, I will update this one)

I am measuring all at once (sequentially per frame) to eliminate potential throttling effects (CPU temperature, cam temp).
Only recognition is measured, e.g., input preparation to the decode function is excluded.  

ZXing-cpp seems more resistant to the less-quality picture. 
I've rechecked and I pass android Bitmap RGB_888 to the ZXing-cpp while for ZXing I am constructing BinaryBitmap.

ph4r05:
Later I measured also the detection success rate (moving avg same alpha), found a position where all libraries were able to detect the presented QR code, waited until the success rate climbs to 0.99:

- QUIRC: 57.767 ms
- ZXing-cpp: 26.724 ms
- com.google.zxing:core: 19.686 ms
- BoofCV: 55.739 ms
- ZBar: 162.398 ms

As you can see the throttling is already active due to increasing temperature but it relatively still matches I think.

Speed when no QR code is present:
- QUIRC: 19.121 ms
- ZXing-cpp: 15.822 ms
- com.google.zxing:core: 10.268 ms
- BoofCV: 23.937 ms
- ZBar: 67.384 ms

axxel:
Thanks for the feedback and the work. Your results are surprising, since I know the Java code is structural equivalent to what ZXingCpp started out with and I made it measurably faster (on my development linux box). Plus I expected compiled c++ code to be at least a little bit faster than Java byte code.

> I've rechecked and I pass android Bitmap RGB_888 to the ZXing-cpp while for ZXing I am constructing BinaryBitmap.

Maybe this is the reason for the speed improvement. Where do you get the BinaryBitmap from / how do you generate it? How do you get the RGB_888 data? I experimented with `QVideoFilter` in one of the examples and there it seemed pretty normal to get video content passed as YUV data.

As that stands, you would obviously use the Java version, right?

ph4r05:
I am also a bit surprised by the timing differences. I also expected cpp version to surpass java. I am using Android 10, arm-x64, maybe the AAR code is quite optimized as well?

BinaryBitmap I am passing to the ZXing-java is constructed in the following way (in kotlin):

```kotlin
val mt = Mat()  // OpenCV Matrix, grayscale. The matrix is input as I use it primarily in the existing project. 
val intArray = IntArray(mt.width() * mt.height())
bmp.getPixels(intArray, 0, bmp.width, 0, 0, bmp.width, bmp.height)
val source = RGBLuminanceSource(bmp.width, bmp.height, intArray)
val bitmap = BinaryBitmap(HybridBinarizer(source))
```

This construction is not included in the measurement. 

I would prefer to use CPP as I want it usable also for iOS and WebAsm. 

Another idea: I managed to speed up other optical recognition code by using these flags compile: 

```
-O3 -ffast-math -funsafe-math-optimizations -fno-math-errno
```

I will try it and if I measure different timing I let you know.

ph4r05:
@axxel so it helped!

ZXing-cpp 0145f51 now runs around 8 ms while java version is 16 ms! Should we add these flags to the .mk files?

axxel:
The effect of your compiler flags is quite surprising as well, I would have expected at most a single digit improvement, not a 2x speedup. These flags are likely to affect the accuracy and the blackbox tests. But given those effects, that might very well be worth it. I'll give them a go on my development machine but they will definitively not have this impact on an Intel CPU.

I think another source of error is your measuring strategy: If you keep the `RGBLuminanceSource` and `BinaryBitmap` creation outside of your measurement but (implicitly) include it in the c++ code path you are comparing apples to oranges. I'm unfamiliar with the Android/OpenCV APIs but I would recommend to stay away from RGB data if you start out with greyscale.

ph4r05:
Yep, the accuracy is not a big deal with error-correcting codes. I didn't observe any success rate drop with these flags. 

If I understand it correctly, I have to pass android Bitmap to the android binding of the ZXing-cpp (as there is only `readBarcode` native function which takes Bitmap). 

Both java and cpp tests are agnostic to the opencv image acquisition, the camera is capturing 1088x1088 RGB image. For recognizer supporting passing grayscale directly, I am converting it to the grayscale before starting the measurement. 

I am currently not aware of method for transforming color bitmap to grayscale with `Bitmap.createBitmap(this.cols(), this.rows(), Bitmap.Config.ARGB_8888)` as `Bitmap.Config` have only RGB constants. 

From the experience, color to grayscale transformation is very fast, usually under a millisecond so I don't expect it to change results significantly.

ph4r05:
Update: so it is enough to keep `-O3` flag to achieve 8 ms running time. That's good news.

axxel:
Another important difference just came to my mind: the Java code defaults to `setTryHarder(false)` while I decided to default that to `true`. That makes a huge difference. Would you please try again with setting it to false on the c++ side?

And as I expected: those flags break a significant number of test cases on my Intel machine (and bring no measurable performance improvement). `-O3` does nothing on my Intel over `-O2` but that is very good news on the ARM side. I will change the default in the `.mk` file.

Yes, the RGB -> greyscale change is not significant. But the `HybridBinarizer + BinaryBitmap` part is.

Looks like we are getting closer to what is really going on here :).

Issue #170, opened 2020-11-19
DataMatrix DMDetetor: traceGap() assertion may fail in noisy area


neycyanshi:
If the image is noisy after binarization, `traceGap()` assertion may fail on this line.
https://github.com/nu-book/zxing-cpp/blob/b93fdeef39342a4af66c5c62a6b28d05379c9772/core/src/datamatrix/DMDetector.cpp#L554
Because there are actually two operations in the main while loop that moves EdgeTracer's current point `p`, and it may stay unmoved. `Project back + centered()` and `traceStep()` happens to move the point in exactly the opposite direction. https://github.com/nu-book/zxing-cpp/blob/b93fdeef39342a4af66c5c62a6b28d05379c9772/core/src/datamatrix/DMDetector.cpp#L576 and https://github.com/nu-book/zxing-cpp/blob/b93fdeef39342a4af66c5c62a6b28d05379c9772/core/src/datamatrix/DMDetector.cpp#L607

axxel:
Could you provide a sample image that triggers this issue?

neycyanshi:
I found one assertion failure in one image in my large dataset. However, I have modified the binarizer and use image pyramid. So the image may or may not trigger failure in the original code. I could provide the binarized BitMatrix image and the tracer state.

axxel:
I need something to reproduce the error on my machine. You should be able to dump your BitMatrix as a b/w image so that I can use that with the code as.

Am I right to assume you are employing an image pyramid to improve the decoding of high resolution images?

neycyanshi:
That's right, pyramid is to improve decoding of high-res image. I am trying to reproduce on your `master` branch. Also, the image contains private info, so I have to remove them. But you can see these lines may have some problem.

neycyanshi:
Fortunately I have reproduced it on the binarized image. Here is the image that cause assertion failure. Manually set try_harder = true in DMDetector.
![Sample_TraceGap](https://user-images.githubusercontent.com/16794154/99758375-7798f380-2b2c-11eb-9095-bca433717eff.png)

neycyanshi:
First set tryHarder = true in DMDetector.cpp Line 779, to enable multi-line scan to detect off-center symbols.

Entry arguments:
ZXing:BitMatrixCursorF = {
    img = {
      object_ = 0x0000000101291bf0
    }
    p = (x = 283.5, y = 92.5)
    d = (x = -1, y = 0.02175528874599993)
…
}
[...]

PR #171, opened 2020-11-20
Remove assertion in traceGaps

neycyanshi:
Remove assertion in traceGaps because it may fail if back-projected to the last position. #170


axxel:
Thanks, used 2 slightly different lines, though.

Issue #172, opened 2020-11-20
Support for inverted symbols (white on black)


axxel:
As @neycyanshi correctly pointed out in #167 inverted codes are not handled inside ZXing at the moment (for none of the formats). It is rather trivial for a client application that needs that feature to implement it outside of ZXing by calling `ReadBarcode` twice and the second time simply invert the input image. There are more performant possibilities if done inside and specific to the different implementation characteristics.

A quick and dirty solution to get the functionality into ZXing would be to do the inversion inside `ReadBarcode` based on a new `DecodingHints` flag.

overriden-sfdd:
I'm not sure if I should post it in this thread, but I have a sample with a strange color scheme. Back is light brown and the text is black. I know that we can use thresholding with the big threshold value. Is there a way to decode such case not relying on heuristic preprocessing?
![2](https://user-images.githubusercontent.com/47534862/126989498-28e71ddc-4655-4273-a2b3-37a29ee415a1.jpg)

axxel:
@overriden-sfdd this has nothing to do with inversion (white on black) but shows another limitation of the `GlobalHistogramBinarizer`. Please delete this comment and post it as a new issue. Suggested Title: 'Improve 1D barcode detection with noisy input'.

axxel:
Triggered by #387 I thought a little more about this and came to the conclusion that I don't know enough about the use case that is supposed to be supported here. The naive approach is obviously the above: try everything with and without inverted input. That is rather expensive in terms of performance.

What I am questioning is the use-case of inverted linear barcodes. It seems to me that those are simply non-existent in terms of the standards? Inverted QR Codes are obviously a thing and so is inverted DataMatrix (#173). I would suggest to limit this feature to symbologies that actually support that and hence occur in the wild and not waste implementation and processing time on stuff that is never used.

If someone has an inverted input due to technical reasons and everything needs to be inverted back per se, then they shall fix that outside of the lib in my opinion.

May I pull in @gitlost and @oehhar s expertise once again? Anyone else has a stake in the matter?

gitlost:
The symbologies known to me that support "reflectance reversal" are Aztec Code, Data Matrix, DotCode, Han Xin Code, Grid Matrix and QR Code (including Micro QR Code & rMQR), so i.e. matrix symbols, no linear or stacked symbols.

[...]

PR #173, opened 2020-11-26
DMDetector: support for inverted DataMatrix

neycyanshi:
Add support for invert DataMatrix detection. Subtask of #172

### DIFF
--- a/core/src/BitMatrixCursor.h
+++ b/core/src/BitMatrixCursor.h
@@ -61,6 +61,9 @@ public:
 		bool operator==(Value o) const { return v == o.v; }
 		bool operator!=(Value o) const { return v != o.v; }
 	};
+    
+    Value c1{true};  // On module color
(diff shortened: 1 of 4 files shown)


axxel:
Thanks for the PR but I don't like the idea of having everyone to pay for this in runtime unconditionally as I believe it to be a rather uncommon use case (dealing with inverted codes I mean). As stated in 172, I'd want to introduce a `DecodingHints` flag to request a search for inverted codes explicitly (3 modes actually: normal, inverted, both). Then the next logical step would be to simply invert the `BitMatrix` for the 2D codes. This would be a tad slower than what you proposed here but would work for all codes at once.

I'll let this PR open as a reminder.

neycyanshi:
As far as I know, DataMatrix inverted code is very common in different industries, depending on the material to print on. People like me will always choose decode both normal and inverted. But maybe my data is biased.

PR #175, opened 2020-11-30
Support PDF417 barcode images flipped along the horizontal axis

### LINKED ISSUE #152: PDF417 South African Driver License reading is not working
Could you have a look at the scanning of South African driver's licenses?

I have tried several ideas (preprocessing, few tweaks in pdf417 detector and decoder), but nothing works so far.
Always I am getting a large number of erasures in the final result. My goal is just to have a valid decoded bytes in the output.

Any advice would be appreciated.

tailsu:
This situation is apparently common with ZA driver licenses.

Fixes #152 
Needs an integration test.


axxel:
Thanks. I will add a sample image to the test set.

PR #179, opened 2020-12-10
fix/ Add C++17 build flag in CMakeList pybind

hbaocr:
Just add 2 flag of CMAKE  in CMakeLists.txt config to force the gcc build using C++17

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

Otherwise  the command 'python setup.py install' will be failed on some machine. Because of requiring CPP > C++14 when building the binding lib

### DIFF
--- a/wrappers/python/CMakeLists.txt
+++ b/wrappers/python/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.14)
 project(ZXingPython)
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 # get pybind11
 include(FetchContent)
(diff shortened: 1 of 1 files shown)


axxel (changes requested):
Thanks for reporting this. The problem is that we did not test the `setup.py` build option properly. If build from the top-level `CMakeLists.txt` the standard is already set to 17. Could you please do me the favor and move the two lines down into the `if (NOT hasParent)` block. That is the path that has the issue.

hbaocr:
> Thanks for reporting this. The problem is that we did not test the `setup.py` build option properly. If build from the top-level `CMakeLists.txt` the standard is already set to 17. Could you please do me the favor and move the two lines down into the `if (NOT hasParent)` block. That is the path that has the issue.

I have changed

axxel:
Thanks!

PR #181, opened 2020-12-16
Aztec::GetDimension() fix for layers 27 (131x131)

gitlost:
This fixes the `Aztec::GetDimension()` formula to work when `nbLayers` is 27. The previous formula returned 129 (it should be 131) causing an out of range exception. Test sample included.

(Bug reported by Andre Maute on the Zint [mailing list](https://sourceforge.net/p/zint/mailman/message/37177097/), though he was using ZXing where the same issue exists.)


axxel (changes requested):
Awesome. Thanks for fixing bugs cross-projects. One minor improvement, though. Could you replace your png with the following 60% smaller one (thanks to optipng). We have been rightly criticized for having too much binary bloat carrying around...
lorem-131x131

gitlost:
Thanks, have done so.

PR #186, opened 2021-01-03
Fix failure to parse text encoding from string

huycn:


axxel:
There would have been a smaller and less intrusive change (no memory allocations for all the std::string objects) by supplying a custom comparator object to the std::map. But I was too slow to suggest that ;).

huycn:
@axxel You are right, I didn't think about memory overhead. I will make the change as you suggested. Thanks!

PR #198, opened 2021-02-16
Fix incorrect class name, KRTextDecoder in KRTextEncoder.h

sidepelican:
`KRTextEncoder.h` has class `KRTextDecoder`. not **En**coder.

This will cause redeclaration error when include both `KRTextEncoder.h` and `KRTextDecoder.h`.


huycn:
Thanks!

PR #199, opened 2021-02-20
[QR] Stop to update bottomRight finder position when the alignment pattern not found

sidepelican:
## Issue

Cannot read the follwing QR codes.

## Cause

zxing-cpp has a unique algorithm that is not found in the original zxing.
Even when the AlignmentPattern was not found, updates the bottomRight position with its own method.
This process seems to work well when the QR code is distorted and the AlignmentPattern is hidden, but it gets to be sensitive to the distortion of the outer border of the FinderPattern.

Here is a visualization of the actual coordinates being sampled. The red dots are sample points.
You can see that a small difference in the image causes a large change in the sample point (look at the bottom right).
sampleGrid2 sampleGrid3

## Solution

As with the original ZXing, if no AlignmentPattern is found, the bottomRight position is determined by simple vector calculation only.

https://github.com/zxing/zxing/blob/44267115986e7fdcb2a9629fd7986f65ab9a2670/core/src/main/java/com/google/zxing/qrcode/detector/Detector.java#L154-L165

The current `QRDetector.cpp` overrides bottomRight even if AlignmentPattern is not found.

https://github.com/nu-book/zxing-cpp/blob/1d4eac6c6e50c43180ea7223b977bd67716aad1c/core/src/qrcode/QRDetector.cpp#L266

## Side effects

This change makes one test case fails.

https://github.com/nu-book/zxing-cpp/blob/1d4eac6c6e50c43180ea7223b977bd67716aad1c/test/samples/qrcode-2/%231314.jpg

However, this test case is very nasty because the AlignmentPattern is intentionally hidden and it is shot at an angle.
The original ZXing can't decode it, and the iOS Code Scanner can't scan it either. I think we can ignore this.


axxel:
Thanks a lot for this detailed PR. I'll have a closer look in the coming days, once I find the time. One thing I noticed: your sample JPG files are empty (according to the github UI).

sidepelican:
You can see the images with `Display the rich diff` button.

axxel:
I checked out your samples and found that only one of them actually fails to be decoded by the old code. After some other experiments I ended up making a change that is inspired by your idea but does not break any existing test cases. Thanks for your input!

sidepelican:
@axxel 
I see https://github.com/nu-book/zxing-cpp/commit/d36bf65da889b9cb67fb83beb6b55fa0dc373bc0 .
This problem is not only version 1.

Current implementation rescues the image that has perspective angle and that has no alignmentPattern (like https://github.com/nu-book/zxing-cpp/blob/1d4eac6c6e50c43180ea7223b977bd67716aad1c/test/samples/qrcode-2/%231314.jpg), but fails in some images that is positive. 

See the next image. This is version 2.



The sampling points are below.

| Current | My proposal |
|---|---|
| sampleGrid0 3 | sampleGrid0 2 |


I think we have to choose whether to rescue an image like this or  `%231314.jpg`.
I think in more cases it is better to rescue the image that is positive.

axxel:
I see your point and would like the code to be able to detect your second example as well. That said, it is somewhat contrived, since the alignment pattern is broken/corrupt in a similar way as it is in `#1314.jpg`. I'll have a look and see if I find a solution that can deal with both cases (trapezoid projections and low resolution/aliasing).

PR #200, opened 2021-03-02
Add ECI handling to Data Matrix

gitlost:
Adds ECI handling to Data Matrix decoder. Non-character set ECIs (i.e. those > 899) are just ignored.

Note uses `State` struct with only one member but future PRs for DMRE and structured append will add further fields.


axxel:
Awesome.

Question: is there some code you used as inspiration for the `ParseECIValue` function or do you have a copy of the specification at hand? The two sample images are your creation?

Note: I also have some initial DMRE support based on the recent changes in the Java code base lying around.

gitlost:
Hi thanks! I have the spec and the code is based on that. The 2 samples were created by Zint (modified unreleased version).

Re DMRE I looked at the recent Java base code changes but unfortunately it's based on an old spec, the stuff I've done is based on ISO/IEC 21471:2020. (I have vague intentions to upstream any changes to Java ZXing at some stage but my Java's weak so may be a while...)

axxel:
Ahh, someone with coding skills and the spec. Nice :-).

DMRE: I'll look into pushing my changes (which are based on upstream/outdated spec) and you could then maybe extend the few missing lines based on the ISO documents? See also this discussion: https://github.com/zxing/zxing/issues/1346

gitlost:
Ha a dangerous combination maybe.

Re DMRE ah I missed that discussion. Please go ahead as you say and I'll base the PR on that.

gitlost:
(P.S. If you'd rather not bother doing your changes let us know and I'll do the PR directly.)

axxel:
I checked and the only thing I have is the (outdated) extension of the table in the `DMVersion.cpp`. That alone will not cut it, though. There are going to be changes required in the `DMDetector.cpp`. Either I lost them or I only imagined them... Please go ahead with your PR and we'll continue the discussion there.

PR #208, opened 2021-03-18
Suppress DecodeHints deprecated warnings (internal)

gitlost:
This suppresses the warnings generated by the recently deprecated `DecodeHints` functions which are very noisy (note external use will still trigger them). See https://stackoverflow.com/a/13459924/664741


axxel:
Thanks for looking into this. I opted for a "cleaner" and compiler independent fix.

PR #213, opened 2021-04-02
Useful updates to cmake file

hartcw:
These are a couple of patches I have locally to improve the cmake file.
First installs the .pdb file allow with the libs on windows, for debug variants of the build.
Second enables position-independent-code compile option.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -454,6 +454,7 @@ add_library(ZXing::ZXing ALIAS ZXing)
 add_library(ZXing::Core ALIAS ZXing)
 
 set_target_properties(ZXing PROPERTIES EXPORT_NAME ZXing)
+set_target_properties(ZXing PROPERTIES POSITION_INDEPENDENT_CODE ON)
 if (PROJECT_VERSION)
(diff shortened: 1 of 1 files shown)


axxel:
Thanks for your contribution. Question: what is your reasoning to generally enable position independent code? Note: if you build a shared library (see `BUILD_SHARED_LIBS`) the default is already true.

@huycn Do you remember what the default value for `BUILD_SHARED_LIBS` is set to `OFF` on windows but `ON` everywhere else?

hartcw:
For my usecase, I am building zxing as a static library, but then using it in my code that gets compiled into a shared library.
So for that to work, zxing needs to be compiled with PIC enabled.

axxel:
I see. We also need this to build the python module. I just noticed that there is actually a 'bug' in the cmake setup in the sense, that disabling `BUILD_SHARED_LIBS` but enabling the python module does not compile. I thought about suggesting to make this flag optional so not everyone has to pay the performance penalty but a quick test revealed that this is unnoticible (on an x64 intel machine). So I'll merge it and add a comment. Thx.

PR #219, opened 2021-04-08
PDF417 Macro: fileId decode; segmentIndex 0 -> -1; Macro only symbol

gitlost:
Decodes the Macro PDF417 "fileId" codewords as 0-899 numbers, each 0-filled to width 3. Not doing any base 900 maths based on the example in ISO/IEC 15438:2015 Annex H.6, where (100, 200, 300) is transmitted as "100200300".

Sets the default of `segmentIndex` in PDFDecoderResultExtra to -1, and checks before setting `StructuredAppendInfo` (though check not really needed).

Allows a symbol to consist solely of a Macro (i.e. ~~removes check for `resultString` being empty~~ checks for non-existence on `resultString` empty check), as stated in Annex H.2 note.

Unit test additions. Also adds test entry that `pdf417-1/03-rot90.png` is readable when rotated 90 degrees.

Also suppresses a GNC maybe-uninitialized warning.

### DIFF
--- a/core/src/pdf417/PDFDecodedBitStreamParser.cpp
+++ b/core/src/pdf417/PDFDecodedBitStreamParser.cpp
@@ -29,6 +29,8 @@
 
 #include <array>
 #include <cassert>
+#include <iomanip>
+#include <sstream>
 #include <utility>
 
 namespace ZXing::Pdf417 {
@@ -571,12 +573,15 @@ DecodeStatus DecodeMacroBlock(const std::vector<int>& codewords, int codeIndex, 
 	if (StatusIsError(status)) {
 		return status;
 	}
-
 	resultMetadata.setSegmentIndex(std::stoi(strBuf));
 
-	std::string fileId;
-	codeIndex = TextCompaction(codewords, codeIndex, fileId);
-	resultMetadata.setFileId(fileId);
+	std::ostringstream fileId;
+	fileId.fill('0');
+	for (int i = 0; codeIndex < codewords[0] && codewords[codeIndex] != MACRO_PDF417_TERMINATOR
+			&& codewords[codeIndex] != BEGIN_MACRO_PDF417_OPTIONAL_FIELD; i++, codeIndex++) {
+		fileId << std::setw(3) << codewords[codeIndex];
+	}
+	resultMetadata.setFileId(fileId.str());
 
 	int optionalFieldsStart = -1;
 	if (codewords[codeIndex] == BEGIN_MACRO_PDF417_OPTIONAL_FIELD) {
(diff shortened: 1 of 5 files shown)


core/src/pdf417/PDFDecodedBitStreamParser.cpp
```diff
@@ -571,12 +573,15 @@ DecodeStatus DecodeMacroBlock(const std::vector<int>& codewords, int codeIndex,
...
-	std::string fileId;
-	codeIndex = TextCompaction(codewords, codeIndex, fileId);
-	resultMetadata.setFileId(fileId);
+	std::ostringstream fileId;
+	fileId.fill('0');
+	for (int i = 0; codeIndex < codewords[0] && codewords[codeIndex] != MACRO_PDF417_TERMINATOR
+			&& codewords[codeIndex] != BEGIN_MACRO_PDF417_OPTIONAL_FIELD; i++, codeIndex++) {
+		fileId << std::setw(3) << codewords[codeIndex];
+	}
+	resultMetadata.setFileId(fileId.str());
```
axxel, line 584: For the record: I stumbled upon this discrepancy between the code and the spec as well. I found at least one other library that does what ZXing did (allow 'random text') and since the id is only really required to be equal within a set, I thought, why not keep the behavior and allow the users to give it some meaning? Now I checked a little further and found other libs that model it as you just did as well as model it as a single integer (which is obviously insufficient).

Question: is your motivation for this change simply to make it 'right (TM)' or do you have a functional requirement?

Note: I just checked the PDR417Writer and there seems to be no support for MacroBlocks at all, so no inconsistencies there... Does zint strictly follow the spec?

gitlost: Yes the motivation is simply to make it 'right (TM)', and try to follow the spec - there's no functional requirement. The random text generated for the spec examples looks pretty weird though.

 (Zint doesn't support structured append for any barcodes, though it's something I want to add at some stage in the next hundred years.)

I'm trying to check out TEC-IT, though only have the demo so will have to massage its output (it accepts text for the id anyway). Will report back....

gitlost: Ok TEC-IT encodes the id using text compaction, so that's one in the "de facto" camp.

gitlost: Ok found a functional reason to follow the spec - depending on the fileId values, e.g. 899, it would decode into shifts and latches etc, dropping the info. Also different fileIds could resolve to the same text encoded value...

axxel: Fair enough. Would add a comment describing those insights? (difference compared to upstream/TEC-IT and the reason to switch)

gitlost: Done. I plan to do this change upstream too so left that bit out.

PR #222, opened 2021-04-16
PDF417: handle ECIs; some refactoring; Macro optional text field -> UTF-8

gitlost:
Adds ECI handling to PDF417.

Refactors Byte Compaction to look ahead to count 5-codeword batches, so as to determine if batch trailing or not (see NOTE in ISO/IEC 15438:2015 5.5.3.2).

Some refactoring of Text Compaction.

Backward-incompatible change of PDF417 Macro optional text fields from ECI 2 (Cp437) to UTF-8, needed to allow for ECIs in field.

Additional checks of `codeIndex` against symbol length (`codewords[0]`).

Passes down character set hint.

Test sample generated by (modified) Zint. Additional unit tests.


gitlost:
Pushed a fix to Numeric Compaction - it shouldn't treat the result as ASCII but respect and process ECIs as do the other compaction modes, as stated in the first paragraph in ISO/IEC 15438:2015 5.5.1. This only affects use of ECI 25 (UnicodeBig) in numerics (very unlikely but possible), as all the other ECIs do straight mappings of ASCII digits.

PR #223, opened 2021-04-17
DMDecoder: add Structured Append handling

gitlost:
Adds handling of Structured Append to Data Matrix.

The `fileId` is being saved as 2 0-filled 3-digit numbers, same as PDF417. Another choice would be to do a 254 base conversion to give a 0-64515 (or 1-64516) number instead.

Unit test.

### DIFF
--- a/core/src/datamatrix/DMDecoder.cpp
+++ b/core/src/datamatrix/DMDecoder.cpp
@@ -33,7 +33,9 @@
 
 #include <algorithm>
 #include <array>
+#include <iomanip>
 #include <optional>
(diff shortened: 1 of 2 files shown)


axxel:
You wrote in your unittest `Values outside 1-254 allowed`. Why then not simply `(fileId1 << 8) | fileId2`? That would look way more natural to me.

gitlost:
Wanted to be like PDF417, to avoid needless variation, but will do it that way if you think it's best.

axxel:
Thanks, appreciated.

PR #224, opened 2021-04-18
MCDecoder: add Structured Append handling

gitlost:
Adds handling of Structured Append to MaxiCode.

Test sample generated by [Labelary](http://labelary.com/viewer.html) and post-processed to reduce size. Unit test.


test/unit/maxicode/MCDecoderTest.cpp
```diff
@@ -0,0 +1,142 @@
...
+// Pad out to max data length 93 (mode 4)
+static void pad(ByteArray& padded)
+{
+	int i = 0;
+	for (i = (int) padded.size(); i < 20; i++) { // Primary
+		padded.push_back(33);
+	}
+	for (; i < 93 + 1; i++) { // Secondary (seem to require extra +1 dummy on Windows/Mac)
+		padded.push_back(33);
+	}
```
axxel, line 40: Those 2 loops don't make sens to me: you seem to simply want to add 33 at the end until the size of `padded` is at least 93 (or 94 which seems fishy regarding the windows/mac crash). Your code is equivalent to
```c++
while (padded.size() < 93)
    padded.push_back(33);
```
About the crash: I suggest you compile your code with an enabled address sanitizer.

gitlost: Re loops, MaxiCode codewords are in 2 parts, the 1st (primary) is 10 codewords of data then 10 codewords of ECC, the 2nd (secondary) is up to 93 codewords of data (mode 4) then the rest (to 144) is ECC.

Re crash, yes `-fsanitize=undefined -fsanitize=address` shows the issue but getting one on `PDF417DecoderTest` now...

axxel: The two loops still don't make sens to me. Maybe you wanted to _insert_ those ECC codewords after the 10th data code word?

gitlost: Yes was talking rubbish, the ECCs aren't even in the array so the simple padding you suggested is enough.

Need the +1 for the mode. Found the PDF417 bug (wasn't checking `codewords[0]` in `ProcessByteECIs`). Still looking at a few other things....

gitlost: Getting a couple of runtime undefined behaviour messages for `stb_image.h` and `NormalizedPattern` on running the blackbox tests with `-fsanitize=undefined`.

The `stb_image.h` one is "left shift of negative value" which is pretty benign I think. I see there are later versions of `stb_image.h` and `stb_image_write.h` available (same warning though) so they could be updated.

For `NormalizedPattern`, `moduleSize` can be zero so the divide afterwards gives a `NaN`, so that could be checked...

PR #229, opened 2021-05-13
cmake: add -Wshadow build flag and fix related shadowed variables

aurelienpre:
Suggestion to enable `-Wshadow` build flag, with fixes for the existing shadowed variables.

Builds fine, no warnings related to shadowing.
Unit tests pass.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -25,7 +25,7 @@ if (MSVC)
     )
 else()
     set (ZXING_CORE_LOCAL_DEFINES ${ZXING_CORE_LOCAL_DEFINES}
-        -Wall -Wextra -Wno-missing-braces -Werror=undef)
+        -Wall -Wextra -Wno-missing-braces -Werror=undef -Wshadow)
 endif()
 
 
--- a/core/src/BitMatrixCursor.h
+++ b/core/src/BitMatrixCursor.h
@@ -43,7 +43,7 @@ public:
 	POINT p; // current position
 	POINT d; // current direction
 
-	BitMatrixCursor(const BitMatrix& image, POINT p, POINT d) : img(&image), p(p) { setDirection(d); }
+	BitMatrixCursor(const BitMatrix& image, POINT _p, POINT _d) : img(&image), p(_p) { setDirection(_d); }
 
 	class Value
 	{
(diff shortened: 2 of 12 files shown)


axxel:
Thanks for your contribution but I'm no friend of those 'forced' renames with clunky names just to be able to silence `-Wshadow` warnings. I see how shadowed variable names can be an issue for readability but I don't like this general approach to avoid them at this cost (which also reduces readability).

And sorry for the late reply, I was pretty much offline for 4 weeks...

aurelienpre:
Ok, understood. If you see value in fixing shadowed variables, shall we think about better names then? Happy to gather suggestions and rework the MR, let me know.

axxel:
What I meant is that I generally agree that there might be situations where choosing a different name is the better choice but approaching the issue with the intention of fixing all occurrences, just so `-Wshadow` is quite is not good. If you have an issue with a specific variable, I'm happy to hear your thoughts.

aurelienpre:
I see. I'm using ZXing as a Conan package in a project where i'm enabling  `-Wshadow` and `-Werror`. ZXing package comes as precompiled lib + headers and I was having shadowed variables warnings as errors. Maybe we can keep the fixes for the public interface headers then. For those, I suggest to prefix the constructor arguments by a `_`, which would be relatively harmless for readbility, at least IMO.

axxel:
Expecting the code does not shadow variables within public headers seems fair, indeed. That said, if you use the API as currently promoted (via `ReadBardcode()`), then there seems to be no client-side visible issue. I tested that by adding `-Wshadow` to the example projects. In which header do you get the error and do you really need to include that header?

aurelienpre:
Just checked again now, I'm getting the shadowed variables warnings in `Point.h` which is included from `ReadBarcode.h` > `Result.h` > `Quadrilateral.h`.  This is with ZXing version 1.1.1 on Linux and using GCC 9.3.0. 
I'm attaching a txt file with the list of warnings for more details: [shadowed_variables_warnings.txt](https://github.com/nu-book/zxing-cpp/files/6568670/shadowed_variables_warnings.txt)

axxel:
Interestingly clang does not complain about the shadowing of constructor parameters. A quick searched returned this gcc bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147. Sigh.
Bottom line: I do believe `Point.h` is coded in proper and correct c++. Changing it just make a questionable gcc warning silent seems wrong to me. Would changing your compiler option to `-Wshadow=local` (see https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) be on option?

aurelienpre:
Interesting indeed, I wasn't aware of the this difference in behaviour (and the `local` / `global` specifiers for `-Wshadow` for that matter). Thanks for the links. I can confirm that gcc doesn't complain with `-Wshadow=local`. Seems like a reasonable way to approach the issue, I'll use that in my project then.

PR #239, opened 2021-07-25
Android: Using direct ByteBuffer to avoid memory allocation

kim-ninh:
I make a little memory optimization when passing preview frame data from Kotlin to C++. Instead of creating an extra bitmap, we can use ByteBuffer from the ImageProxy. This ByteBuffer is guaranteed to be a direct ByteBuffer so we can get a pointer to it in native code.

### DIFF
--- a/wrappers/android/app/src/main/java/com/example/zxingcppdemo/MainActivity.kt
+++ b/wrappers/android/app/src/main/java/com/example/zxingcppdemo/MainActivity.kt
@@ -17,6 +17,7 @@
 package com.example.zxingcppdemo
 
 import android.Manifest
+import android.annotation.SuppressLint
 import android.content.Context
 import android.content.pm.PackageManager
 import android.graphics.*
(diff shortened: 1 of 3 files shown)


axxel:
I like the general idea (avoiding the allocation and extra memcopy). That said, I have two questions:
 1. Do you notice an improvement in the execution time (left number in the top row in the app) compared to the old approach?
 2. The buffer in the ImageProxy object is blocked a lot longer (during the complete zxing computation, instead of only the copy). I'm no Android developer by any means but understood that those buffers are a scarce system resource and need to be 'released' (rather sooner than later?). That said, if I knew that access path to the underlying buffer, I would probably have done exactly that in the first place. :).

Note: if you care about saving memory, make sure to use the current HEAD, not version 1.2.0 as I recently removed a couple extra copies inside the cpp part.

I would not want to merge this PR as is, because it results in unnecessary code duplication. Do you want to give a 'cleanup' a go or would you rather have me pick the important part from your PR and commit it myself?

kim-ninh:
1. I will give it try to measure the execution time before and after this PR's change.
2. I'm not sure how long the ImageProxy object is blocked. I just think Google provides a new API to work with the camera starting Camera2 so it may be worth to try it.

Feel free to pick any modification of this PR that is suitable. :)

kim-ninh:
Well, the performance between the two versions is pretty close. On my Xiaomi Redmi 5+, it's about 30 ms.

axxel:
Thanks a lot! I cherry picked the interesting parts from your PR fixed a few issues and pushed it.

PR #245, opened 2021-08-24
DMDetector: fix one endless loop case with tryHarder config

neycyanshi:
When set `tryHarder` to `true`, I encountered an endless loop reading a DataMatrix code attached.
Endless loop found in traceGaps()
(105, 972) step
(106, 971) step
(107, 970) drift too much, project back to (104, 973) which doesn’t make any progress along line direction
and step back to (105, 972) fall in endless loop

A fix is to make sure when projecting back to traced line we are making progress in the line direction, the new position is not close to the old position.
[2021-08-24-09-05-32.png.zip](https://github.com/nu-book/zxing-cpp/files/7039156/2021-08-24-09-05-32.png.zip)


axxel:
I had a detailed look at the problem and your solution seems indeed to be the most effective approach. I'll add a comment and a minimized sample image. Thanks again.

PR #250, opened 2021-08-27
Build against system deps, Python test fixes

pabs3:
Add mode to build against system versions of dependencies.

Use right Python interpreter, port Python tests to Python 2.


pabs3:
@axxel did you get a chance to look into using an unpatched stb version?

axxel:
I totally forgot about it...

If you feel like it, please feel free to look into the possibility to treat stb_image similar to the other external dependencies. I.e. fetch it directly from github in case it is not provided by the system libraries. Then we can completely remove the copy from the repo, I believe.

pabs3:
Hmm, I thought the upstream version caused a big slowdown in the tests,
which was the reason for keeping the modified copy? You mentioned above
a more elaborate change which should fix that though?

The Debian package of libstb-dev contains all the stb headers used, so
once the slowdown issue is fixed, the same approach as now used for the
other dependencies should work for stb too.

A cmake patch for switching from the modified copy to the upstream and
system stb would be fairly easy, but it should probably only get merged
after fixing the slowdown.

Perhaps the right way forward would be to keep the patch modifying stb,
but delete the embedded stb copy, then auto-download stb from GitHub
and when the download happens, apply the patch modifying stb before
building against the modified stb. Then when using the system
libraries, have a separate option to build against the modified stb,
for the people who mostly want system libs but still want the speedup.

I've implemented this approach and am filing a new pull request.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #261, opened 2021-10-11
Enable -Werror=return-type and fix compilation

dfaure-kdab:
The compiler doesn't know that the switch covers all cases
because technically any int can be stuffed into an enum value.

ReadBarcode.cpp:91:1: error: control reaches end of non-void function [-Werror=return-type]


axxel:
Question: what compiler are you using. The ones I have here don't have an issue seeing that the switch statements are in fact covering all possible values.

dfaure-kdab:
I use gcc 11, but I always have -Werror=return-type in my cxxflags. Without it, it's easy to miss the warning scrolling by.

gcc (SUSE Linux) 11.2.1 20210816

PR #266, opened 2021-11-19
+ install pdb only in shared build

moodyhunter:
... when packaging into vcpkg, it seems that for a static build there's no PDB generated, which caused an error

image


axxel:
Thanks for the PR. I'm not a windows user, so I was wondering whether that really is the best practice way to go about it. It seems the default behavior of VS regarding the debug info is to embed it into static libs, so indeed no pdb file to be found. The following SO question suggest a slightly more concise solution: https://stackoverflow.com/questions/40860435/how-to-get-cmake-to-install-pdb-files-for-targets (simply add `OPTIONAL` to the install command).

Would you be willing to update your PR or should I just add this one line myself?

moodyhunter:
@axxel Thanks for the suggestion, PR updated!

PR #278, opened 2022-02-03
Remove the embedded copy of the stb image reading/writing headers

pabs3:
These headers have needed updates occasionally, removing them
means that they will be automatically up to update when used.

Search for stb among the installed system libraries and use them when
available, otherwise fetch the latest headers from the stb GitHub project.

The embedded headers were patched to improve the speed of some tests, so
keep the patch and apply it after the stb git repository has been fetched.

Use a separate variable to BUILD_SYSTEM_DEPS so that folks who mostly
want system libraries but want the patched stb speedup can have that.

Since the WebAssembly builds happen from a separate cmake file and
presumably can't use the system library, always fetch stb for them
and factor out the fetching logic into a common cmake include file.


CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ option (BUILD_BLACKBOX_TESTS "Build the black box reader/writer tests" ON)
 option (BUILD_UNIT_TESTS "Build the unit tests (don't enable for production builds)" OFF)
 option (BUILD_PYTHON_MODULE "Build the python module" OFF)
 set(BUILD_SYSTEM_DEPS "AUTO" CACHE STRING "Use system dependencies (AUTO/ALWAYS/NEVER)")
+set(BUILD_SYSTEM_DEPS_STB "${BUILD_SYSTEM_DEPS}" CACHE STRING "Use system stb dependency (AUTO/ALWAYS/NEVER).\nThe tests are faster with the patched embedded copy.")
```
axxel, line 12: Hence, no point in introducing a special `BUILD_SYSTEM_DEPS_STB` variable.

CMakeLists.txt
```diff
@@ -52,11 +53,35 @@ if (BUILD_UNIT_TESTS AND (NOT BUILD_WRITERS OR NOT BUILD_READERS))
...
 if(NOT BUILD_SYSTEM_DEPS IN_LIST BUILD_SYSTEM_DEPS_LIST)
     message(FATAL_ERROR "BUILD_SYSTEM_DEPS must be one of ${BUILD_SYSTEM_DEPS_LIST}")
 endif()
 
+set_property(CACHE BUILD_SYSTEM_DEPS_STB PROPERTY STRINGS ${BUILD_SYSTEM_DEPS_LIST})
+if(NOT BUILD_SYSTEM_DEPS_STB IN_LIST BUILD_SYSTEM_DEPS_LIST)
+    message(FATAL_ERROR "BUILD_SYSTEM_DEPS_STB must be one of ${BUILD_SYSTEM_DEPS_LIST}")
+endif()
+
+if ((BUILD_READERS OR BUILD_WRITERS) AND (BUILD_EXAMPLES OR BUILD_BLACKBOX_TESTS))
```
axxel, line 67: `(BUILD_READERS OR BUILD_WRITERS)` is always true, hence can be dropped here.

FetchSTB.cmake
```diff
@@ -0,0 +1,14 @@
+include(FetchContent)
+FetchContent_Declare (stb
+        GIT_REPOSITORY https://github.com/nothings/stb.git)
+FetchContent_MakeAvailable (stb)
+execute_process(COMMAND git -C ${stb_SOURCE_DIR} reset --hard @)
```
axxel, line 5: See above: that custom patch has no use anymore.

example/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ set (CMAKE_CXX_STANDARD 11)
 if (BUILD_READERS)
     add_executable (ZXingReader ZXingReader.cpp)
 
-    target_include_directories (ZXingReader PRIVATE ../thirdparty/stb)
+    target_link_libraries (ZXingReader stb::stb)
```
axxel, line 7: Please combine that with the subsequent line like so `target_link_libraries (ZXingReader stb::stb ZXing::ZXing)`. The same for all other similar occurrences below.

wrappers/wasm/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ cmake_minimum_required (VERSION 3.10)
 
 project (ZXingWasm)
 
+include (${CMAKE_SOURCE_DIR}/FetchSTB.cmake)
```
axxel, line 5: Good catch, I would have overlooked the wasm wrapper ;).

axxel (changes requested):
First of all: thanks for the immediate and generous contribution of your time :).

There has been a misunderstanding: my recent commit removed the use of the former "custom speedup patch", so there is absolutely no point in preserving that. My change increased the runtime of the Blackboxtests about 100ms on my machine and it localized it in the 'loading' part of the measurement, which I found acceptable.

pabs3:
Woops, I forgot to git fetch before updating the stb removal patch.

That test runtime increase seems acceptable indeed.

I've made most of the requested changes to the patch.

I'm not convinced (BUILD_READERS OR BUILD_WRITERS) is always true, when
both BUILD_READERS and BUILD_WRITERS are off and BUILD_UNIT_TESTS is
off, then BUILD_READERS and BUILD_WRITERS will not be turned back on.
Please let me know if I missed something there though.

Thanks for being open to using system libs and dropping code copies,
that is becoming rarer to see under currently popular dev practices.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

pabs3:
Fixed the WASM and PkgConfig based GitHub action failures.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

CMakeLists.txt
pabs3: I'm not convinced `(BUILD_READERS OR BUILD_WRITERS)` is always true, when
both BUILD_READERS and BUILD_WRITERS are off and BUILD_UNIT_TESTS is
off, then BUILD_READERS and BUILD_WRITERS will not be turned back on.
Please let me know if I missed something there though.

axxel: It might be that having both set to OFF is not technically prohibited by cmake as configured atm. but having both OFF makes no sens, the library would then do absolutely nothing. So, if you feel the need to fix that potential mis-configuration by the user then a specific check for that case should be added so that the rest of the code can rely on at least one being ON.

CMakeLists.txt
```diff
@@ -57,6 +57,24 @@ if(NOT BUILD_SYSTEM_DEPS IN_LIST BUILD_SYSTEM_DEPS_LIST)
...
+        pkg_check_modules (STB IMPORTED_TARGET stb)
+    elseif (BUILD_SYSTEM_DEPS STREQUAL "ALWAYS")
+        find_package(PkgConfig REQUIRED)
+        pkg_check_modules (STB REQUIRED IMPORTED_TARGET stb)
+    endif()
+
+    if (NOT STB_FOUND)
+        include (${CMAKE_SOURCE_DIR}/FetchSTB.cmake)
+    else()
+        add_library(stb::stb ALIAS PkgConfig::STB)
```
axxel, line 73: nice: I was about to suggest that change before already, but did not want to be too pedantic ;)

CMakeLists.txt
pabs3: Added a check for it, dropped that part of the stb check.

CMakeLists.txt
pabs3: I had neglected to test BUILD_SYSTEM_DEPS=AUTO with stb not installed, luckily the CI tests that scenario :)

pabs3:
Awesome, thanks again for the pleasant pull request experience :)

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #280, opened 2022-02-05
Fix spelling errors

pabs3:
Suggested-by: codespell & spellintian


core/src/BitArray.h
```diff
@@ -168,7 +168,7 @@ class BitArray
 
 	/**
 	* @param i bit to get
-	* @return true iff bit i is set
+	* @return true if bit i is set
```
axxel, line 171: Admittedly a bit math-slangy but that is not a spelling mistake but means "if and only if". Maybe spelling out like that would be best, indeed. See also the other occurrences.

axxel (changes requested):
Thanks!

I feel like there should be a way to quickly remove individual changes from a PR via the github web interface, that way I would not have to bother people for such trivialities...

pabs3:
I've reverted the iff removal, I'll leave it up to you decide whether
to change that to something longer.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #281, opened 2022-02-07
GS1 GSIN country lookup updates/fixes; CMake BUILD_EXAMPLE_OPENCV option

gitlost:
Updates the GSIN country lookup for GS1 company prefixes to the latest GS1 [listing](https://www.gs1.org/standards/id-keys/company-prefix) and fixes a few typos in the `COUNTRIES` array.

Also fixes a non-found issue where for instance "test/samples/ean13-2/23.png" ("1920081045006") would return "FR" (should return nothing).

Adds dealing with EAN-8 which requires adding an optional arg to `LookupCountryIdentifier()` to differentiate it from UPC/E.

There's a non-related fix to make building the OpenCV example optional - needed by me to build as the OpenCV package is found on my machine but I don't have all the other dependencies that are required.

Add unit test.


axxel:
Welcome back! ;).

Would you please remove the cmake and OpenCV related change and let me know in a different issue what the problem with "the other dependencies" is? I don't like the extra "BUILD_OPENCV_EXAMPLE" option and would rather fix that properly (and independent of this GTIN change). Thanks in advance.

gitlost:
Hi yes have removed the OpenCV change and will open an issue about it.

axxel:
Thanks. Sounds like you had a broken OpenCV installation. I decided to squash and merge to prevent the back-and-for change of the cmake files. In general I started to prefer simple merges because they do show more prominently who contributed the changes. I believe a way to update your pull request would have been to `git commit --amend` (revert the cmake changes) and then `git push -f` to update the PR. Just in case something similar should show up again and you'd prefer a 'standard merge' from my side.

gitlost:
Thanks, will keep the PR update tip in mind...

PR #286, opened 2022-02-14
To get Windows build to work add depname arg to zxing_add_package

gitlost:
Seems need to use "googletest" as arg to `FetchContent_Declare()` and `FetchContent_MakeAvailable()` for Windows to build properly, so add a `depname` arg to `zxing_add_package()` (and update calling instances).

Also seems need to set `gtest_force_shared_crt` before calling `FetchContent_MakeAvailable()`, so add hacky test for this to `zxing_add_package()`.


axxel:
Thanks for looking into this. I found this issue this morning, which hinted at a cmake/windows problem: https://gitlab.kitware.com/cmake/cmake/-/issues/22659. Anyway...

Issue #214, opened 2021-04-02
Follow upstream with support for Symbology Identifiers?


axxel:
For those who missed the discussion: https://github.com/zxing/zxing/pull/1372.

Anyone interested in porting that feature into this c++ code base?

PR #288, opened 2022-02-17
ODCode128Reader: use FNC1 to determine if GS1, remove assumeGS1 (#265)

### LINKED ISSUE #265: Consider Code128 symbols to be potentially GS1-128 by default?
Given the issue reported [upstream](https://github.com/zxing/zxing/issues/1457), I'm wandering whether or not we should consider changing the default value of `DecodingHints::assumeGS1` to true? Note the complete meaning of the flag would be "assume the code to be GS1-128 iff the first character is \FNC1".

Any opinion on that @gitlost?

gitlost:
Closes #265
Part-implements #214

For Code128, uses position of FNC1 to determine if GS1 (or AIM) barcode rather than `assumeGS1` hint which is no longer used and is marked as deprecated in `DecodeHints`.

Also sets new `symbologyIdentifier` attribute added to "Result.h" (implemented for Code128 only for now).

Adds processing of optional  ".result.txt" files in "test/samples" that contain keyed Result attributes and values for additional validation.

Implementation of further symbology identifiers will follow in subsequent PRs (also checking of other Result attributes in ".result.txt" files).

### DIFF
--- a/core/src/DecodeHints.h
+++ b/core/src/DecodeHints.h
@@ -74,6 +74,10 @@ public:
 	TYPE GETTER() const noexcept { return _##GETTER; } \
 	DecodeHints& SETTER(TYPE v) { return _##GETTER = std::move(v), *this; }
 
+#define ZX_PROPERTY_DEPRECATED(TYPE, GETTER, SETTER) \
+	[[deprecated]] TYPE GETTER() const noexcept { return _##GETTER; } \
+	[[deprecated]] DecodeHints& SETTER(TYPE v) { return _##GETTER = std::move(v), *this; }
+
 	/// Specify a set of BarcodeFormats that should be searched for, the default is all supported formats.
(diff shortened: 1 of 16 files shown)


axxel:
Since you've been tuning your patch a few times (it seems): do you consider the current version final?

I'd merge it as is but would like to suggest to include only those metadata fields in future .result.txt files that are non-default (like you did for all of the new ones except the first 2).

gitlost:
Hi, thanks, removed the non-default fields as requested. Plus fixed some sp -> tab formatting.

Yes the previous force-pushes were me forgetting to add stuff (the ".result.txt" files - 2 goes) and remove stuff (the "convertFNC1" field from "ODCode128Reader.h"). In future will annotate any force-pushes and add comment indicating ready for review.

This PR is ready for review.

axxel:
Awesome. Thanks as always.

PR #289, opened 2022-02-19
AZDecoder: implement symbology identifier (#214)

gitlost:
Part-implements #214

Implements `symbologyIdentifier` for Aztec Code.

Also changes `AZDetector::IsValidPoint()` to check for `y >= 0` instead of `y > 0` following ZXing PR [1430](https://github.com/zxing/zxing/pull/1430) (no unit test for this as couldn't make it matter but seems like a simple typo).

Adds Aztec GS1 sample as none previously; adds unit test for symbol identifiers also.

Also removes `MakeLarger()/inflate()` from "AZDetectorTest.cpp" as no longer seems necessary.


gitlost:
This PR is now ready for review.

axxel:
If you care to improve your PRs (in my opinion) even more in the future, please consider:
a) split something like this PR here in at least 3 commits (`IsValidPoint` fix + `MakeLarger()` removal + rest)
b) add the more descriptive comment at the level of the commit, not the PR, because the latter will get lost in the git history

Thanks and till next time :).

gitlost:
Thanks, will do...

PR #290, opened 2022-02-22
Implements symbologyIdentifier for DataMatrix Code (#214)

gitlost:
Part-implements #214

Implements symbologyIdentifier for DataMatrix Code.

Adds unit test & 2 samples - one for standard DM, one for DMRE.

Also fixes unit tests for Structured Append which must be 1st codeword, and Reader Init sample which wasn't actually Reader Init ([8066b527161e074dcb2025cd60978625de9dc956])


gitlost:
This PR is ready for review.

core/src/datamatrix/DMDecoder.cpp
```diff
@@ -158,7 +166,7 @@ static Mode DecodeAsciiSegment(BitSource& bits, std::string& result, std::string
 {
 	Shift128 upperShift;
 
-	while (bits.available() >= 8) {
+	for (int position = 1; bits.available() >= 8; position++) {
```
axxel, line 169: I believe the introduction of this "manual position" counting is not necessary: the `BitSource` class has a member `byteOffset()` that tracks the current position in the bit-stream. It seems we are only reading one byte a time (`bits.readBits(8)`) and you increment position for every byte read. I.e. you can simply use `byteOffset() + 1` where you need to evaluate the position I believe.

gitlost: Excellent, have force-pushed the change [e7ecad8](https://github.com/nu-book/zxing-cpp/commit/e7ecad81948c0efeb55f770a3ce647a413fc6ae8)

gitlost:
This PR is ready for review again.

axxel:
Thanks a lot!

PR #293, opened 2022-02-24
Implements symbologyIdentifier for QR Code (#214)

gitlost:
Part-implements #214

Implements symbologyIdentifier for QR Code.

Adds GS1 sample and unit test.


gitlost:
This PR is ready for review.

test/unit/qrcode/QRDecodedBitStreamParserTest.cpp
```diff
@@ -96,3 +96,49 @@ TEST(QRDecodedBitStreamParserTest, HanziLevel1)
...
+	result = DecodeBitStream({0x20, 0x09, 0x40}, version, ecLevel, "");
+	EXPECT_EQ(result.symbologyIdentifier(), "]Q1");
+	EXPECT_EQ(result.text(), L"A");
+
+	// GS1 "FNC1(1st) NUM(4) 2001"
+	result = DecodeBitStream({0x51, 0x01, 0x0C, 0x81, 0x00}, version, ecLevel, "");
+	EXPECT_EQ(result.symbologyIdentifier(), "]Q3");
+	EXPECT_EQ(result.text(), L"2001"); // "(20)01"
+
+	// GS1 "NUM(4) 2001 FNC1(1st) 301" - FNC1(1st) can occur anywhere
```
axxel, line 116: This seems to contradict the spec 7.4.8.2: "...it shall only be used once in a symbol and shall be placed
immediately before the first mode indicator used for efficient data encoding...". @gitlost Why did you add this test?

gitlost: Well it's a simplified test and a bit lazy - the FNC1(1st) could also (and should according to the spec) occur after any ECI or Structured Append codewords (same with FNC2(1st)). So it would be better if e.g. an ECI byte sequence were used instead of NUM(4). Also decoding is permissive and allows for variations from the spec and unlike most symbologies QR Code dedicates a specific codeword for FNC1(1st) so it can appear unambiguously anywhere... The comment should definitely be improved though.

PR #301, opened 2022-03-06
Fix: Minor refactor to prefer standard algorithm: `std::accumulate`

tcoyvwac:
Refactored to prefer `std::accumulate` over raw range-for loop.


axxel:
Thanks. In part I believe it requires more cognitive effort to understand what exactly is happening after the change but in general I am on the same page (and did similar changes in the past). Please see also the `Reduce` function in `ZXContainerAlgorithms.h` which is trying to anticipate a c++20 ranges style API.

Issue #44, opened 2018-11-30
Make this the "official" zxing-cpp?


simonschmeisser:
There are currently hundreds of forks of zxing-cpp out there with two main "base"-forks as far as I can tell: this one and https://github.com/glassechidna/zxing-cpp

For concentrating efforts and for packaging in distributions (linux, chocolately, home-brew, ports, ...) it would be great to have one "main" fork. This one seems to be a) the most active in recent times, b) has unit tests for all codes and c) has a version number/release since yesterday. So if you agree it would be great to focus on this one.

Any thoughts or objections?

We could even "move" it to zxing/zxing-cpp or zxing-cpp/zxing I guess

simonschmeisser:
there was a question about API compatibility at https://github.com/glassechidna/zxing-cpp/issues/68#issuecomment-443129574 would you like to comment?

axxel:
I would certainly welcome making this fork the "official" one. Drawing some attention was my secret plan when I commented on a couple of issue tickets in the 'other' fork ;). Probably unique to this fork (as I perceived it, when I was looking for a c++ version of xing originally) is that it is not based in any way on the original zxing-cpp code base but is a complete rewrite based on the then most recent Java version with a 'native c++-11' flavor in mind (instead of trying to mimic the JAVA api 1:1. I.e. e.g. no `Ref` class in the API).

Even though I contributed a bit of c++ 'cleanup' type of changes, a few minor detection improvements for 1D bar codes and reimplemented the DataMatrix detector, this fork is really @huycn work. So it is definitively his call regarding how this fork should move forward (or not).

I expect there might be some more 'plumbing' necessary if the code base is exposed to more OSes / compiler versions but nothing I'd worry about.

huycn:
Me too, I have no objection and will be happy make this the "main" c++ fork. I'm certain that my boss, @ChantalLemay, would agree with me on that. I wonder what is the most efficient way to make this happen. I will probably contact maintainers of original ZXing repo.

simonschmeisser:
Great news! Contacting the ZXing maintainers sounds like a very good way forward. It would probably also simplify distributing maintenance effort across more shoulders (if desired). Thanks!

huycn:
Thank to Sean Owen, we have now a link from ZXing project page to this repo.

axxel:
[...]

PR #309, opened 2022-04-04
Add Android library project

markusfisch:
Makes it possible (and easy) to use ZXing C++ with any Android app.

Android components are distributed as so-called Android Archives (AAR) which are basically JARs with Android resources. This commit adds a library project that bundles the native libraries and the necessary (JNI) wrapper in a generic form.

The resulting AAR file can easily be dropped in any Android app and makes ZXing C++ immediately useable without having to clone and build the ZXing C++ core libraries and JNI wrapper in the context of another app.

To check the fitness of the library, there's a sample/demo app beside the library project. This app, and the library, is based on the sources in `wrappers/android`.

However, the JNI wrapper of this library differs from `wrappers/android` to make it usable for Android's legacy camera API, Camera2 as well as CameraX. The returned result (object) now contains all available data (including binary data and meta stuff). And there are functions for barcode generation too.

The sample app was migrated from Kotlin synthetics to simple `findViewById()`s because Kotlin synthetics are [deprecated](https://developer.android.com/topic/libraries/view-binding/migration) (in favor of view binding, which I didn't use because I think a sample should be as simple and basic as possible).

There's also an (untested!) JitPack integration to automatically publish this library on jitpack.io as soon as there's a Release on GitHub for version "v1.2.1" (see "afterEvaluate" in zxingcpp/build.gradle). JitPack should build this library as soon as it is requested - I'm just unsure if JitPack will find the (correct) `build.gradle` because it's in `wrappers/aar` and not in the root folder. Unfortunately, there's no way to test this apart from trying it.

I've tried to stick to the coding style and conventions of the root project as much as possible. If there's anything you don't like, please let me know and I will fix it.


huycn:
Thank you for your contribution!
I agree with your changes for going with automatic binding and making it work with other camera API.
The ability to export image as text or SVG is interesting.
And I have no complains about the coding style.

My only concern is now there are 2 wrappers for Android with some code duplication. It's maybe good idea to to merge them together? @axxel may have some opinions?

axxel:
I have not looked at the code, yet, but also don't like the code duplication aspect of it. I'll gladly take a closer look later this week (maybe today) and come up with a suggestion on how I would personally prefer to proceed. Grüße nach N aus B ;)

markusfisch:
Danke und schöne Grüße zurück 😉 

I just noted that it's probably wrong to just put `Result.rawBytes()` (from the C++ side) into `Result.rawBytes` (on the JVM), as I do in the PR. `Result.rawBytes()` contains the raw data _before_ decoding - but what an Android app needs is the decoded data in an unconverted form. Before String conversion, which may corrupt binary data in UTF32 to UTF16 conversion.

As it's now, it's unreliable to read barcodes with binary content (DataMatrix, QR Codes in binary mode) from Android because `Result` will contain potentially corrupted data in `Result.text` and undecoded, obscure data in `Result.rawBytes`.

My "fix" would be to put `wstring Result.text()` into a (JVM) `ByteArray` and give this as `rawBytes` to Android. If you like that approach, would you prefer another commit in this PR or an extra PR for this change? Or would you prefer an additional member in the (JVM) `Result` for the decoded `ByteArray`?

As for the duplication aspect, I just didn't want to be too invasive and make grave changes to the existing wrapper.

axxel:
Thanks for your effort and intention to contribute! I'm not an Android guy and appreciate input from people who know the best practices in this regard. As I understand, your PR has a couple separate intentions, like
 * making the library more accessible by providing a binary bundle as the primary goal
 * fixing some deprecation warnings in the example app
 * supporting a deprecated camera API
 * enable some form of barcode generation support

All of that seems generally reasonable to me. That said, I'm strongly opinionated against the introduction of those in basically a copy of the existing wrapper. If you feel like spending the extra time is worth it for you, I'd like to propose the following way forward: you come up with a new PR that only deals with the first goal and does that as a change of the existing wrapper. After that worked out, we proceed from there.

markusfisch:
Well, I was hoping to replace the existing wrapper with this solution, but I understand that it might be a bit much for just one PR. There are a lot of changes and some of them must seem unrelated or opinionated to you.

And I just realized that I kind of messed up the description of this PR 🙈 Because it's already possible to build an AAR (e.g. a binary bundle) with the existing wrapper (just run `$ ./gradlew :zxingcpp:assembleRelease`). Strictly speaking there's nothing that _needs_ to change for the goal of providing a binary bundle that can be used in other apps.

What is missing, and what I can make another PR for if you want, is the automatic distribution of that binary bundle over jitpack.io (would be just a few lines in `wrappers/android/zxingcpp/build.gradle`, essentially [this](https://github.com/nu-book/zxing-cpp/pull/309/files#diff-7da007f8c08f761b114ba0a0c53263ffd65fd2086367bfc28caca76f7962dc4bR44) block).

But - the resulting binary bundle would have a few disadvantages:
* It's package name would be `com.example.zxingcpp`, what's not a good name for a production library.
* It's not [minified](https://developer.android.com/studio/build/shrink-code) and cannot be minified (in a project that makes use of this library) because it's missing the proper configuration (specifically [this](https://github.com/nu-book/zxing-cpp/pull/309/files#diff-13875b93725ffe32274ae5aae9844871ddba8c609d8e8c3c5081a43814e3b250R24)).
* It depends on CameraX. CameraX is just a support library that's based on the Camera2 API. So an implementor would have to use CameraX too, even if the app works with the Camera2 API.
* It requires minSdkVersion 26, what excludes many devices today. Especially outside the western hemisphere.

If you want, I can fix any of these things in the new PR too. Just as you choose. However, removing the CameraX requirement would bring us a good step closer to this PR 😉 These are probably the biggest problems for Android developers that may keep them from using this as a ready-to-use binary bundle. At least these are the reasons I can't use it.

What I should have described better is that this solution not only provides a binary bundle, but also adds (almost) all features to make it a convenient replacement for the Java version of ZXing. This includes barcode generation, but also the extended result and especially the ability to use it with whatever camera API and Android version there is.

So, instead of having everyone solve the same problems again and again, I thought it would be a good idea to improve the existing wrapper to be a full-fledged replacement for the (slow and stagnant) Java version of ZXing. This is why there's so much in this PR. Sorry for the mess and for not being clearer.

Please tell me what from the above things you want to have in the new PR and I will make one.

axxel:
Thanks for the detailed feedback and analysis. I suggest we start with a prioritization of "low hanging fruit" / importance.
 * package name: you chose "com.nubook.android.zxingcpp". why "android"? Maybe we could use the opportunity and try to get a little closer to #44 by using a more generic name than 'nubook'. Upstream is apparently called "com.google.zxing" which seems weird to me, now that I think about it. Any suggestions?
 * not sure what minification flag you are specifically talking about but sounds like a no-brainer
 * CameraX: I got the impression that this is the way to go according to Google when releasing anything new related to camera stuff. I'd very much like to offer a dead-simple and 'natural feeling' API and `ImageProxy` seemed to me to be the right level of abstraction (requiring the user to know that she has to pick the first of several color planes to make this work is too low level in my opinion).
 * minSdkVersion: what would be a reasonable value in your opinion and why?

In general: if you can separate your changes as nicely as you did in your description, please separate them into individual commits in any future PR. :)

markusfisch:
> package name: you chose "com.nubook.android.zxingcpp". why "android"?

The "android" part is kind of a best practice to clearly separate this package from a plain Java package. This is useful when working with other Java (only) libraries. One can see at first glance, that this package belongs to and requires Android.

Since the wrapper requires the Android SDK, I put it under "com.nubook.android.zxingcpp". But the "android" part is completely optional. If you don't like it, I will remove it.

> Upstream is apparently called "com.google.zxing" which seems weird to me, now that I think about it. Any suggestions?

"com.google.zxing" is the original package name of the Java implementation. I wouldn't choose it because it potentially collides with the Java implementation (although I can't think of a sane reason why someone would want the Java *and* the C++ version in the same project, apart from comparing performance, of course).

I chose "com.nubook.android.zxing" because of the GitHub account where this fork lives. Of course, it can be anything, but traditionally Java packages are based on domain names (of the author of this package). And because of the GitHub account, the dependency path for JitPack will be `com.github.nu-book:zxing-cpp:x.x.x` anyway. But if you don't want to include "com.nubook", any other domain that can be associated with this project would be perfectly fine too.

In principle, even just "zxingcpp" would do (instead of "com.nubook.android.zxingcpp"), but it sure would stand out a bit 😉 

> not sure what minification flag you are specifically talking about but sounds like a no-brainer

I was talking about this [flag](https://github.com/nu-book/zxing-cpp/blob/master/wrappers/android/zxingcpp/build.gradle#L28), and the corresponding proguard/r8 [configuration](https://github.com/nu-book/zxing-cpp/pull/309/files#diff-13875b93725ffe32274ae5aae9844871ddba8c609d8e8c3c5081a43814e3b250R24).

Minification is important on Android to minimize download size and to enable optimizations.

> CameraX: I got the impression that this is the way to go according to Google when releasing anything new related to camera stuff. I'd very much like to offer a dead-simple and 'natural feeling' API and ImageProxy seemed to me to be the right level of abstraction (requiring the user to know that she has to pick the first of several color planes to make this work is too low level in my opinion).

Yes, I can totally understand why you had this impression. Google tries very hard to make CameraX the new default. But unfortunately, it's not that simple. I really wish it was 😉 

Google provides CameraX to help make camera app development easier. And it's true that Google recommends starting new apps with CameraX. But it's important to understand that it is just a convenience library on top of the Camera2 API that makes a lot of common tasks a lot easier. Of course, this comes at the cost of having another dependency, which can be a burden, if the Camera2 API is enough for a certain application. If I don't need it, why would I want to make my app larger, more complex and more error-prone?

So in my opinion this comes down to giving users of this library the freedom to choose whatever fits their situation.

This is also why I added support for the legacy Camera API. Technically there is no reason why this library shouldn't work with the legacy Camera API too, and in apps with a minSDKs before 21. There are many devices out there that are implementing the new Camera2 API just in [LEGACY](https://source.android.com/devices/camera/versioning#camera_api2), what means it's basically the legacy Camera API behind the Camera2 API. The performance of these implemenations can be much worse than using the legacy API directly, unfortunately. So it's complicated, and there can be perfectly good reasons to not use Camera2 or CameraX.

> minSdkVersion: what would be a reasonable value in your opinion and why?

Google currently recommends API 21 (Android 5.0 Lollipop) and claims this would address ~98% of devices (you can see this when you're creating a new project in Android Studio). This percentage is based on devices that connect to Google Play, what isn't the case in some parts of the world, so this should be taken with a grain of salt.

Generally it's a good idea to go as low as possible and only increase the minSDK when required. Especially for a library, of course. Following this idea, we can go as low as API 9 as soon as we remove the CameraX requirement (where this PR is, see [here](https://github.com/nu-book/zxing-cpp/pull/309/files#diff-7da007f8c08f761b114ba0a0c53263ffd65fd2086367bfc28caca76f7962dc4bR11). Of course, this PR _can_ also be used with CameraX).

Again, this is about freedom of choice. The lower the minSDK, the fewer the hurdles.

> In general: if you can separate your changes as nicely as you did in your description, please separate them into individual commits in any future PR. :)

Yes, absolutely, I should've done so from the start 🙈

axxel:
* package name: I was not suggesting the use of `com.google.zxing` if at all, my suggestion would have been `com.google.zxingcpp` but that seems pretentious to me. I would have thought maybe `org.zxing.zxingcpp` would potentially be acceptable. I see that neither `zxing.com` nor `zxing-cpp.com` is used by anyone. What about `com.zxingcpp` then?
 * minification: go for it
 * minsdk/CameraX: I would suggest to go for 21 then, keep the `read(ImageProxy)` API (meaning CameraX, which is indeed less complex and less error-prone which is the whole reason for existence :)) but maybe also add the `ByteArray` version, so non Camera2 devices can be used natively (if I got that correctly?).

After that, your completion of the `Result` struct seems to be a good next step. I suggest we keep that for a separate PR though.

markusfisch:
I think `com.zxingcpp` is fine 👍 

And of course, I can keep `read(ImageProxy)` as you like. Which means that we can't go any lower than minSdk 21 anyway, because the library will remain to have a CameraX dependency ([ImageProxy](https://developer.android.com/reference/androidx/camera/core/ImageProxy) is part of CameraX).

But please let me make this clear one last time: This means the Android library, the binary bundle, can only ever be used in apps that have a minSdk of at least 21 and that are willing and able to include CameraX as well (no matter if they use it or not). These requirements are not strictly necessary, but a compromise to make the library easier to use (because there's no need to manually pick a plane). No hard feelings, just so we are clear about that point 😉 

If that is fine with you, I will make a new PR tomorrow.

The `ByteArray` function was for the legacy camera API where the camera frame is delivered as a `ByteArray` in [onPreviewFrame](https://developer.android.com/reference/android/hardware/Camera.PreviewCallback#onPreviewFrame(byte[],%20android.hardware.Camera)). In Camera2 we get an [Image](https://developer.android.com/reference/kotlin/android/media/Image).

axxel:
Yes, that is what I understood and would like to merge. Thanks for your understanding.

Before you start with your PR(s): a little rant about my current state of mind regarding the 'ideal' PR ;). For a set of independent changes (like the ones discussed here), I'd prefer to see them as individual commits inside a PR. A problem might occur if a change is requested, because piling up 'fix' commits in this PR causes noise. I know there is a relatively easy way to replace the last commit in a PR (`git commit --amend` and then `git push -f`). If you know a way to swap out any individual commit, then please let me know ;). Anyway, the best approach resulting from that would be a 'rolling' review where any potentially controversial commit comes last in the PR, might be changed after review and only then pushed the next commit. The other option is simply to push individual PRs but that produces other noise (lots of merge commits). Maybe this is over-complicating things, but you seem like the type of person who would be capable and willing to participate in this 'ideal PR' crafting ;).

markusfisch:
Yes, I understand that completely, and I care about the git history very much too 😉 

Which is why I will compose my PR(s) just as you like them, so they fit into the history of this project seamlessly. For example, and because I already started, here's the branch I created for the new PR we agreed on:
https://github.com/kurzdigital/zxing-cpp/commits/improve_android_wrapper

It just contains two (hopefully) good commits regarding the package name and the minification. Would it be okay to make one PR from that branch? Or would you rather like two PRs?

Personally, I prefer an alternative approach to merge commits and force pushes. I'm always squashing PRs into _one_ nice commit, what makes the actual commits in a PR (branch) irrelevant. Of course, that only works well if a PR contains just one topic/feature. If this is not the case, I break it up in separate topics/features and merge these in separate commits instead of the PR as a whole.

So I handle PRs exactly like the well known feature/branch workflow with squash commits. The idea (of squashing commits) is to have exactly one meaningful, well written commit for each feature and loose all the irrelevant noise that accrued during development. Ideally, this connects all relevant changes (for a given feature) in just one commit. The results is a clean history (without merge commits), with commits you can easily work with later on (cherry-pick, revert, bisect, etc).

But that's just what works for me. And of course, this may seem a little odd for someone making that huge PR 😉 I really should've included all the work commits that were necessary to get to this result to make the changes clear and easy to follow. But since I commit a lot in feature branches, I was afraid it would be too messy, and that you are probably only interested in a high level description.

Well, anyway, another interesting way to manage PRs would be to use an interactive rebase (`git rebase -i`). This would allow to reorder, fix and merge commits in the PR branch before merging the PR. But of course, this would mean a force push, too.

axxel:
That branch you started looks very good. Thanks. I suggest you also add a commit with the `minSdkVersion` downgrade to 21. That should be a very small change as well, right?

About squashing: That certainly works for the one-PR-per-one-fix/improvement approach, which is completely adequate for a single bug fix or the like and also easy to use from a maintainer side with the github squash-and-merge feature. One advantage of still merging instead of squashing (from my side) is that the [network](https://github.com/nu-book/zxing-cpp/network) view provides a visual representation of where the changes came from.

About interactive rebasing: that seems to be the gold standard, at least for people who know how to use it: make individual commits, append commits with fixes after review and then squash the fix commits into the respective original commits until there is a 'clean' series of commits that can all stand on their own and could have been merged separately. If you want to try that, you can pile up more commits in your branch and we'll see how it goes. I have no personal experience with the interactive rebase feature of git.

markusfisch:
Oh, right, I forgot to lower the minSdk 🙈 Sorry. But now it's in and I've opened a PR from that branch:
https://github.com/nu-book/zxing-cpp/pull/310

> One advantage of still merging instead of squashing (from my side) is that the [network](https://github.com/nu-book/zxing-cpp/network) view provides a visual representation of where the changes came from.
Squash merging a PR keeps the original author for the new, squashed commit. So you would still see this there. This is what a squashed PR looks like when merged, for example:
https://github.com/markusfisch/BinaryEye/commit/9e395d19288be8ef763c328e9d45e1d30b262944

> If you want to try that, you can pile up more commits in your branch and we'll see how it goes.
I'd love to!

axxel:
> Oh, right, I forgot to lower the minSdk see_no_evil Sorry. But now it's in and I've opened a PR from that branch: #310

I could merge that as is but... (see below)

> Squash merging a PR keeps the original author for the new, squashed commit.

I know. I've gone that way before to not have all those merge commits but I changed my mind to make the origin of the commits visible in the network graph. It's purely cosmetic and github centric but it felt to me like paying more respect to the authors.

> I'd love to!

Ok, then I will wait with the merging of your new PR until further notice and let you add more features. What about the extension of the `Result` struct? The barcode writer stuff is something that I'd like to discuss separately before putting more work into this direction.

I started to look into View Binding as the recommended replacement of the deprecated kotlin-extension in the demo app, just to let you know.

markusfisch:
Okay, I'll close this PR then and go over to the new one.

PR #310, opened 2022-04-07
Improve Android library project

markusfisch:
So it becomes ready-to-use for other apps.


axxel:
question about proguard/minifyEnable: from what I understand by reading a bit about best practices is that proguard usage in this case (open source native library with very thin kotlin wrapper) this does indeed not make sense (and can be removed as you did). But that is completely opposite your original claim. So you changed your mind, right?

markusfisch:
Yes, I didn't think about this being unimportant for a library when I saw that `minifyEnabled` was set to `false`. I should have realised earlier that the whole `release` block makes no sense here. I guess `minifyEnabled` triggered me somehow 😉 

Sorry about the confusion. But minification is still important! Just not in a library project 😉

axxel:
In case we are waiting on each other: this PR is mergable as is but I understood that you'd want to pile up more commits, right? Just let me know once you are done.

Question: did you use the terminal interface of git for the interactive rebase or did you find something more "beginner friendly"?

markusfisch:
Well, I think it probably makes sense to merge it now, because these changes are probably the most urgent ones to make the library useable in the context of another app. And I can always create new PRs 😉   

If you want, I could add the JitPack integration to make this library available via jitpack.io. This would make it very easy for Android developers to include the library in their apps, without having to clone this repository and create the library themselves.

However, this would require adding a file (`jitpack.yml`) in the root folder of this project, unfortunately. Without it, JitPack wouldn't find the library it should build. Would that be acceptable?

About the extension of the `Result` class - I'm not so sure anymore that this is really necessary. Maybe I was just scratching my own itch here 😉 Is this something you would care about? And if so, would you care about all the fields form the C++ `Result` class, or just specific ones?

> Question: did you use the terminal interface of git for the interactive rebase or did you find something more "beginner friendly"?

I used the terminal interface, sorry 😉

axxel:
ad jitpack: adding a `yml` file is certainly possible. After the experience with the python and NuGet package repository I'm not totally convinced that a fully automatic publication/release of an android library is required. But if you could provide knowledge, maybe even in form of a readme or in form of a manually triggered CI script (see [build-winrt](https://github.com/nu-book/zxing-cpp/blob/master/.github/workflows/build-winrt.yml) on how to do that, that would certainly be helpful.

ad `Result`: the current limitation to basically only include the text is mainly due to the fact, that I did not feel enough pressure to figure out the rather cryptic JNI solution of returning a custom struct. I would definitively like to see the position information returned and add a visual clue inside the demo app. orientation, ecLevel and symbologyIdentifier would also be nice, I think. For the rest, I'd be happy to wait until someone complains that it's missing. You mentioned your own itch: what of this data were you interested in?

markusfisch:
Well, JitPack only builds when a certain (git) tag is requested, and then keeps the artefact for subsequent requests for the same tag (which usually is a version number of course). I'm not sure if this is fully automatic or manually triggered (by setting a tag) 😉 

Anyway, as you see I added a `README.md`, in `wrappers/android`, that explains how to build and use the AAR in other apps.

About the `Result`, `position`, `orientation`, `ecLevel` and `symbologyIdentifier` are now added, too.

I am especially interested in a raw byte representation (in the JVM) of the decoded content. There are quite some barcodes that contain binary data (DataMatrix or ICAO Visual Digital Seals), and in order to properly parse them on the JVM, the raw bytes are required. Binary data can be corrupted by encoding it as `String`. But that is probably a quite specific application and it's completely reasonable to wait until someone really needs it.

Just because you mentioned the visual clue to the demo app, am I right that the C++ fork has no equivalent for [DecodeHintType.NEED_RESULT_POINT_CALLBACK](https://zxing.github.io/zxing/apidocs/com/google/zxing/DecodeHintType.html#NEED_RESULT_POINT_CALLBACK)? Of course, this wouldn't be necessary for the visual clue (as soon as a barcode has been detected). But [ResultPointCallback](https://zxing.github.io/zxing/apidocs/com/google/zxing/ResultPointCallback.html) also delivers points for partial detections, which can be helpful sometimes. Nothing crucial, but I thought I ask in case I have been missing something here.

axxel:
Thanks for the 2 new commits!

Regarding the result stuff: I forgot, that I did pass the result struct as a parameter so I could modify its fields from inside the cpp code and that the extension with more fields is then rather trivial. IIRC you used another approach in your initial PR where you construct the Result object on the c++ side and pass it back, right? Would that not be the 'right thing' to do? Also, the position is not a rectangle, is it a quadrilateral (a convex polygon with 4 corners). I guess one would have to build a custom type on the JVM side just as I did on the c++ side? If you don't feel like spending more time on this, let me know and I'll merge this and cook up something along those lines myself.

About the raw bytes: if I understood your comments in the other PR correctly, what you need is neither provided by the `rawBytes` nor the `text` member on the c++ side. So it is not JVM related at all at this point. If so, I suggest you open a new issue with an example image and the expected output. Maybe @gitlost could jump in on that discussion then as well, as he is the 'en/de-coding specialist' around here ;).

No `ResultPointCallback` in the c++ port.

markusfisch:
> Also, the position is not a rectangle, is it a quadrilateral (a convex polygon with 4 corners).

Ah, of course, my mistake 🙈 I added a new commit in which I transfer the four corner points in a `Position` data class instead.

> IIRC you used another approach in your initial PR where you construct the Result object on the c++ side and pass it back, right? Would that not be the 'right thing' to do?

Yes, at least in my opinion. But I didn't want to be too invasive again. I created the Result object on the c++ side to avoid unnecessary allocations when no barcode could be found. But of course, these allocations aren't that critical, so it's probably more a matter of taste.

For transferring the four points of the position polygon I changed to that approach to avoid constructing `Point()`s at 0, 0 on the JVM side.

> If you don't feel like spending more time on this, let me know and I'll merge this and cook up something along those lines myself.

If it's okay for you I'd be happy to see this merged 😄 

I can't really use the Android wrapper in this form in my apps, so maybe this would be a good time to hand this over to you.

>  what you need is neither provided by the rawBytes nor the text member on the c++ side. So it is not JVM related at all at this point.

It is JVM related because the corrupting encoding happends when the `text` member is encoded as a Java String [here](https://github.com/nu-book/zxing-cpp/blob/master/wrappers/android/zxingcpp/src/main/cpp/JNIUtils.cpp#L55). So I'm interested in the unmodified contents of the `text` member, which is a `std:wstring` that can savely be used to store binary data.

markusfisch:
Great! 😊 Thanks for merging!

> I think there might be something off with the white-space…

Ah, sorry about that. Just out of curiosity, is it the continuing indent? Should it have been a double indent? I tried very hard to match the style of the project 😉 

> And the reason is my insistence on the dependency on CameraX?

Essentially yes - I just can't use CameraX everyhwere 🤷‍♂️  I've got more than one app still using the legacy camera API and cannot switch (yet) without loosing performance and/or users, unfortunately.

But it's no big deal. I will use a fork for now and am glad to have contributed! There are probably enough people who already use CameraX, and for them it will definitely be easier to use the (wrapper) library now. Also, I might use the official wrapper at some point for apps that use CameraX.

> Any ideas on how you want to fix this issue for your use-case?

I chose a very pragmatic approach: I simply create a (JVM) `ByteArray` and fill it with the first byte of each `wchar_t` from `text`. See [here](https://github.com/markusfisch/zxing-cpp/blob/master/wrappers/aar/zxingcpp/src/main/cpp/ZxingCpp.cpp#L74).

axxel:
> I tried very hard to match the style of the project

and succeeded 100% ;). I was just fooled (again) by an inconsistency in githubs rendering of tabs in the diff...

> Essentially yes - I just can't use CameraX everyhwere man_shrugging I've got more than one app still using the legacy camera API and cannot switch (yet) without loosing performance and/or users, unfortunately.

My understanding was that the (only) downside of the CameraX dependency for you is that your app would be bigger than it needs to be but you would not have to use the library at all. And I'm not against adding a "legacy" function that allows to work on `ByteArray` instead of `ImageProxy`.

markusfisch:
> My understanding was that the (only) downside of the CameraX dependency for you is that your app would be bigger than it needs to be but you would not have to use the library at all.

Actually it's more the unused dependency (in case the app does use the legacy camera API) that would be a problem for me. The library itself, CameraX, would probably be optimized out by R8's tree shaking (although that's just a guess, haven't tried yet).

But because of this dependency, I would have to increase the minSDK, which is another thing I deeply care about. I also think that dependencies should generally be chosen wisely, because over time they can (and in my experience usually will) become a burden. I've burnt my fingers on that more than once 😉 And yes, also and especially with Google's stuff. The deprecation of RenderScript comes to my mind as a most recent example.

Of course, YMMV, and maybe I am too conservative in that regard 😉 In my first PR, I was trying to build a full replacement for the Java implementation of ZXing, which is what I need, in its most flexible form. Because I have a couple of apps at work and at least one personally (Binary Eye), that would really benefit from using the C++ fork over the Java implementation. But I realize that my demands maybe rather special.

> and if you would be in the position to quickly add a simple "draw 4 lines over the preview based on the position data" feature to the demo app, that would be much appreciated. ;)

I can do that tomorrow! I did this a couple of times now 😉 

> Of maybe just a hint to the API to use "in a modern android app" for this kind of task would also save me some considerable google time, I guess... :)

I would use a simple `View` in front of the view finder, overwrite `onDraw` and use [Canvas.drawPath()](https://developer.android.com/reference/android/graphics/Canvas#drawPath(android.graphics.Path,%20android.graphics.Paint)) to draw the polygon. The interesting part is mapping the coordinates from the camera frame to the view, but there's [image2View](https://github.com/nu-book/zxing-cpp/blob/master/wrappers/android/app/src/main/java/com/example/zxingcppdemo/MainActivity.kt#L237) already.

axxel:
@markusfisch I'm wondering why your name does not show up on the contributors page: https://github.com/nu-book/zxing-cpp/graphs/contributors

Maybe this is the reason: https://stackoverflow.com/a/43527384/2088798 ?

markusfisch:
@axxel Yes, that could be the cause! I just added my work email to my profile.

I used my work email, as I did it on behalf of my company.

axxel:
Either the change is not immediate or it did not help. But now I understand why some of your branches appeared under the kurzdigital name ;).

Do you think Hendrik might be interested in up-streaming the iOS wrapper that he seems to be working on?

markusfisch:
Yes, definitely, he's just not quite finished yet. But there will be a PR soon! 😉

axxel:
Looking forward to :). Your change did do the trick after all: github just needed to update their cached data. You are now the 4th most productive contributor! ;).

markusfisch:
Oh wow! 😉 Thank you for pointing me to enter my work email!

PR #311, opened 2022-04-12
android: draw result points/position on screen

markusfisch:
To highlight a detected barcode on the screen.

I changed as little as possible, but I had to touch `image2View` to make it handle 180 degree rotations of arbitrary coordinates. This is important when using the Java implementation for 180 degree image rotations (which is usually one of the landscape orientations).

Please note that scanning 1D barcodes just draws one line as ZXing return just two result points for them.

Also, the result points from the Java implementation do not match exactly with the C++ implementation, at least not for QR Codes. But the difference is probably intentional, I guess.

This code could be optimized a little by storing the result of the scale and offset calculation in `image2View` in the object instance to avoid having to recalculate them for every point.

### DIFF
--- a/wrappers/android/app/src/main/java/com/example/zxingcppdemo/DetectorView.kt
+++ b/wrappers/android/app/src/main/java/com/example/zxingcppdemo/DetectorView.kt
@@ -1,0 +1,48 @@
+package com.example.zxingcppdemo
+
+import android.content.Context
+import android.graphics.*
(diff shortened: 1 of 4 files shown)


axxel:
Thanks a bunch! I'll merge it, play with it and possibly tune it a bit to my liking... ;).

markusfisch:
Great! 😊 And thanks for merging!

PR #313, opened 2022-04-14
android: remove clearing overlay canvas

markusfisch:
Fixing a bug I introduced 😬 

Clearing the Canvas is unnecessary, and doing so clears out the camera frame when the underlying surface is the same, which it is for some devices/graphics drivers (just noticed this on a Xiaomi Mi A2 Lite).

Besides, I really like that you moved drawing the crop rectangle into the overlay, and made use of the canvas matrix to map the coordinates. So much simpler 👍


axxel:
Thanks. I tried the "Rebase and merge" button, now. Curious how this looks in the history/network graph.

markusfisch:
Thank you 😄

PR #317, opened 2022-04-19
Fix threshold test in VS2019 Debug build.

corbers:
The setup code for the test creates a buffer using vector's reserve.  This works for the release build but in debug build VS uses checked iterators which causes an error to be generated when the vector is accessed using [].  I changed reserve to resize.


axxel:
Your profile does not show up in the contributors list: https://github.com/nu-book/zxing-cpp/graphs/contributors. If you care for that, please have a look at https://stackoverflow.com/a/43527384/2088798. That helped @markusfisch before to fix the same issue.

PR #318, opened 2022-04-19
android: use row stride for Java impl too

markusfisch:
Because the row stride (of the Y plane) is (on some devices) greater than the image width, I was not able to scan barcodes with the Java implementation on some of my test devices.


axxel:
Thanks! (I only ever tested the code on a Pixel 3...)

markusfisch:
Thanks again for merging 😄 

Unfortunately, there are issues on Android that only occur on some devices but not on others. Part of the game, I guess 😉

PR #320, opened 2022-04-20
Fix PDF417 infinite loop

### LINKED ISSUE #316: ZXing hangs on a certain image
I have found that this image caused ZXing (build from commit 1e39cbe0505dd3ae25836a123a0498ea90a850d1) to run forever. Looks like it's running the PDF417 reader.
bug

Note: This image is a resized version of [this one](https://github.com/zxing/zxing/blob/master/core/src/test/resources/blackbox/pdf417-4/02-02.png) from the ZXing repo.

duanqn:
Backstepping should not violate `startRow`.
Fixes #316 

bug-vis

In the image above, red & blue polygons are the detected PDF417 regions. The green line is the 171st row. This change restricts the backstepping so that it can find the two regions below the green line. Before this fix, the program would find the previous 2 regions again and thus end up in an infinite loop.

### DIFF
--- a/core/src/pdf417/PDFDetector.cpp
+++ b/core/src/pdf417/PDFDetector.cpp
@@ -162,10 +162,11 @@ FindRowsWithPattern(const BitMatrix& matrix, int height, int width, int startRow
 {
 	bool found = false;
 	int startPos, endPos;
+	int minStartRow = startRow;
 	std::vector<int> counters(pattern.size(), 0);
 	for (; startRow < height; startRow += ROW_STEP) {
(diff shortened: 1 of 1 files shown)


axxel:
Thanks a lot for reporting *and* fixing this :).

Interestingly, this code has not been touched since it has been directly ported from the upstream Java code base. So I would assume they have the same issue in their code base as well (https://github.com/zxing/zxing/blob/8265242784be6f5493f1ede4b1fd1432b1666a54/core/src/main/java/com/google/zxing/pdf417/detector/Detector.java#L199) zxing.org does not hang, though, with that input image, so maybe we have some change upward in the callstack that caused this loop.

PR #326, opened 2022-05-04
Fixed comparison bug in MultiFormatReader.cpp

ptc-brainer:
This PR fixes a potential crash issue caused by an invalid comparison happening the `MultiFormatReader.cpp`. The comparison uses the `<=` operator when comparing the x position of barcodes, which could, if the positions are identical, result in an invalid comparison result were `a > b == b > a`. 
Depending on the compiler and sort implementation, this could cause a crash due to the object trying to be moved onto itself.

Changing the comparison operator to `<` fixes this problem.

### DIFF
--- a/core/src/MultiFormatReader.cpp
+++ b/core/src/MultiFormatReader.cpp
@@ -90,7 +90,7 @@ Results MultiFormatReader::readMultiple(const BinaryBitmap& image, int maxSymbol
 	std::sort(res.begin(), res.end(), [](const Result& l, const Result& r) {
 		auto lp = l.position().topLeft();
 		auto rp = r.position().topLeft();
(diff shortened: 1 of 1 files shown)


axxel:
Thanks for the PR, sounds like I did not design that functor carefully enough ;).

Question: how did you find it? Did you come a cross an instance where it actually crashed? I'm curious because there should never be two symbols with the same topLeft position. If so, then there would be another bug.

ptc-brainer:
> 

I came across this crash while testing this on iOS with life camera feed with a stress-test scenario with all supported barcode types printed to a sheet of paper. It happened very rarely but consistently. As to why there where multiple detections with the same top-left position, I cannot say.

axxel:
I see. If you have the possibility to let it run again with this setup (without the fix) and somehow get access to the input image in that instance, that would be most welcome.

PR #327, opened 2022-05-04
Generate the PDF417 ratio table at compile time

vkrause:
This not only avoids the entire computation at runtime, it also moves the
required memory from .bss (allocated at runtime per process) to .rodata
(shared between all processes and only loaded on demand), cutting down the
.bss section size by about 90%.


vkrause:
Unfortunate timing, I know :) Just stumbled over this by chance today while checking that the 1.3.0 update had worked here.

Issue #334, opened 2022-05-08
How to improve binary data support? (Community feedback requested)


axxel:
Both @vkrause and @markusfisch recently took part in discussions touching on the issue of handling binary data in the `Result` of `ReadBarcode`, see e.g. [here](https://github.com/nu-book/zxing-cpp/issues/128#issuecomment-1120196476). I believe there have been others in the past as well. As far as I currently understand the problem space, it looks like this:

Barcodes can contain text (as in unicode) and/or binary data. Our current interface does not give any clue what kind of data is stored inside the `std::wstring Result::text()` member. Then we additionally have `Result::rawBytes()` which is possibly also not on the right abstraction level (something about still containing error correction codes or format specific block layout issues that the client of that code might need to know about comes to my mind but my memory is foggy and I did not want to do a thorough investigation right now).

But as @tailsu rightly pointed out [here](https://github.com/nu-book/zxing-cpp/issues/211#issuecomment-811273690), we actually have potentially the situation where those two parts of the content would need to be kept separate (text/binary) in the result to really be able to handle it properly. Heuristics like the one incorporated [here](https://invent.kde.org/frameworks/prison/-/blob/master/src/scanner/videoscannerworker.cpp#L177) by @vkrause are bound to fail, once there are actually both types of data contained inside one symbol. Maybe that is a totally esoteric/niche use case that we don't need to worry about? I'm not interested in adding a complicated 'ordered list of segments of varying data types' kind of interface that no one needs!

May I ask @gitlost how zint is handling this potential mix of data types in one symbol?

Side note: somewhat related to this discussion is the fact I'd very much like to switch from the Windows centric `std::wstring` to a utf8 based `std::string` representation of the result text (see http://utf8everywhere.org/ for a motivation :)). This would basically remove the need for everyone to deal with `TextUtfEncoding` helper functions and remove the glaring inconsistencies in the `Result` regarding string types we have today.

gitlost:
Hi apart from QR Code (and Han Xin and Grid Matrix, which have similar issues), I'm not really getting the data type text/binary issue, and I don't see a problem either encoding or decoding, but I haven't followed the discussions fully here or upstream so I'm probably not familiar enough with the issues, on the decoding side anyway.

[...]

Issue #333, opened 2022-05-07
Please bump SONAME following API breakage


hosiet:
Hi,

I was working on packaging zxing-cpp in Debian/Ubuntu. As I read in https://github.com/nu-book/zxing-cpp/releases/tag/v1.3.0 , v1.3.0 introduces API breakage by removing deprecated APIs. Essentially we should bump library [SONAME](https://en.wikipedia.org/wiki/Soname) to avoid older binaries that dynamically linked to `libzxing.so.1` from crashing due to missing symbols. Ideally this should be done before each version release that introduces API/ABI breakage.

May I suggest decoupling `SONAME` with `{PROJECT_VERSION_MAJOR}` and bump `SONAME` to 2, given that we are obviously not following semver? The following patch should do the job. Let me know if you prefer me to submit a Pull Request instead.

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required (VERSION 3.14)
 
 project (ZXing VERSION "1.3.0" LANGUAGES CXX)
+set (ZXING_SONAME 2)
 
 option (BUILD_WRITERS "Build with writer support (encoders)" ON)
 option (BUILD_READERS "Build with reader support (decoders)" ON)
diff --git a/core/CMakeLists.txt b/core/CMakeLists.txt
index 0d4bff9..a9cf460 100644
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -455,7 +455,7 @@ set_target_properties(ZXing PROPERTIES EXPORT_NAME ZXing)
 set_target_properties(ZXing PROPERTIES POSITION_INDEPENDENT_CODE ON)
 if (PROJECT_VERSION)
     set_target_properties(ZXing PROPERTIES VERSION ${PROJECT_VERSION})
-    set_target_properties(ZXing PROPERTIES SOVERSION ${PROJECT_VERSION_MAJOR})
+    set_target_properties(ZXing PROPERTIES SOVERSION ${ZXING_SONAME})
 endif()
 
 include (GNUInstallDirs)
```

This would greatly help downstream Linux distributions to track library API/ABI changes, and ensure that consumers of zxing-cpp would transition to a new library version seamlessly.

Please let me know if you have any questions or concerns. Thanks!

axxel:
[...]

Issue #332, opened 2022-05-07
Discussion about the future of the barcode generation / writer part of this project


axxel:
I've not touched any of that code ever since I started investing time here. As mentioned earlier I would in fact prefer to drop that code if there was a better OSS replacement. It seems to me the [zint](https://github.com/woo-j/zint) project with @gitlost apparently being the lead maintainer now, would be a very good candidate if it was not for the licensing limitations (GPLv3)? I was about to pose the question of whether there could be done something about that but luckily stumbled upon the following statement in their README `Please DO NOT post messages asking for us to change the license arrangements. You will be ignored.`. I will henceforth not speak of it again. ;)

If anyone knows another alternative that is already better than this code base here, please speak up. A 15 minute google search did not reveal anything. Interestingly, though, the name zxing did not come up in that search either, which only underlines my impression that nether this nor even the upstream project are inherently associated with the generation side of the equation. That said, if there is no better commercial license compatible alternative then that is how it is.

Current limitations as far as I see them:
 * inconsistent interface with the `ReadBarcode` side
 * no `DataBar` support (#197)
 * possibly quite a few missing minor improvements that upstream Java code acquired over the last 3-4 years
 * either the code is rock solid or simply unused, based on the bug-tracking history of the last couple of years

Questions to all: who does use this part and for what? What are you missing the most?

vkrause:
We do use the PDF417 generator from ZXing in https://invent.kde.org/frameworks/prison/. For the other barcode formats that is using a mix of own implementations and single-format external libraries (libqrcode, libdtmx). This is mostly for historic reasons, with this largely predating our use of ZXing.

This isn't ideal obviously, as mentioned elsewhere I'd personally like to see this de-duplicated/unified, and ZXing looked like the right option for that. Should generation move out of scope here that wouldn't be the end of the world either though, we'd just need to find a replacement for the PDF417 code then.

axxel:
@vkrause Thanks for the feedback, so there is someone actively using that functionality. Good to know! Please don't get me wrong: I see no value in dropping working code for no good reason and without providing a better alternative. I just thought there might be something 'better' (in terms of actively maintained) that I don't know about and we might somehow incorporate or collaborate with that other project. If that does not show up, then sure, why not use more of zxing and reduce your list of dependencies...

@gitlost is already contributing substantially to the decoder part and would certainly have the expertise to also help improving the encoder but I assume working on the same problem in two different projects is not really appealing for you? ;)

[...]

PR #337, opened 2022-05-24
Adding a wrapper for iOS

parallaxe:
Co-authored-by: Christian Braun <braun.christian.bamberg@gmail.com>

This pull request adds a wrapper of ZXing for iOS. It offers one class for reading and one for writing. 
ZXing itself will be compiled to an universal framework, `xcframework`, that contains the binaries and
headers for the iOS-simulator and for iOS-devices. The wrapper-code is written in Objective-C++, whereas
the public part of the wrapper is Objective-C only. By this, it can also be used in Swift.

To build the `xcframework`, I had to add a section to the CMakeLists.txt. This enables building `frameworks`
for Apple platforms. Those frameworks are only either for the simulator, or for real devices. But they can be
combined to a single `xcframework`. This process is automated by the `build-release.sh`-script. 
The section in the CMakeLists.txt should not change the behaviour on other platforms (tested on Linux so
far).

To use the wrapper, you first have to build the `xcframework` by calling the build-script and then adding
the `Package.swift`-file to your project. A demoproject is included, where also the usage is demonstrated.


axxel:
Hallo Hendrik und Christian, thanks for your offer to contribute, much appreciated. I skimmed through your PR and have a few comments and suggestions for improvement. First, a few general things:

 1. Are you willing and in the position to 'take care' of this code in the longer term? I'm asking this because I have no experience with ios/ObjectiveC and I can't even compile the code myself. The public facing API is going to change in the future (see e.g. #334, #332 and #333 for a few current discussions), so the wrapper will need to be kept up-to-date and if the code does not even build anymore in 6 months from now, then there is not much point in adding it.
 2. You have no license attribution attached to your source files. Please follow the recent addition of `// SPDX-License-Identifier: Apache-2.0` comments. Also, please consider removing the comments that state the file-name and change the `Created by ...` lines to `Copyright 2020 <name of entity>`.
 3. Is the ObjectiveC/C++ part minimal? Or would there be room to convert more of that code to Swift? (On the android side there is everything in Kotlin and nothing in Java).
 4. Have you looked into the possibility to replace the `build-release.sh` with standard cmake code?
 5. We definitively need to hook that wrapper into the CI system before merging (see also point 4.)
 6. Do you care about the writing part yourself (right now) or did you add this more for sake of completeness?
 7. I see 2 names related to the packge/lib as a whole: `ZXing` and `ZXingWrapper`. I'd like to stay as consistent with the rest of the project as possible. The python and the android wrappers use the name `zxing-cpp` / `zxingcpp` (e.g. the python package is called `zxing-cpp` but the module is called `zxingcpp` because modules can not have '-' in their names). The names were chosen primarily to prevent a naming clash with "upstream" wrappers. The project keeps moving away from Java ZXing (in terms of API, capabilities and performance) and therefore I became more and more fond of actually having that different name. The c++ library itself is called `ZXing` but from today's perspective, I'd rather have that be `zxing-cpp` as well. How that would map best to the two names and how much best practice from the ios world went into your choice, I can't say. Please explain.

axxel:
Since you did not react to my comments, yet, I'm wondering if I completely put you off with that feedback. If so, I'd like to invite you shortly describe what part of it was too much, so I may learn from my mistakes :).

parallaxe:
Hi Axxel, thanks for your thoughtful analysis.

Regarding your questions:
1. We (and with 'we' I mean the company KURZ Digital Solutions) will maintain the wrapper over long term, yes. We already use the 'old' zxing in multiple apps and will migrate to this fork over time and also use it for new apps. If there as a need to update the iOS wrapper, we will do so. 
2. Updated and pushed 👍 .
3. Yes, I consider this as minimal. As there is no interoperability between C++ and Swift, we need ObjC++ as glue-code. But as the included demo-project shows, the API is also nicely usable with Swift out of the box.
4. To be honest, no. I circle around cmake and its obscure syntax whenever possible. If it is a hard requirement, I will migrate it. But from our experience and other build-scripts we are using, the xcodebuild-commands are stable over the years and we expect no real effort in maintaining the build-script.
5. Yes, I agree. It seems like that the CI is based on github workflows. I haven't worked with them yet, but it looks like that there are also free macos-runner that we could use to build the `xcframework` and the `demo`-project. I will try to configure a workflow for it. Or can we modify the existing macos-configuration? Does this sound feasible to you?
6. We have an app in the Apple AppStore ("Code Scan") that allows the user to create barcodes. This is currently done by the 'old' zxing, but we will migrate it to this fork. This is currently the main source of the requirement for the writer and the reason why it is included. There are currently no plans of integrating the writer-part in other apps, but you never know. 😬 
7. The naming-scheme of the `ZXing.xcframework` is quite common (starting with uppercase-letters), so I would leave it as it is. I agree that the name of `ZXingWrapper` is a bit off and we should change it. But I have to look up the conventions of SPM-package-names (Swift Package Manager) and if there even is a conclusion yet by myself first.

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
```diff
@@ -0,0 +1,109 @@
...
+    return result;
+}
+
+- (nullable ZXIResult *)readCGImage: (nonnull CGImageRef)image error:(NSError **)error {
+    CGColorSpaceRef colorSpace = CGColorSpaceCreateWithName(kCGColorSpaceGenericGray);
+    CGFloat cols = CGImageGetWidth(image);
+    CGFloat rows = CGImageGetHeight(image);
+    NSMutableData *data = [NSMutableData dataWithLength: cols * rows];
+
+    CGContextRef contextRef = CGBitmapContextCreate(
```
axxel, line 44: I don't know about the iOS world here, but I assume the 'native' formats of the cameras are similar compared to the android world, i.e. the most common format would be some type of YUV pixel forma. In that case there would be no image conversion necessary and simply a waste of computing resources, because all we need is the Y-plane. Please also look at [the qt example](https://github.com/nu-book/zxing-cpp/blob/bee813aebf01d42476b1b5c463f9e07ed0067443/example/ZXingQtReader.h#L165) for reference.

wrappers/ios/Sources/Wrapper/Reader/ZXIResult.h
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#import <Foundation/Foundation.h>
+#import "ZXIFormat.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ZXIResult : NSObject
+@property(nonatomic, strong) NSString *text;
+@property(nonatomic, strong) NSData *rawBytes;
```
axxel, line 12: please rename to `binary` for consistency. You seem to need access to 'binary' data? Please have a look at #334 and join the discussion if you find something is missing.

wrappers/ios/Sources/Wrapper/ZXIFormat.h
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#import <Foundation/Foundation.h>
+#ifndef ZXIFormat_h
+#define ZXIFormat_h
+
+typedef NS_ENUM(NSInteger, ZXIFormat) {
+    NONE, AZTEC, CODABAR, CODE_39, CODE_93, CODE_128, DATA_BAR, DATA_BAR_EXPANDED,
+    DATA_MATRIX, EAN_8, EAN_13, ITF, MAXICODE, PDF_417, QR_CODE, UPC_A, UPC_E,
+    ONE_D_CODES, TWO_D_CODES, MICRO_QR_CODE, ANY
```
axxel, line 12: please reorder and put the `MICRO_QR_CODE` right next to `QR_CODE`.

wrappers/ios/demo/demo/SceneDelegate.swift
```diff
@@ -0,0 +1,50 @@
+//  demo
+//
+// Copyright 2022 KURZ Digital Solutions GmbH
+//
+
```
axxel, line 5: license comment missing here

wrappers/ios/demo/demo/ViewController.swift
```diff
@@ -0,0 +1,70 @@
+//  demo
+//
+// Copyright 2022 KURZ Digital Solutions GmbH
+//
```
axxel, line 4: and here...

wrappers/ios/demo/demo/ViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+                mediaType: AVMediaType.video,
+                position: .back)
+
+            let device = discoverySession.devices.first!
+            let cameraInput = try! AVCaptureDeviceInput(device: device)
+            self.captureSession.beginConfiguration()
+            self.captureSession.addInput(cameraInput)
+            let videoDataOutput = AVCaptureVideoDataOutput()
+            videoDataOutput.setSampleBufferDelegate(self, queue: self.queue)
+            videoDataOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: Int(kCVPixelFormatType_32BGRA)]
```
axxel, line 37: maybe choose another format here? (see above comment about YUV formats)

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
```diff
@@ -0,0 +1,109 @@
...
+    CGContextDrawImage(contextRef, CGRectMake(0, 0, cols, rows), image);
+    CGContextRelease(contextRef);
+
+    ImageView imageView = ImageView(
+              static_cast<const uint8_t *>(data.bytes),
+              static_cast<int>(cols),
+              static_cast<int>(rows),
+              ImageFormat::Lum);
+
+    Result result = ReadBarcode(imageView, [ZXIBarcodeReader DecodeHintsFromZXIOptions:self.hints]);
```
axxel, line 62: Please use `ReadBarcodes()` here instead and set `DecodeHints().setMaxNumberOfSymbols(1)`. This does almost the same thing as `ReadBarcode()` but not quite for historical reasons. Also have you thought about making the wrapper return a list of multiple codes right from the start instead of adding it later on and then having to support two variants?

wrappers/ios/Sources/Wrapper/Reader/ZXIResult.h
```diff
@@ -0,0 +1,20 @@
+// Copyright 2022 KURZ Digital Solutions GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#import <Foundation/Foundation.h>
+#import "ZXIFormat.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ZXIResult : NSObject
```
axxel, line 10: would you be willing to add a few more useful properties from `Result`? Not necessarily all but especially the `position()` seems useful to me to help the user understand which barcode has just been found. There might be more than one which would lead to confusion.

core/CMakeLists.txt
```diff
@@ -462,13 +463,34 @@ if (PROJECT_VERSION)
...
+  XCODE_ATTRIBUTE_MODULEMAP_FILE "wrappers/ios/Sources/Wrapper/module.modulemap"
+  XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES
+  MACOSX_FRAMEWORK_IDENTIFIER "com.zxing_cpp.ios"
+  CMAKE_XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH NO
+  #MACOSX_FRAMEWORK_INFO_PLIST Info.plist
+  PUBLIC_HEADER "${PUBLIC_HEADERS}"
+  XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "iPhone Developer"
+  XCODE_ATTRIBUTE_ENABLE_BITCODE "YES"
+)
+
```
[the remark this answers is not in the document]
axxel: I took care of that during the conflict resolution with my latest changes.

ryantrem:
I'm looking at possibly using zxing-cpp for cross platform barcode scanning, so I was happy to see this PR 🎉. I just checked out this branch and tried to follow the steps and ran into a few issues though:
1. There were somewhere around 550 generated build files that were being tracked by git. I added a `.gitignore` file to the wrappers/ios directory (similar to the other wrapper directories) with the following:
```
.swiftpm
_builds
ZXing.xcframework
xcuserdata
```
2. The pbxproj has your DEVELOPMENT_TEAM hard coded in it (the "Team" under "Signing & Capabilities" in Xcode). I think it would be a little better to set the Team to "None" and save the file that way, which makes it more clear that a Team needs to be specified.
3. The BitArray.h file is missing from the `ZXing.xcframework/ios-*/ZXing.framework/Headers` directories and so the build fails for the sample app. If I manually copy that file over, then it compiles successfully. I'm guessing something is supposed to be copying this file but is not for some reason.
4. The sample app crashes on launch for me with `Library not loaded: <repo root>/zxing-cpp/wrappers/ios/_builds/core/Release-iphoneos/ZXing.framework/ZXing`. It seems like the lib is either not loading or not present. This is where I got stuck, so not sure if there are any other issues beyond this!

axxel:
@ryantrem Thanks for joining the discussion and sharing your experience. Since I can't test the code myself, it is highly appreciated to have some 'external' feedback. I'd like to say something about 2 of your points:

 * ad 1. sounds good, should be included one way or the other (cmake generally prefers out-of-source builds but on the android side we have a similar issue, so if this is how the iOS projects are to be setup...)
  * ad 3. That is a compile regression that I accidentally introduced with my merge commit. I believe the right way forward is to remove the dependency on this header inside the writer implementation by not using the `BitMatrix::getRow()` function and access the bits directly via `BitMatrix::get(x, y)`.

I'll let the authors comment on the other two.

wrappers/ios/Package.swift
```diff
@@ -0,0 +1,31 @@
...
+    ],
+    products: [
+        .library(
+            name: "ZXingWrapper",
+            type: .static,
+            targets: ["ZXingWrapper"])
+    ],
+    targets: [
+        .binaryTarget(
+            name: "ZXing",
```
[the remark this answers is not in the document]
ChristianNorbertBraun: Done

wrappers/ios/README.md
```diff
@@ -0,0 +1,14 @@
+# ZXing-C++ iOS Framework
```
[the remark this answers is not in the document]
ChristianNorbertBraun: Done

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
```diff
@@ -0,0 +1,109 @@
...
+    if(result.status() == DecodeStatus::NoError) {
+        const std::wstring &resultText = result.text();
+        NSString *text = [[NSString alloc] initWithBytes:resultText.data()
+                                                  length:resultText.size() * sizeof(wchar_t)
+                                                encoding:NSUTF32LittleEndianStringEncoding];
+
+        std::string s(resultText.begin(), resultText.end());
+        NSData *rawBytes = [[NSData alloc] initWithBytes:s.c_str() length:s.size()];
```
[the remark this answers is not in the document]
ChristianNorbertBraun: Done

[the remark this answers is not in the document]
axxel: please merge with current HEAD where this function was renamed to `bytes()` and adapt accordingly. Thanks.

wrappers/ios/Sources/Wrapper/Reader/ZXIDecodeHints.h
```diff
@@ -9,10 +9,16 @@ NS_ASSUME_NONNULL_BEGIN
 @interface ZXIDecodeHints : NSObject
 @property(nonatomic) BOOL tryHarder;
 @property(nonatomic) BOOL tryRotate;
+@property(nonatomic) BOOL tryDownscale;
+@property(nonatomic) BOOL maxNumberOfSymbols;
```
axxel, line 13: I don't know ObjectiveC but that `BOOL` seems wrong.

wrappers/ios/demo/demo.xcodeproj/project.pbxproj
```diff
@@ -316,7 +316,7 @@
 				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = 25MXZ8DDDM;
```
axxel, line 319: I guess you accidentally reverted your earlier change here..

ChristianNorbertBraun: Yes you are right 🙈I'll take care of this tomorrow.

wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
```diff
@@ -0,0 +1,27 @@
...
+                      width:(int)width
+                     height:(int)height
+                     format:(ZXIFormat)format
+                      error:(NSError **)error;
+
+-(nullable CGImageRef)write:(NSString *)contents
+                      width:(int)width
+                     height:(int)height
+                     format:(ZXIFormat)format
+                    charset:(ZXICharset)charset
```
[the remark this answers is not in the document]
ChristianNorbertBraun: Not sure what the requirements exactly are. I will have a talk with my colleagues about this and come back to this probably next week.

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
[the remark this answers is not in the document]
axxel: while you are at it, I suggest to use the term `bytes` instead of `rawBytes` for sake of consistency.

axxel:
@ChristianNorbertBraun thanks for picking this up again. Sorry for being so blunt but we seem to move 2 steps forward and then 1 step back:
 * You partly reverted changes I made recently (completely unrelated to your wrapper, see `Content` and `Result`).
 * The `core/CMakeLists.txt` now seems to contain your additions 3 times.
 * Your new Point related files have the wrong license/copyright comments again.

ChristianNorbertBraun:
@axxel All good, thank you for your feedback. I'm sorry for the unconvenience. I will try to clean up this mess tomorrow.

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
ChristianNorbertBraun: I've looked into this but it's not as straight forward as expected. The [colorspace](https://developer.apple.com/documentation/coregraphics/cgcolorspace/color_space_names?changes=latest_minor) property of a `CGImageRef` does not even have the option to be YUV. The YUV colorspace is only available for pixel buffers which are obtained directly from the capture session. So I think we would have to create a designated API that deals with camera pixel buffers and not `CGImages`. I have to research a bit more on this topic.

axxel: Thanks for looking into that. It might very well be worth it in terms of performance in the long-run. If I profile a standard situation with a single QRCode symbol and process RGB data, then a significant amount of time is spent on the simple RGB to Greyscale conversion (over 30% IIRC), which completely goes away if you can input the `Y` channel directly.

ryantrem:
> @ryantrem Thank you for checking out this PR! I tried to reproduce the error you mentioned in point 4, unfortunately without success. Could you check if the framework is listed within the `Frameworks, Libraries, and Embedded Content` under the `General` tab of the demo application?

Yes it seems to be:
image

ryantrem:
Would folks be open to splitting this PR into two:
1. Just the changes to core/CMakeLists.txt
2. Then the new iOS wrapper

Personally I'm wanting to use zxing-cpp in a cross platform c++ context where I already have things like camera abstractions that provide raw camera frames for different platforms, so I only need #1 (not #2), and #2 seems like the long pull for this PR.

axxel:
@ryantrem you are referring to [this `if (APPLE)` block](https://github.com/nu-book/zxing-cpp/blob/7f6256f32f58d362f4a0029e713a905ef9af9fa6/core/CMakeLists.txt#L504) and maybe line 528? That is referring to a `XCODE_ATTRIBUTE_MODULEMAP_FILE` which would not be present without the rest of the PR, so that will likely not work as expected?

Considering the relatively slow progress we are making here, feel free to provide a PR with changes required for your application and we can figure out how to merge those requirements.

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
axxel: Nice! And that is already used in the demo app? Did you notice any performance difference?

ChristianNorbertBraun: Yes, we tested it with a M1 Mac simulator. The direct passing of the luminance channel seems to be much faster! To make a more accurate statement, we would have to test it on a device. 
The demo app already uses the new interface.

core/CMakeLists.txt
```diff
@@ -465,7 +466,7 @@ add_library (ZXing
 
 target_include_directories (ZXing
     PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>"
-    INTERFACE "$<INSTALL_INTERFACE:include>"
+    INTERFACE "$<INSTALL_INTERFACE:include/ZXing>"
```
axxel, line 469: This will not work. See my recent change, doing exactly this and then reverting it after feedback from @vkrause .

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
```diff
@@ -0,0 +1,151 @@
...
+    return results;
+}
+
+- (NSArray<ZXIResult*> *)readImageView: (ImageView)imageView {
+    Results results = ReadBarcodes(imageView, [ZXIBarcodeReader DecodeHintsFromZXIOptions:self.hints]);
+
+    NSMutableArray* zxiResults = [NSMutableArray array];
+    for (auto result: results) {
+        if(result.status() == DecodeStatus::NoError) {
+            const std::wstring &resultText = result.text();
```
axxel, line 74: Please merge the latest master changes and add the `#define ZX_USE_UTF8` before including `ReadBarcode.h` or in your project settings. See [here](https://github.com/nu-book/zxing-cpp/issues/338#issuecomment-1160989296) for background info. `text()` will return a utf8 encoded `std::string` then (and only that moving forward).

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
```diff
@@ -0,0 +1,151 @@
...
+    CGContextDrawImage(contextRef, CGRectMake(0, 0, cols, rows), image);
+    CGContextRelease(contextRef);
+
+    ImageView imageView = ImageView(
+              static_cast<const uint8_t *>(data.bytes),
+              static_cast<int>(cols),
+              static_cast<int>(rows),
+              ImageFormat::Lum);
+
+    Results results = ReadBarcodes(imageView, [ZXIBarcodeReader DecodeHintsFromZXIOptions:self.hints]);
```
axxel, line 115: This should call the above `readImageView` thereby removing the following code duplication.

wrappers/ios/Sources/Wrapper/Reader/ZXIPosition+Helper.mm
```diff
@@ -0,0 +1,17 @@
+// Copyright 2022 KURZ Digital Solutions GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+
+#import "ZXIPosition+Helper.h"
```
axxel, line 6: That is a bit of weird name. Why not simply put this helper in `ZXIPosition.h`?

wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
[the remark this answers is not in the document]
axxel: That is still an open question.

wrappers/ios/Sources/Wrapper/ZXIFormatHelper.h
```diff
@@ -0,0 +1,13 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#import <Foundation/Foundation.h>
+#import "ZXing/BarcodeFormat.h"
+#import "ZXIFormat.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+ZXing::BarcodeFormat BarcodeFormatFromZXIFormat(ZXIFormat format);
```
axxel, line 11: Same here: can those 2 functions not simply be added to `ZXIFormat.h`?

axxel (changes requested):
This fmtlib-src git submodule that you accidentally added needs to be removed again.

I was looking for the developer team name (or whatever it was called) to see if it was cleared but could not find it anymore.

core/CMakeLists.txt
ChristianNorbertBraun: Reverted

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
ChristianNorbertBraun: Thank you for pointing that out!

wrappers/ios/Sources/Wrapper/Reader/ZXIPosition+Helper.mm
ChristianNorbertBraun: This is a category/extension on ZXIPosition. It's common in Objective-C and swift to make this clear by using a + within the name. 
https://stackoverflow.com/questions/12938577/why-do-some-objective-c-file-names-have-a-minus-or-plus-sign-in-their-name

I have to create an extra file as the api of the category uses one of the types defined within your c++ library. If we want a type to be available in Swift it must not use c++ specific types. 
The hellper file (all of them) are therefore only visible within the wrapper.

ChristianNorbertBraun:
The team was set to none here: https://github.com/nu-book/zxing-cpp/pull/337/commits/7f6256f32f58d362f4a0029e713a905ef9af9fa6

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
```diff
@@ -0,0 +1,129 @@
...
+        .setMaxNumberOfSymbols(hints.maxNumberOfSymbols);
+    return resultingHints;
+}
+
+- (NSArray<ZXIResult*> *)readImageView: (ImageView)imageView {
+    Results results = ReadBarcodes(imageView, [ZXIBarcodeReader DecodeHintsFromZXIOptions:self.hints]);
+
+    NSMutableArray* zxiResults = [NSMutableArray array];
+    for (auto result: results) {
+        if(result.error() == Error::None) {
```
axxel, line 114: This does not compile. The way to go here is using `result.isValid()`.  But if you don't pass `DecodeHints::returnErrors(true)` and the default is not to, then all returned results are valid, so this check can completely be removed.

Also, the next line uses `.c_str()` which will fail if the `text()` property contains a `\0` byte, which may be the case.

wrappers/ios/Sources/Wrapper/Reader/ZXIPosition+Helper.mm
axxel: Thanks for the explanation.

wrappers/ios/Sources/Wrapper/ZXIFormat.h
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#import <Foundation/Foundation.h>
+#ifndef ZXIFormat_h
+#define ZXIFormat_h
+
+typedef NS_ENUM(NSInteger, ZXIFormat) {
+    NONE, AZTEC, CODABAR, CODE_39, CODE_93, CODE_128, DATA_BAR, DATA_BAR_EXPANDED,
+    DATA_MATRIX, EAN_8, EAN_13, ITF, MAXICODE, PDF_417, QR_CODE, MICRO_QR_CODE, UPC_A, UPC_E,
+    ONE_D_CODES, TWO_D_CODES, ANY
```
axxel, line 12: I just recently replaced the terms "OneD"/"TwoD" with "Linear" and "Matrix" as they are used in the standard specs. Please follow that change. See also the deprecation warning you currently get.

wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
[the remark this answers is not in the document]
ChristianNorbertBraun: Could you be more specific about what you mean by the interface could change? Should the possibility to pass a charset be completely removed? Otherwise the enum could be adapted at any time.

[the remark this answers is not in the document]
axxel: Maybe removed or replaced by a completely different set of constants. To get an idea where this might be going, see #332. I spent so much time in the last 4 weeks thinking about and taking care to not break existing code or make the transition smooth. So adding stuff to the public API comes at a considerable cost. If you need that feature right now, I'd like to understand what for. And if you don't, then I'd like it to see removed from the PR.

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
axxel: Very nice. (I only saw your feedback now...)

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
ChristianNorbertBraun: Thanks for pointing this out.

Regarding the `c_str()`. I'm afraid I don't quite understand. The function call actually expects a null terminated string. And doesn't a cstring contain a `\0`byte per definition?

axxel: Yes, a cstring ends with a `\0` but the point is that the utf8 encoded content here may contain a `\0` byte somewhere in the middle of the string. That is uncommon but possible and might lead to weird/silent bugs down the line. Better to fix it right here. Please use [`initWithBytes`](https://developer.apple.com/documentation/foundation/nsstring/1407339-initwithbytes) with `NSUTF8StringEncoding`. That should do the trick.

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
```diff
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#define ZX_USE_UTF8
```
axxel, line 5: Sorry for the back and forth but since this PR was too late for 1.4.0 and will now be part of 2.0 (I hope) this define is not necessary anymore and would merely be noise.

wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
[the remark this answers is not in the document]
ChristianNorbertBraun: I removed the charset api.

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
ChristianNorbertBraun: All good, I removed the define

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
ChristianNorbertBraun: Done

axxel:
Thanks a lot for your continued and cooperative work on this. Let's hope it further broadens the reach of the project.

axxel:
Something I only noticed now: the build time of the ios CI block is longer than any other. The demo-app alone takes 5 minutes to build, which seems excessive also compared to the faster build of the framework which has way more files to process (and is c++ instead of swift/objective-c). Is that also the case on your development machine? Any idea why that is and if something could be done about that? maybe a missing parallel build flag?

ChristianNorbertBraun:
We will take a look into it! Thank you for the collaboration.

ChristianNorbertBraun:
Building the demo app on my machine takes about 1 minute. 
Looking into some of the more recent builds this is also what the ci runner needs:

- https://github.com/nu-book/zxing-cpp/runs/7317496133?check_suite_focus=true
- https://github.com/nu-book/zxing-cpp/runs/7331720273?check_suite_focus=true

So maybe the 5 minute build was some problem with a single runner. As I can not reproduce these long build times I consider this problem as solved.

PR #351, opened 2022-07-04
gcc13: add missing header file

marxin:
Include `<cstdint>`


axxel:
Thanks for the effort but I have an issue with your fix: the header is not the right place to add the include as the missing type is not used in the header. But if that is really all that is required to make g++-13 happy, I can quickly add the one missing line to the cpp file. If you'd rather update your PR (and get the credit as the contributor), even better. :)

marxin:
Sure, I've moved the include to the cpp file.

PR #356, opened 2022-07-08
README: add link to Flutter pub

khoren93:
Hi there. Thanks for this awesome library!

This pull request adds a link to Flutter pub, which I hope someone finds useful. iOS and Android platforms are currently supported.
I'm open to suggestions for other features to be implemented.

Thanks a lot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ It was originally ported from the Java [ZXing Library](https://github.com/zxing/
   * Android
   * WebAssembly
   * [Python](wrappers/python/README.md)
-  * [Flutter](https://pub.dev/packages/flutter_zxing)
+  * [Flutter](https://pub.dev/packages/flutter_zxing) (external project)
 
 ## Supported Formats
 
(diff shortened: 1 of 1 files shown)


axxel:
Interesting. I was not aware of this project. I had a glance at your github repository and noticed that you have a verbatim copy of the library sources inside your ios code. I was wondering whether #337 would change anything regarding this requirement of yours? If there is anything missing there, please let me know. Any other feedback also welcome as I'm not an iOS developer...

Regarding your readme update: what about adding some kind of hint to that link, making it clean that it is an external project?

khoren93:
Thank you for the quick response.

I had to duplicate all the source code in the ios folder due to [CocoaPods restrictions](https://docs.flutter.dev/development/platform-integration/ios/c-interop#step-2-add-cc-sources). 
If anyone can help me get around this limitation, then i can add the pure Dart bindings to this repository without c++ files duplication, if you want)))

As for #337, unfortunately it won't help, because it adds support for native iOS, but my project uses dart FFI technology, which allows it to have one codebase for both iOS and Android.

I added a hint in the readme to make it clear that this is an external project.

axxel:
Thanks for the update. will merge.

About the inclusion of the sources: maybe a git submodule could be of help to remove this duplication inside your project? I'm not really familiar with them and don't know if you could include only parts of another repository. Or maybe simply a symlink?

PR #364, opened 2022-07-27
MCDecoder: allow for dodgy data in primary message

gitlost:
For `MCDecoder` allow for dodgy data in primary message, including postcode data/length mismatches and out-of-range postcode/country/class values.


gitlost:
Bah forgot a line change, will re-submit.

PR #376, opened 2022-08-09
ODReader: check next size before shifting & extending (next may be null)

gitlost:
ODReader: check next size before shifting & extending (next may be null, see `DataBarReader::decodePattern()`).

Adds `BUILD_SANITIZE` option to CMake and github ci job "build-macOS-sanitize job" (using clang as its sanitizer catches it).


axxel:
Thanks a lot for the initiative. I looked into the issue to understand what is going wrong (found that nothing was going wrong but decided to make the sanitizer happy nonetheless, so we can use it at all). I decided to apply a more general fix though and also wanted to change other details about your solution, so decided there was no point in asking you to change your PR but simply commit my take on it. Hope you don't mind.

PR #381, opened 2022-08-16
Fix C4996 when compiled with MSVC

duanqn:
ByteArray.h is a installed header and it triggers MSVC warning C4996 when used in other projects. This PR fixes it by using MSVC-specific alternatives when they are present.

### DIFF
--- a/core/src/ByteArray.h
+++ b/core/src/ByteArray.h
@@ -38,11 +38,13 @@ inline std::string ToHex(const ByteArray& bytes)
 	std::string res(bytes.size() * 3, ' ');
 
 	for (size_t i = 0; i < bytes.size(); ++i)
+	{
 #ifdef _MSC_VER
-		sprintf_s(&res[i * 3], 3, "%02X ", bytes[i]);
+		sprintf_s(&res[i * 3], 4, "%02X ", bytes[i]);
 #else
 		sprintf(&res[i * 3], "%02X ", bytes[i]);
 #endif
+	}
 
 	return res.substr(0, res.size()-1);
 }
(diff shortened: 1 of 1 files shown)


axxel:
The windows CI test fails. Could you please investigate what went wrong?

duanqn:
> The windows CI test fails. Could you please investigate what went wrong?

I read the log but I honestly have no idea about it. Do you know how to run the tests locally?

gitlost:
My guess is it's picking up on the need for `+ 1` on the buffer alloc to allow for NUL-termination by `sprintf()`. Either way I think it would be better if the function wasn't inline.

duanqn:
I tried to run the same command sequence on my local computer. It succeeded.
image

duanqn:
Before the change:
image

After the change:
image

So yeah, I think @gitlost is correct. It crashed because there is an overflow when `sprintf` tried to append the null character.

duanqn:
Should I just `+1` to the string size?

axxel:
> Should I just `+1` to the string size?

Yes please. That is definitively a buffer overflow that I caused there. What is puzzling me is why the ASAN is not picking up on it.

To execute the tests locally, just execute `ctest`.

axxel:
Thanks for the update. Looking at your change and reading the documentation of `sprintf_s` I now understand why the ASAN did not complain about the earlier code: there was nothing to complain about, it was perfectly fine. A `std::string` of size N has a buffer of at least N+1 bytes because it is also terminated by a `\0`. So my "I've screwed up" conviction was premature. The problem was the wrong second parameter `3` to `sprintf_s` in your original patch. Could you please revert all the partly misleading comments and changes and simply use the original code with a 4 instad of a 3. Thanks.

duanqn:
OK. Sorry I'm not very familiar with `sprintf` stuff :/

gitlost:
My "guess" confused everyone, apologies.

PR #387, opened 2022-08-24
Support reading inverted barcodes

markusfisch:
Because inverted barcodes are fairly common (especially QR Codes) and there should be a general option to read these. Otherwise, we have to invert the input image in all clients what is not always trivial or as performant as doing it in the library.

This commit adds a new hint "invert" which makes the threshold algorithms return an inverted binary image so inverted codes can be read. Please note this means *only* inverted barcodes can be read when "invert" is set. It's not an also-try-this type of hint.

The client is supposed to control when to flip an image, because reading inverted barcodes doubles the processing time. In mobile environments it may therefore be better to just flip every second camera frame instead of trying to read positive and negative codes in one frame.

Of course, this is just a first proposal to get some feedback of yours on the matter. I've read https://github.com/nu-book/zxing-cpp/issues/172 and tried to add this feature in the least invasive way I was able to come up with. What do you think?

Maybe it would be nice to also add an additional "tryInvert" hint to do the double processing, too. Would make it easier for all clients that want to do both 🤔


axxel:
Thanks for this initiative. I have 2 concerns with your approach:
 1. introducing a non-const function pointer dereference (which is not inline-able to my knowledge) in the innermost loop is a considerable performance pessimization.
 2. I assume that the relative importance of the use cases or the likelihood of them occurring is `E(non-inverted) >> E(both) >> E(inverted)`. This would suggest to emphasize speed more on the `try-both` use-case than the `inverted-only` one.

From those two premises, I conclude that your approach is no the one I'd take, because trying both with your approach would mean to have the binariazation being performed twice. Especially in the linear-codes use-case this a considerable part of the total runtime. And especially there it is trivial to modify the result of the binarization such that it is inverted (implementation knowledge that would unlikely have had).

In terms of interface I was thinking about something like `tryInvert(InversionMode)` with options like `No|Yes|Only` (or some other set of words mapping to the 3 states of either one, the other or both. If you have a suggestion that is better in terms of selfdocumenting, let me know.

markusfisch:
I understand and fully share your concerns. I should have included a performance analysis right away to show how much this change would affect performance. And after looking at it more closely, I have to admit that it's probably really not the best idea. I just found it charming because it would have no extra cost - but the function invocation in the innermost loop is much worse indeed.

So I will close this PR and start a new one.

axxel:
> So I will close this PR and start a new one.

Before you starting working on another direction that might again not be convincing, I suggest we first agree on where this should be going. I'd be especially interested in your thoughts on the best API for this.

As for the implementation, I'll need to revisit my ideas that I had a couple months back when thought about how to most efficiently implement this.

markusfisch:
> Before you starting working on another direction that might again not be convincing, I suggest we first agree on where this should be going. I'd be especially interested in your thoughts on the best API for this.

Good idea! 👍 Of course, I'm already considering other approaches but it's always good to talk and think about it first.

About the API, I like your `tryInvert(InversionMode)` suggestion. This would make pretty much the ideal interface I think. It's explicit, and it would still allow alternating use on every second camera frame.

> As for the implementation, I'll need to revisit my ideas that I had a couple months back when thought about how to most efficiently implement this.

Please go ahead!

At the moment, I just see these options:

* Inverting the `LumImage`, maybe in [SetupLumImageView](https://github.com/zxing-cpp/zxing-cpp/blob/master/core/src/ReadBarcode.cpp#L89). But it's also not a cheap way.
* Somehow flipping the `BitMatrix` from `BinaryBitmap::getBitMatrix()` **and** the `PatternRow` from `BinaryBitmap::getPatternRow()`.

PR #396, opened 2022-08-31
Include version number of QR Code and DataMatrix in Result

markusfisch:
Because it's helpful to know which version was actually used for a particular code. We're working with QR Codes a lot and it's often useful to be able to check the exact version.

I realize this is a pretty niche request, but it would have some value for us 😉

### DIFF
--- a/core/src/DecoderResult.h
+++ b/core/src/DecoderResult.h
@@ -25,6 +25,7 @@ class DecoderResult
 	Content _content;
 	std::string _ecLevel;
 	int _lineCount = 0;
+	int _versionNumber = 0;
 	StructuredAppendInfo _structuredAppend;
 	bool _isMirrored = false;
 	bool _readerInit = false;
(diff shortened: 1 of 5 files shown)


axxel:
I thought about adding something like that a while back as well. I would have used a different type though, namely a string. That is because the version attribute of a QRCode can actually contain letters (MicroQRCodes are simply QRCodes with version strings starting with 'M'). I'll simply merge your commit and later change the public API before the next release, so be prepared for that.

markusfisch:
Great! Thank you! 😄 

I used an `int` because that's the type in `Version` - but the MicroQR argument makes a lot of sense, of course. Thanks for the warning 😉

axxel:
I just realized (after the 2.0 release unfortunately) that the DataMatrix specification actually does not have a notion of a _version number_. That number is more an implementation detail of the library but does not translate to "googleable" information like the QRCode version does. It seems to me that the right information to be returned here is the size of the symbol, e.g. `22x22` or `8x18` (see also https://www.activebarcode.de/codes/datamatrix_examples).

Having typed this text, I just realized that a more general information than this "version" could have been the size of the symbol (width and height) in modules right from the start. That would be pretty much self-explanatory and also work for all currently supported 2D symbologies.

Since you proposed it initially: how do you use that version information?

markusfisch:
In my company we actually use the (QR Code) version number always in conjunction with the amount of modules, which we compute from the version number (`versionNumber * 4 + 17`).

This is useful, for example, when we have to create QR Codes in very limited space (on packages and documents) that still have to be readable with normal smartphones. Knowing the exact number of modules makes this process easier and faster. We also found it quite useful when we have to check (and compare) QR Codes from other solutions.

So, yes, a more general information giving just the size of the read symbol in modules would be just as good! And it would probably make more sense for other barcode formats too.

PR #408, opened 2022-09-09
Add some more DeocdeHints to the iOS-Wrapper

parallaxe:
There are a few more hints that are missing in the wrapper (i.e. isPure, returnCodabarStartEnd) that would need more effort and can be added sometimes else.


axxel:
Thanks for your work. I trust that the code does what you intended it to do :).

Issue #375, opened 2022-08-09
1.4.0 dylibs not generated on Mac platforms


davesworking:
We recently changed from 1.3.0 to 1.4.0. On the Mac platforms, the dylibs are not generated for version 1.4.0. CMake log shows:

`Linking CXX shared library ZXing.framework/ZXing`

Is this correct? Is there a way to generate dylibs?

Thanks!
Dave

axxel:
I was not aware of any change that would cause/explain a different behavior with regards to dylibs between 1.3 and 1.4. I checked the diff (`git diff v1.3.0 v1.4.0 core/CMakeLists.txt`) and that does not explain it either to me. There has been a change _after_ 1.4.0 that might be affecting that, though (see `if(APPLE)` block in #337).

Bottom line: no idea what is going on here and I can't dive deeper into that, since I don't have a mac setup.

davesworking:
The if(APPLE) block was the issue. I commented it out. Thanks!

axxel:
So linking against the Framework is not an option for you? I'd like to prevent more issue reports like this before releasing 2.0. Do we need a configurable option to switch between dylib and framework generation? @markusfisch any comment on that?

davesworking:
The comment I received from our developer was : "We _could_ use the current framework build of zxing but it's significantly harder than the usual headers+dylib"

markusfisch:
Well, I have to forward this question to @ChristianNorbertBraun who did the iOS wrapper.

axxel:
@markusfisch right. Sorry for somehow confusing you, too. very unprofessional... ;-/

markusfisch:
@axxel No problem, after all we work for the same company and I already pinged him to have a look 😉

parallaxe:
As co-author of the ios-wrapper and in deputy of Chris: @davesworking We don't see how and why a framework should be harder to use than headers+dylib in anyway, let alone *significantly*. Frameworks bundle the binary, headers, modulemap and more in a standardised way and are the standard to distribute libraries on Apple Systems.

Can you please elaborate on the problems?

davesworking:
sorry, I am still fairly new to the company so I don't know what process the Developers use to include libraries. I don't have insight into what is more difficult. I am building the dylibs now and everything is fine. Thanks!

axxel:
[...]

PR #415, opened 2022-09-21
Conditionally include iOS-Wrapper project

Alex-MSFT:
### Description
This PR adds a new compile definition check for conditionally including the iOS wrappers project, and adds it to build-release.sh.

Using the iOS wrappers added in #337 requires running an external build script, and integrating another universal framework into the consuming project.  This adds some extra toolchain and integration complexity for projects, which do not make use of the iOS Swift/Objective-C wrappers.

I also had to disable bitcode to get build-release.sh to successfully build libzueci.

### Validation
Ran build-release.sh locally and validated the demo app on device.

Built and ran a ZXing-cpp enabled iOS app which does not include the iOS wrappers and validated its behavior on both iOS and Android.


axxel:
Thanks for your effort to contribute. Looks like you had (and fixed) the same issue as discussed in #375? As mentioned there, I'm happy to include an option to en/disable the framework setting if convinced that this is useful. You wrote that with this option "integrating another universal framework into the consuming project" is necessary. Can you elaborate? See also @parallaxe's [comment](https://github.com/zxing-cpp/zxing-cpp/issues/375#issuecomment-1210871844).

Either way, I'm not happy with your choice of words here: `INCLUDE_IOS_WRAPPERS`. The main effect this setting has is the `FRAMEWORK TRUE` line. So a more suitable name would e.g. `BUILD_XCODE_FRAMEWORK` or `BUILD_APPLE_FRAMEWORK` be.

Alex-MSFT:
Thanks for the response!  I did not see that this issue was already discussed in #375, but this is indeed addressing the same issue.

> You wrote that with this option "integrating another universal framework into the consuming project" is necessary. Can you elaborate? See also @parallaxe's [comment](https://github.com/zxing-cpp/zxing-cpp/issues/375#issuecomment-1210871844).

The main pain point for our integration is the necessity to run the build-release.sh script to generate the framework outside of CMake, this adds some extra work for integrating it into our dev and build pipeline which currently calls CMake directly to build the XCode project.

Additionally, since the barcode reader component of our app is written directly in C++ without having platform specific implementations there's no need for us to include the framework as we can consume the ZXing C++ code directly as a pure C++ library and share the same code across all platforms. Even if we were to switch over to using the framework we would not opt to use the Apple specific wrapper code, so this just ends up including unneeded Objective-C code in our project. Where I think the framework has strong value is for integrating into iOS specific implementations given its nice integration with Swift/Objective-C.

> Either way, I'm not happy with your choice of words here: `INCLUDE_IOS_WRAPPERS`. The main effect this setting has is the `FRAMEWORK TRUE` line. So a more suitable name would e.g. `BUILD_XCODE_FRAMEWORK` or `BUILD_APPLE_FRAMEWORK` be.

I'm totally fine with renaming the definition, and those suggestions make sense to me. I would probably go with BUILD_APPLE_FRAMEWORK.

ryantrem:
Hey @axxel, just wanted to add my voice on this one. The main issue is that the iOS wrapper introduced language bindings (Swift/Objective-C) for use on iOS, and the changes to the core CMakeLists.txt now *require* you to build the language bindings in order to build the core code on for any Apple target. This means even if you have no need of the language bindings (Swift/Objective-C), you still have to build them to consume zxing-cpp in a project that builds for Apple platforms. As @Alex-MSFT said, in our case we are using zxing-cpp in platform agnostic code that is pure C++ with no Swift/Objective-C (or Java, or C#, or any other platform specific languages). In this case, there is no reason to build Swift/Objective-C language bindings, as it is just building extra platform specific code that won't be used.

axxel:
Thanks @Alex-MSFT and @ryantrem for the details but there is one thing I don't understand: I thought the question of whether or not a framework is build is independent of whether or not any ObjectiveC language bindings are build? In that case some other property of those specified in the `if (APPLE)` block is actually the one that you want to avoid?

axxel:
@parallaxe: do you have anything to say about the proposed `XCODE_ATTRIBUTE_ENABLE_BITCODE "NO"` change?

Alex-MSFT:
> Thanks @Alex-MSFT and @ryantrem for the details but there is one thing I don't understand: I thought the question of whether or not a framework is build is independent of whether or not any ObjectiveC language bindings are build? In that case some other property of those specified in the `if (APPLE)` block is actually the one that you want to avoid?

For just excluding the iOS wrappers I believe the main part that needs to be omitted is the inclusion of the wrappers module map `XCODE_ATTRIBUTE_MODULEMAP_FILE "wrappers/ios/Sources/Wrapper/module.modulemap"`.  Moving the block to just this line does allow the project to build without having to manually generate the framework by running build-release.sh.

I opted to make the flag cover the whole iOS block for a few reasons:
- Building and linking Xcode frameworks works slightly differently in cmake, so giving the option to use a library may make the integration a little bit cleaner and match more closely with other iOS libraries for certain projects.
- We can side step hard coding things like enabling bit code, ARC, and which architectures to include that may differ based on the consuming project.
- Consumers don't need to worry about signing settings for the library, which must be manually set when using a framework.

My personal opinion is that it is best to limit platform specific integration details where possible to keep integrating the project as similar as possible across platforms, and to avoid potential conflicts with platform specific configurations in consuming projects.

parallaxe:
Sounds fine for me. Our use case for integrating zxing into iOS-apps obviously does not cover all use-cases of zxing on Apple-platforms, and the reasonings in the discussions make sense.
Regarding enabling / disabling / side stepping bitcode: Since Xcode 14 / iOS 16 (released in September this year), bitcode is deprecated. Previously, it was recommended to build apps and frameworks with bitcode turned on but not enforced.
Thus, I prefer to turn bitcode `off` per default.

axxel:
Thanks to everyone for the valuable input on this.

PR #441, opened 2022-11-21
Fix narrow_cast in threshold binarizer

duanqn:
The narrow_cast is applied too early. `p - lastPos` can be out of `uint16_6` range. For example when the original image is 800x600 and there is a 150-pixel consecutive range of white pixels, then `p - lastPos` can be 150x600=90000 which is larger than 65535.


axxel:
Thanks.

PR #446, opened 2022-11-28
Avoid big-scary cmake warning for optional lib

zander:
As the lib is optional, lets avoid the big CMake warning when the opencv library is not found. It mostly has info for the maintainers, not for people building this.

Instead add a one line message stating the cause and effect.


axxel:
Question: do you have Qt installed or does the complaint from the `find_package(Qt5 COMPONENTS Gui Multimedia Quick)` line not annoy you as much as the `OpenCV` one does?

zander:
Frankly, I didn't notice Qt not being a non-optional requirement until after I made this PR.

So, to answer your question, I have it installed.

axxel:
I see. I checked the cmake output of your PR on the CI system where there is neither Qt nor OpenCV installed: https://github.com/zxing-cpp/zxing-cpp/actions/runs/3564843023/jobs/6010005518

The output now looks like:
```
CMake Warning at example/CMakeLists.txt:26 (find_package):
  By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt5" with any of
  the following names:

    Qt5Config.cmake
    qt5-config.cmake

  Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
  to a directory containing one of the above files.  If "Qt5" provides a
  separate development package or SDK, be sure it has been installed.


INFO: Missing OpenCV dependency, skipping building of example ZXingOpenCV
```
Would you be willing to ammend your PR to do the same thing for Qt, so we get a consistent behavior?

example/CMakeLists.txt
```diff
@@ -23,7 +23,13 @@ if (BUILD_READERS)
 endif()
 
 if (BUILD_READERS)
-    find_package(Qt5 COMPONENTS Gui Multimedia Quick)
+    find_package(QT NAMES Qt5 Qt6 REQUIRED COMPONENTS Gui Multimedia Quick)
```
axxel, line 26: I assume that the `REQUIRED` clause breaks the build if there is no qt: https://github.com/zxing-cpp/zxing-cpp/actions/runs/3585591723/jobs/6040040429.

I'd also prefer not to mix the message cosmetic with a change of the used Qt version in one commit.

axxel:
Thanks for the update. It builds again. Looking at the output now:
```
INFO: Qt not found, skipping examples based on them. (need modules Multimedia + Quick)
INFO: Missing OpenCV dependency, skipping building of example ZXingOpenCV
```
I'd like to kindly ask to make one minor improvement: please use the same phrasing for both cases as they convey the exact same kind of information. Here is my proposal:
```
INFO: Qt (Gui/Multimedia/Quick) not found, skipping Qt examples
INFO: OpenCV not found, skipping ZXingOpenCV example
```

axxel:
Awesome. Thanks for your contribution (and patience ;)).

PR #449, opened 2022-12-01
Make Qt6 based example actually work.

zander:
Without 'active' the camera is never started.


axxel:
The sample works just fine on my machine. Since it works even when I set `active: false` I assume there is something going on with my webcam setup, so I believe your PR is helping even if I can not reproduce it. -> will merge.

PR #451, opened 2022-12-01
Avoid crash

zander:
I've observed that a camera frame which is a QVideoFrame with format jpeg becomes a QImage of invalid type on convert.
Not sure why, likely some corruption in the frame.

This catches such issues early and avoids the deeper code throwing an uncaught exception.

### DIFF
--- a/example/ZXingQtReader.h
+++ b/example/ZXingQtReader.h
@@ -143,6 +143,8 @@ inline QList<Result> QListResults(ZXing::Results&& zxres)
 inline QList<Result> ReadBarcodes(const QImage& img, const DecodeHints& hints = {})
 {
 	using namespace ZXing;
+    if (img.format() == QImage::Format_Invalid)
+        return QList<Result>();
 
 	auto ImgFmtFromQImg = [](const QImage& img) {
 		switch (img.format()) {


zander:
> I believe it is better to let the client code know that it provided invalid data

If a single frame in a video feed is not capable of being decoded, I don't qualify that as "client provided invalid data". Your example is built around the slot simply taking data from the Qt video sink and forward that.
The next frame is fine, so... :shrug: 

Alternatively you can close this PR and put a try-catch around the call (again, in the slot that handles the sink data). Personally I think that is wrong, I think an early detection and return is the way to go. Hence this PR.

axxel:
Are you using Qt5 or Qt6?

zander:
I'm using Qt6.

axxel:
Please check if my recent change solves your problem.

zander:
Since you found a different way, I'll close this PR.

Just want to point out that for Qt5 you do not check the resulting qimage, which can still fail due to a frame being corrupted.

axxel:
> Just want to point out that for Qt5 you do not check the resulting qimage, which can still fail due to a frame being corrupted.

You are right. Fixed that as well now. Thanks.

PR #453, opened 2022-12-02
Avoid mapping when not calling the bits() method

zander:
As the docs says, in some cases doing a 'map' is expensive. So don't do it unless you need to actually call bits()

Tested on Qt5 and Q6, ensuring I run via the codepath that does not do the mapping.

### DIFF
--- a/example/ZXingQtReader.h
+++ b/example/ZXingQtReader.h
@@ -180,17 +180,6 @@ inline QList<Result> ReadBarcodes(const QVideoFrame& frame, const DecodeHints& h
 {
 	using namespace ZXing;
 
-	auto img = frame; // shallow copy just get access to non-const map() function
-#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
-	if (!frame.isValid() || !img.map(QAbstractVideoBuffer::ReadOnly)){
-#else
-	if (!frame.isValid() || !img.map(QVideoFrame::ReadOnly)){
-#endif
-		qWarning() << "invalid QVideoFrame: could not map memory";
-		return {};
-	}
-	auto unmap = qScopeGuard([&] { img.unmap(); });
-
 	ImageFormat fmt = ImageFormat::None;
 	int pixStride = 0;
 	int pixOffset = 0;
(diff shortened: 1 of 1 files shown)


example/ZXingQtReader.h
```diff
@@ -272,17 +261,29 @@ inline QList<Result> ReadBarcodes(const QVideoFrame& frame, const DecodeHints& h
...
+#endif
+			qWarning() << "invalid QVideoFrame: could not map memory";
+			return {};
+		}
+		QScopeGuard unmap([&] { img.unmap(); });
+
 		return QListResults(ZXing::ReadBarcodes(
 			{img.bits(FIRST_PLANE) + pixOffset, img.width(), img.height(), fmt, img.bytesPerLine(FIRST_PLANE), pixStride}, hints));
-	} else {
+	}
```
axxel, line 277: Thanks for this PR. I'll merge this as is but please try to avoid unrelated code style changes like this in the future.

zander: The change is related.
The number of lines in the if/then/else grew too large to allow it to be 1 unit that would be collapsed. It would have been bad practice to allow one collapsing codeblock of the new size.

axxel: I choose to disagree. This is purely a code style change that goes against the style that is specified in the clang-format config of the project.

zander: The argument that its bad because it violates the automatic code-formatting tool is likely the worst reason possible.

My argument was that it helps readability. Now, you can disagree on that natually. The coding style is personal, afterall. 

Making a decsion based on some tool instead of "its more readable for the majority of potential contributors" is interesting. I wonder if I can convince you to re-think that point of view in a much used open source project. It might help that you have 10x commits compared to the 2nd most active contributor.

Either way, all my bugs are fixed, thank you for this project!

axxel: > The argument that its bad because it violates the automatic code-formatting tool is likely the worst reason possible.

I'd like to respectfully make you aware of the fact that your style of communication is perceived as rather condescending ("unreadable code"?!?). As the maintainer of this code base I find it interesting to be exposed to this kind of feedback and an interesting challenge to find the most constructive response but I feel that you could be more successful with contributing to other peoples code bases if you selected your words a bit more carefully when criticizing their work, especially when it is related to subjective metrics.

My argument here is that this style change is against the styling rules of this project (that manifest themself in said code formatting config file, which I deliberately authored). I claim that the overall readability of the code is higher with a consistent style, no matter which individual style is subjectively considered most readable. That said, the code base as a whole is definitively not consistent, yet.

> Either way, all my bugs are fixed, thank you for this project!

Glad to hear you find this project useful. Half of the reason I spent so much time over the years on this is that I hoped people would find it useful.

PR #455, opened 2022-12-06
Do not require targets on Windows to find dependencies.

correa:
On Windows with MSVC the users of the library won't link with pthreads and usually won't do find_package(Threads) on their own. Exposing Threads::Threads as a public dependency allows CMake to propagate the dependency and avoid erros such as this one on windows: CMake Error at CMakeLists.txt:69 (add_executable):
  Target "visao-tg" links to target "Threads::Threads" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?

### DIFF
--- a/core/ZXingConfig.cmake.in
+++ b/core/ZXingConfig.cmake.in
@@ -1,6 +1,9 @@
 @PACKAGE_INIT@
 include("${CMAKE_CURRENT_LIST_DIR}/ZXingTargets.cmake")
 
+include(CMakeFindDependencyMacro)
+find_dependency(Threads REQUIRED)
+
 # this does not work: add_library(ZXing::Core ALIAS ZXing::ZXing)
 # this is a workaround available since 3.11 :
 if(NOT(CMAKE_VERSION VERSION_LESS 3.11))


correa:
I don't think it should be limited to compiler and on platform it's just a coincidence that most platforms would already link against pthreads, maybe there are other platforms that don't (?). After further researching the matter, I found [CMake bug #22363](https://gitlab.kitware.com/cmake/cmake/-/issues/22363) and according to [this post] (https://discourse.cmake.org/t/cmake-3-21-rc-1-threads-threads-not-found/3630/4) it's reproducible even on Linux, according to my colleague it happens also on 3.24 as well. I will try latest and see.

axxel:
I had a look at the two links you posted but fail to see how they apply here or how they corroborate your intended change. In `core/CMakeLists.txt` there is already an explicit linking to `Threads::Threads`.

axxel:
I had another look at the error message and came to the conclusion that you must have run into a cmake bug (see [here](https://discourse.cmake.org/t/cmake-3-21-rc-1-threads-threads-not-found/3630, which links to your mentioned bug). So, while your proposed PR might be a workaround for said cmake bug, I don't believe merging this PR is in the best interest of this project. If there is a bug in your cmake install, please update cmake.

If I should have overlooked oder misunderstood something, please let me know.

PR #458, opened 2022-12-11
add missing #include <cmath> for std::abs

th1722:
This fixes and avoids "error: call to 'abs' is ambiguous".

### DIFF
--- a/core/src/HRI.cpp
+++ b/core/src/HRI.cpp
@@ -9,6 +9,7 @@
 
 #include "ZXAlgorithms.h"
 
+#include <cmath>
 #include <sstream>
 
 namespace ZXing {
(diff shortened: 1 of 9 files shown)


axxel:
Thanks for the PR. Makes sens to me. But I have a question: where did you see this error message? (platform/compiler/version?)

th1722:
> Thanks for the PR. Makes sens to me. But I have a question: where did you see this error message? (platform/compiler/version?)

platform: Ubuntu 22.10
compiler: Android NDK r19(19.2.5345600)/r20(20.1.5948944)
version: 1.4.0

I received that error message when I built LibreOffice with zxing-cpp.

th1722:
> Thanks for the PR. Makes sens to me. But I have a question: where did you see this error message? (platform/compiler/version?)

platform: Ubuntu 22.10
compiler: Android SDK 33, Android NDK r19c(19.2.5345600) / r20c(20.1.5948944)
version: 1.4.0

I received that error message when I built LibreOffice with zxing-cpp.
However, that error occurred only in building PDFEncoder.cpp.

PR #460, opened 2022-12-12
Ignore swp files

markusfisch:
Ignore Vim's swap files too.

Completely unimportant of course, and probably it's just me that's affected, so I completely understand if you don't care about this. Still worth a try 😉


axxel:
Looks like there is popular demand for Vim... :)
While I was at it resolving the conflict, I decided to add the one line from #621 as well... ;)

PR #463, opened 2022-12-16
c++20: add missing include

th1722:
Needed for std::movable.

### DIFF
--- a/core/src/Generator.h
+++ b/core/src/Generator.h
@@ -15,6 +15,7 @@ namespace std {
 	struct default_sentinel_t {};
 }
 #else
+#include <concepts>
 #include <coroutine>
 #endif
 


axxel:
Question: does this solve a compilation error on your setup?

The `#include` would need to move 3 lines up into the `#else` block to not (potentially) break the Android NDK support again.

axxel:
The force-pushed changes you provided last week don't address my concern. Are you planing to modify the PR again?

th1722:
> Question: does this solve a compilation error on your setup?
Yes, this solves "error: no type named 'movable' in namespace 'std'".

> The `#include` would need to move 3 lines up into the `#else` block to not (potentially) break the Android NDK support again.
> The force-pushed changes you provided last week don't address my concern. Are you planing to modify the PR again?

Fixed.

th1722:
> Question: does this solve a compilation error on your setup?

Yes, this solves "error: no type named 'movable' in namespace 'std'".

> The `#include` would need to move 3 lines up into the `#else` block to not (potentially) break the Android NDK support again.

> The force-pushed changes you provided last week don't address my concern. Are you planing to modify the PR again?

Fixed.

axxel:
Thanks.

PR #467, opened 2022-12-26
Support "NW" style decode for OneD barcode

hellovigoss:
This pr provide a "NW" style reader for OneD barcode.
What N means Normail bar or space, and W means Wide bar or space.
NWCode style decode can be call like this
```c++
hints.setFormats(BarcodeFormat::NWCode);
hints.setTryDownscale(false).setTryInvert(false).setTryRotate(false).setTryHarder(false);
```

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -289,7 +289,9 @@ if (BUILD_READERS)
         src/oned/ODReader.cpp
         src/oned/ODRowReader.h
         src/oned/ODRowReader.cpp
-    )
+            src/oned/ODNWReader.h
+            src/oned/ODNWReader.cpp
+        )
 endif()
 if (BUILD_WRITERS)
     set (ONED_FILES ${ONED_FILES}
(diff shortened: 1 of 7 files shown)


axxel:
I had a glance at your code and it seems its purpose is to generate a text containing of the letters 'N' and 'W' depending on the width of the bars/spaces. Based on that I assume you are interested in debugging other 'real' codes?

hellovigoss:
Our project need origin NW style made me do this thing.
I will be glad to debug 'real' code.

axxel:
I don't know what you mean by "origin NW style". Can you provide a link to some specification or other technical description?

hellovigoss:
ref: https://en.wikipedia.org/wiki/Interleaved_2_of_5

axxel:
Interleaved 2 of 5 is already supported, see `BarcodeFormat::ITF`.

hellovigoss:
got it! thx.

Issue #335, opened 2022-05-09
zxing.pc is broken when CMAKE_INSTALL_{INCLUDE,LIB}DIR is absolute


alexshpilkin:
As per title: `zxing.pc.in` has

```
prefix=@INSTALLDIR@
exec_prefix=${prefix}
libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
```

and so can’t handle absolute paths in `CMAKE_INSTALL_{INCLUDE,LIB}DIR`. This leads to a broken .pc file on NixOS in particular.

Similar to open-source-parsers/jsoncpp#1199 and KhronosGroup/SPIRV-Tools#3905. See “[Concatenating paths when building pkg-config files][1]” for a discussion of the problem and a suggested fix (I don’t know CMake myself, sorry).

[1]: https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files

axxel:
Thanks a lot for the input. (Libreoffice uses this project?!? who would have thought... ;)).

Question: what version of cmake is generally available to your audience? 3.19 or newer by any chance? See jtojnar's comment here https://stackoverflow.com/questions/57382992/joining-paths-in-cmake. And also my question to him: https://github.com/jtojnar/cmake-snips/issues/7.

axxel:
I now discovered your patch
```
      --replace '$'{exec_prefix}/@CMAKE_INSTALL_LIBDIR@ @CMAKE_INSTALL_FULL_LIBDIR@ \
      --replace '$'{prefix}/@CMAKE_INSTALL_INCLUDEDIR@ @CMAKE_INSTALL_FULL_INCLUDEDIR@
```
But based your mentioned [Concatenating paths when building pkg-config files](https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files) doc, this solves one issue while possibly introducing another? (I don't understand the inner workings of the pkg-config workflow related to interpolations.) If I was sure your patch does not break other users/distros, I'd certainly apply it.

alexshpilkin:
@axxel Correct, that patch works for the purposes of Nixpkgs, where the user isn’t really supposed to move built software around and the paths passed to CMake are even guaranteed to be absolute, but for general usage you probably want that join_paths monster jsoncpp used. I don’t understand CMake well enough to advise you on it, unfortunately.

[...]

PR #468, opened 2023-01-01
zxing.pc.in: fix when CMAKE_INSTALL_<dir> is absolute

bjornfor:
pkg-config files are not designed to be relocatable, so the only thing
we need to do is using `CMAKE_INSTALL_FULL_<dir>` to avoid broken path
concatenation when `CMAKE_INSTALL_<dir>` is already absolute (like in
Nixpkgs).

https://cmake.org/cmake/help/v3.25/module/GNUInstallDirs.html

Fixes https://github.com/zxing-cpp/zxing-cpp/issues/335

### DIFF
--- a/zxing.pc.in
+++ b/zxing.pc.in
@@ -1,7 +1,7 @@
 prefix=@INSTALLDIR@
 exec_prefix=${prefix}
-libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
+libdir=@CMAKE_INSTALL_FULL_LIBDIR@
+includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
 
 Name: ZXing
 Description: ZXing library set


bjornfor:
Amended to fix s/xzing/zxing/ typo.

bjornfor:
Thanks!

axxel:
Just noticed, looking at a test-run of `make install`, that my zxing.pc does not look like I expected:
```
prefix=/home/axel/contrib/zxing-cpp.install
exec_prefix=${prefix}
libdir=
includedir=
```
Ideas?

bjornfor:
Oh, sorry, I should have tested before submitting. I think we need to update the cmake code that generates the final .pc file too. Will submit fix later today.

bjornfor:
@axxel: Fix here: https://github.com/zxing-cpp/zxing-cpp/pull/469

(This time I did a build to verify the change.)

axxel:
Thanks. Your fix works but I don't understand why it is required because a) the formerly used non-FULL versions should be defined iff the FULL versions are and b) the fix that has been used and working on the NixOS side was exactly what your original PR did.

Anyway, I reorganized the .pc file and CMakeConfig generation and this takes also care of the missing GNUInstallDirs include. That said, I'd appreciate if you could test the code especially with respect to those changes. I don't have a setup that properly tests the validity of the generated .pc file.

bjornfor:
> Thanks. Your fix works but I don't understand why it is required because a) the formerly used non-FULL versions should be defined iff the FULL versions are and b) the fix that has been used and working on the NixOS side was exactly what your original PR did.

I think cmake itself defines `CMAKE_INSTALL_<dir>`, but require GNUInstallDirs for `CMAKE_INSTALL_FULL_<dir>`. (I don't know about anything this working on NixOS with `-DCMAKE_INSTALL_<dir>=/absolute` without using GNUInstallDirs.)

> Anyway, I reorganized the .pc file and CMakeConfig generation and this takes also care of the missing GNUInstallDirs include. That said, I'd appreciate if you could test the code especially with respect to those changes. I don't have a setup that properly tests the validity of the generated .pc file.

`libdir=` and `includedir=` look fine, and are the only variables  used by the .pc file itself. I don't know if anyone is using `prefix=` (which is now removed).

PR #469, opened 2023-01-02
cmake: include GNUInstallDirs

bjornfor:
So that we get access to `CMAKE_INSTALL_FULL_<dir>` variables, ref.
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

Before:
```
  $ head -4 _install/lib64/pkgconfig/zxing.pc
  prefix=/home/bf/inbox/zxing-cpp/_install
  exec_prefix=${prefix}
  libdir=
  includedir=
```

After:
```
  $ head -4 _install/lib64/pkgconfig/zxing.pc
  prefix=/home/bf/inbox/zxing-cpp/_install
  exec_prefix=${prefix}
  libdir=/home/bf/inbox/zxing-cpp/_install/lib64
  includedir=/home/bf/inbox/zxing-cpp/_install/include
```

Fixes: aacc1bc6857d ("zxing.pc.in: fix when `CMAKE_INSTALL_<dir>` is absolute")

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ if (NOT DEFINED INSTALLDIR)
 endif()
 
 IF (NOT WIN32 OR MINGW)
+    include(GNUInstallDirs)
     configure_file(zxing.pc.in zxing.pc @ONLY)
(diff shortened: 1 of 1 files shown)


axxel:
See my upcoming comment in your last PR.

PR #474, opened 2023-01-04
Keep the result with no error instead of the first found

duanqn:
Finish a TODO item in ReadBarcode.cpp


axxel:
Thanks for looking into TODO items. Looking at your suggestion I noticed that a) your code does not achieve what you intended it to and that is because one valid and one invalid result are never equal according the current implementation of the `Result::operator==`, i.e. the "replace existing one" line will never be executed. And b) it also means the TODO comment is somewhat outdated or misleading.

The question is: when are two results "equal"? I don't recall my reasoning why they should be not-equal even if they contain the same error. That does not make sens to me. I changed that and removed the TODO comment.

PR #476, opened 2023-01-06
fix crash when the source image is less than 3 pixels width/height

liulex:
If the last witdh or height of the last layer is less than `N`, it would create another layer with empty image and cause crash.

### DIFF
--- a/core/src/ReadBarcode.cpp
+++ b/core/src/ReadBarcode.cpp
@@ -76,7 +76,8 @@ public:
 
 		layers.push_back(iv);
 		// TODO: if only matrix codes were considered, then using std::min would be sufficient (see #425)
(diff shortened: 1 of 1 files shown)


axxel:
for that crash to happen you would need to call it with `threshold < 3`, right? That value would make absolutely no sens. The smallest meaningful value would be in the order of 100.

liulex:
No, it's not the threshold, it's the `N`, which has a value of 3 and cannot be configured.

liulex:
Suppose the threshold is 500, a 600x1 image would cause the crash without the fix.

liulex:
Yes.

axxel:
Thanks for the patch and the explanation.

PR #477, opened 2023-01-09
Implementing Feature to Save Images on the Android Wrapper Demo App

Sergio-:
Hello @axxel Good day.
This is some simple modification to Add the Feature to save images on the Android Wrapper Demo App. Please, review and merge it if you think it's a good addition.
Best Regards.

### DIFF
--- a/wrappers/android/.idea/gradle.xml
+++ b/wrappers/android/.idea/gradle.xml
@@ -5,10 +5,9 @@
     <option name="linkedExternalProjectsSettings">
       <GradleProjectSettings>
         <option name="testRunner" value="GRADLE" />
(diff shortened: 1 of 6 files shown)


axxel:
Hi Sergio. Thanks a lot for your contribution. I had a quick look and got the feeling that your PR contains changes that are not related or necessary for the goal of adding the capture and save functionality. Would you be willing and in the position to go over your changes and try to minimize them (e.g. related to permission handling and the xml-file modifications).

axxel:
One comment about the permission thing: if I understood your code correct, there was some logic involved that made sure the code would run on older and newer SDKs. For this sample app, I don't care about backward compatibility and would prefer to have it only run on the latest Android if that makes the diff smaller.

PR #483, opened 2023-01-11
This changes enable the feature of saving images to the phone gallery.

Sergio-:
Hi @axxel Good day.
Here's a new pull request to enable the image saving to gallery on the Android Demo App. I've tried to reduce it as much as possible keeping the original code intact as much as possible. Best regards.

### DIFF
--- a/wrappers/android/app/src/main/java/com/example/zxingcppdemo/MainActivity.kt
+++ b/wrappers/android/app/src/main/java/com/example/zxingcppdemo/MainActivity.kt
@@ -17,6 +17,7 @@
 package com.example.zxingcppdemo
 
 import android.Manifest
+import android.content.ContentValues
 import android.content.Context
 import android.content.pm.PackageManager
 import android.graphics.*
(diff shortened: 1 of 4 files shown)


wrappers/android/app/src/main/java/com/example/zxingcppdemo/MainActivity.kt
```diff
@@ -61,14 +64,37 @@ class MainActivity : AppCompatActivity() {
...
-			//TODO: save image
-
+			takePhoto()
 			// Re-enable camera controls
 			it.isEnabled = true
 		}
 	}
 
+	private fun takePhoto() {
+		val imageCapture = imageCapture ?: return
```
axxel, line 74: The whole PR looks very nice. Thanks a lot! Just one question: what is this line supposed to mean? You introduce a local variable with the same name and same value as the outer scope one? Is this supposed to simply check if `imageCapture` is null?

PR #491, opened 2023-01-16
Updates Debug Image Saving code and enable Torch control

Sergio-:
Removed the code to save High Resolution Camera Images.
Enabled saving of ImageProxy Buffer Bitmap to the Pictures Directory.
Added Function to Turn Camera Torch On and Off.

### DIFF
--- a/core/src/BitMatrix.cpp
+++ b/core/src/BitMatrix.cpp
@@ -6,7 +6,6 @@
 
 #include "BitMatrix.h"
 
-#include "BitArray.h"
 #include "Pattern.h"
 
 #include <algorithm>
--- a/core/src/GenericGFPoly.cpp
+++ b/core/src/GenericGFPoly.cpp
@@ -8,7 +8,6 @@
 #include "GenericGFPoly.h"
 
 #include "GenericGF.h"
-#include "ZXConfig.h"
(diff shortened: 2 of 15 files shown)


axxel (changes requested):
Also the fact that you generate this `imageBitmap` content on every frame, regardless of whether you want to save it later or not, is a waste of computing resources. You can not store a reference to the ImageProxy object because the underlying memory needs to be passed back to the camera subsystem. Therefore my approach would be to simply set a boolean `takePicture` flag inside the `onClickListener` and then inside the analysis thread do the (one) bytes to JPEG conversion and save these bytes to a file.

Also, it might be possible to generate a greyscale JPEG image directly from the yPlane without the copy of all 3 planes. Since we are only looking at the yPlane inside zxing-cpp, this would make sens and could be faster.

Lastly, your PR contains an "empty" merge commit. No idea where that came from. To keep the history somewhat clean, I would squash the commits into one (loosing a little bit of visual attribution in the [network view](https://github.com/zxing-cpp/zxing-cpp/network) ).

If you feel like you have more important stuff to do to advance your project, please just let me know. I can then merge your PR as is and modify the bits that I find questionable later.

PR #492, opened 2023-01-17
Implement Method to Save Buffer Image and Add Torch Control

Sergio-:
Removed function to save High Resolution Camera Image.
Implemented Method to Save Image Buffer as passed to Decoding Algorithm.
Added Button to Turn Camera Torch On and Off.


axxel:
Thanks!

Issue #498, opened 2023-01-29
2.0.0: test suite is failing in `UnitTest` unit


kloczek:
on first look it looks like gcc 13.0.1 related issue (whhich I'm using)
<details>

```console
+ cd zxing-cpp-2.0.0
+ /usr/bin/ctest --test-dir x86_64-redhat-linux-gnu --output-on-failure --force-new-ctest-process -j48 cd wrappers/python
Internal ctest changing into directory: /home/tkloczko/rpmbuild/BUILD/zxing-cpp-2.0.0/x86_64-redhat-linux-gnu
Test project /home/tkloczko/rpmbuild/BUILD/zxing-cpp-2.0.0/x86_64-redhat-linux-gnu
    Start 1: UnitTest
    Start 2: PythonTest
1/2 Test #2: PythonTest .......................   Passed    0.12 sec
2/2 Test #1: UnitTest .........................Subprocess aborted***Exception:   0.34 sec
Running main() from /home/tkloczko/rpmbuild/BUILD/googletest-1.13.0/googletest/src/gtest_main.cc
[==========] Running 262 tests from 56 test suites.
[----------] Global test environment set-up.
[----------] 1 test from PDF417WriterTest
[ RUN      ] PDF417WriterTest.DataMatrixImageWriter
[       OK ] PDF417WriterTest.DataMatrixImageWriter (0 ms)
[----------] 1 test from PDF417WriterTest (0 ms total)

[----------] 1 test from PDF417ScanningDecoderTest
[ RUN      ] PDF417ScanningDecoderTest.BadSymbolLengthDescriptor
[       OK ] PDF417ScanningDecoderTest.BadSymbolLengthDescriptor (0 ms)
[----------] 1 test from PDF417ScanningDecoderTest (0 ms total)

[----------] 8 tests from PDF417HighLevelEncoderTest
[ RUN      ] PDF417HighLevelEncoderTest.EncodeAuto
[       OK ] PDF417HighLevelEncoderTest.EncodeAuto (0 ms)
[ RUN      ] PDF417HighLevelEncoderTest.EncodeAutoWithSpecialChars
[       OK ] PDF417HighLevelEncoderTest.EncodeAutoWithSpecialChars (0 ms)
[ RUN      ] PDF417HighLevelEncoderTest.EncodeIso88591WithSpecialChars
[       OK ] PDF417HighLevelEncoderTest.EncodeIso88591WithSpecialChars (0 ms)
[...]

PR #499, opened 2023-01-29
Fix out-of-bound access in ConcentricFinder.h

duanqn:
The current code accesses pattern[N], which may be caught by strict memory checks.


duanqn:
@axxel Please help review, thanks!

axxel:
Thanks!

axxel:
Minor suggestion for future PRs like this: if you had written "fixes" instead of "resolves" then merging would have automatically closed #498.

PR #500, opened 2023-01-29
Prototype for supporting multiple binarizers

duanqn:
This is a prototype to show the idea of supporting multiple binarizers in ReadBarcodes


axxel:
The code is now outdated and the idea can be easily implemented outside the zxing-cpp API.

PR #503, opened 2023-01-30
Make buffer security check optional

duanqn:
Adds an option to enable people building ZXing-cpp with buffer security check.


axxel:
I did not see a need to make this configurable. The default is clearly `/GS+` and if someone wants to change that, they can fiddle with `CMAKE_CXX_FLAGS` themselfs.

PR #508, opened 2023-02-02
Reduce min iOS deployment target to 11.0

kientux:
Min iOS deployment target supported by latest Xcode (14.0) is still 11.0 so I think it's reasonable to reduce iOS wrapper's deployment target to 11.0. I built and tested it, still works fine.


axxel:
Thanks! I trust that this will not have any detrimental effect on others ;).

Issue #512, opened 2023-02-13
Python: typo refers to "Mirco" QR codes


cbm755:
I noticed the Python library has `BarcodeFormat.MircoQRCode`.    Note "MIR" "CO"...  Not "Micro"

Seems like a typo to me?

```
grep * -rie mirco
core/src/qrcode/QRFormatInformation.cpp:	// We don't use the additional masking (with 0x4445) to work around potentially non complying MircoQRCode encoders
wrappers/python/zxing.cpp:		.value("MircoQRCode", BarcodeFormat::MicroQRCode)
```

I tracked this to: commit d2031396c, where it certainly looks like a typo...  I'll send a PR, but not sure what to deal with the deprecation.

cbm755:
> not sure what to deal with the deprecation.

What about this?  Sadly `BarcodeFormat.MircoQRCode` was exposed and available during 2.0.0 (I'm using that symbol downstream for example).

When 2.0.1 comes out I can just pin on `>= 2.0.1`, so its no problem for me **personally**, just pointing it out, in case you strictly follow https://semver.org for example.

axxel:
You surely have a point but I could not find information on how to generate a depreciation warning for the use of a pybind11 enum field. I do (try to) follow semver but only on the c++ API so far. I could be convinced of adding the old value back until 3.0 comes out at which point it will simply disappear without having been warned before.

cbm755:
No need to do so on my account.

In case it helps someone else, I'm doing this:
```python
    try:
        micro = BarcodeFormat.MicroQRCode
    except AttributeError:
        # workaround github.com/zxing-cpp/zxing-cpp/issues/512
        micro = BarcodeFormat.MircoQRCode
```

PR #513, opened 2023-02-13
Correct "Mirco" to "Micro"

cbm755:
Closes Issue #512.


axxel:
Thanks. That was indeed an embarrassing typo.

Issue #488, opened 2023-01-16
2.0.0 does not provide wheels for Python 3.8


cbm755:
I'm building on Ubuntu 20.04 which still has Python 3.8, and I notice it builds from source.  

Is that intentional?  See https://pypi.org/project/zxing-cpp/2.0.0/#files where only Python 3.9 -- 3.11 are shown.

Relatedly, what is the minimum supported Python version?

axxel:
> Is that intentional?

Yes. At some point in the past I made the (more or less informed decision) that shipping wheels for the most recent 3 python versions would be the way to go. I see how that is annoying if you are an Ubuntu LTS user.

> Relatedly, what is the minimum supported Python version?

That is a good question. It is basically dependent on what pybind11 thinks about this matter. I would imagine that 3.0 still works but I have no clue. That said, `setup.py` actually `python_requires=">=3.6"`. But I don't remember where that number comes from.

cbm755:
thanks for confirming that 3.8 is in principle supported.

I guess nothing else to do here, so closing.

PR #514, opened 2023-02-13
python: explicitly list supported python for PyPI

cbm755:
See Issue #488.  The minimum listed version is 3.6, which matches the `python_requires` variable.

- - - -

This information adds information in pypi.org, for example here's what NumPy looks like:

image

compared to `zxing-cpp` which looks like:

image

### DIFF
--- a/wrappers/python/setup.py
+++ b/wrappers/python/setup.py
@@ -71,6 +71,12 @@ setup(
     classifiers=[
         "Development Status :: 4 - Beta",
         "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
         "License :: OSI Approved :: Apache Software License",
         "Operating System :: OS Independent",
         "Topic :: Multimedia :: Graphics",


axxel:
May I ask what the purpose of this list is?

cbm755:
I think it is intended so that users can look at PyPI and see what versions are supported...

I think---but have not verified---that `python_requires=">=3.6"` does not appear to users.  If you look in #488, I asked about minimum supported Python, as I did not find the answer myself: this would ensure I would've.

cbm755:
Hmmm, I just checked my own project and it seems I'm wrong about `python_requires`.  It **does** appear, above in "meta".

I guess I'm not sure why: "everyone else is doing it?".

(Given the info is elsewhere, go ahead and close this if you prefer)

cbm755:
image

cbm755:
> You should always include at least which version(s) of Python your package works on, which license your package is available under, and which operating systems your package will work on. For a complete list of classifiers, see https://pypi.org/classifiers/.

https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata

I'm not sure if that means Python 2 versus 3 (i.e., an historical thing) or if they are recommending what I've done in this PR.

axxel:
Looking at https://stackoverflow.com/questions/45778790/specifying-the-minimum-python-version-required-for-a-module-in-pip I believe `python_requires` is the correct place to specify this functional limitation. And since your original issue (not visible on pypi.org) was an oversight. I'd like to keep it as it is.

PR #528, opened 2023-03-13
added qt include file

lightX2:


axxel:
What are you trying to achieve with this? Please note that this project uses cmake as the build tool and I don't currently intend to add another one, especially not a rather 'niche' one like qmake.

lightX2:
qt apps often uses qmake builder, so that pri file make it easy to include this lib to project

axxel:
I understand but I don't want to start maintaining more than one build system. You can certainly just use cmake to build (and install) this library and then use it from your qmake project like any other (system) library. And if you have the need to actually include all the sources into your own project and compile them there, then you'll have to put this `pri` file into your project, I'm afraid.

PR #530, opened 2023-03-17
Refactor Python wrapper so the sdist includes the core library code

stumpylog:
This PR allows the sdist to include the core library code in the archive, instead of downloading the code at build time.  This _should_ be enough for Piwheels to build armv7 wheels, though I'm not positive, mostly due to the existence of `zxing_add_package_stb`.

The change uses symlinks to create the `core` directory and required `zxing.cmake` file in the wrapper folder, without copying the files from the parent at archive creation time.  The MANIFEST.in then includes these files (and follows the symlinks properly) to create a self contained archive.

I've tested all the following:
- Inside the directory `wrappers/python`
  - `pip wheel .`
  - `python3 -m build --wheel`
  - `python3 -m build --sdist`
- build-python-dist: https://github.com/stumpylog/zxing-cpp/actions/runs/4448756738
- Main CI: https://github.com/stumpylog/zxing-cpp/actions/runs/4448730130
- Building the sdist in an isolated directory and virtualenv


wrappers/python/CMakeLists.txt
```diff
@@ -14,50 +14,23 @@ if (NOT hasParent)
...
-            add_subdirectory(${zxing-cpp_SOURCE_DIR}/core ZXing EXCLUDE_FROM_ALL)
-        endif()
+        message(FATAL_ERROR "Unable to locate zxing source code")
     endif()
 else()
     zxing_add_package(pybind11 pybind11 ${pybind11_git_repo} ${pybind11_git_rev})
 endif()
 
+# Check pybind helpers are availible
+find_package(pybind11 CONFIG)
```
axxel, line 28: `find_package` is automatically called inside `zxing_add_package` if `BUILD_DEPENDENCIES` is set to `AUTO` which happens in the toplevel `CMakeLists.txt` which is only available/evaluated if `hasParent` is true. So in the `sdist` case, the following line should do the trick if added near line 16 above: `set(BUILD_DEPENDENCIES "AUTO")`.

With that in-place one single `zxing_add_package` call and no extra `find_package` should hopefully work for all use cases. Could you please check if I'm correct with this assessment?

wrappers/python/pyproject.toml
```diff
@@ -4,5 +4,6 @@ requires = [
     "setuptools_scm",
     "wheel",
     "cmake>=3.14",
+    "pybind11[global]>=2.10.4",
```
axxel, line 7: * This line would force the presence of the pybind11 module (to be later found via `find_package`) right?
* What does the `[global]` do?

axxel (changes requested):
Thanks for your effort to help with this, much appreciated.

wrappers/python/pyproject.toml
stumpylog: - I believe it was the Windows build which couldn't find pybind11, even with this change and spawned the extra find package above
- [Docs here](https://pybind11.readthedocs.io/en/stable/installing.html#include-with-pypi). It's use here in the `pyproject.toml` build system is how it's meant to be used.  Without it, all the systems failed to locate pybind11 when build (not just windows)

wrappers/python/CMakeLists.txt
stumpylog: I'll have to have a look.  I recall this was a workaround for a Windows build failure, when it couldn't locate pybind.  Will check if your suggestion helps

stumpylog: Worked great with that change across all the platforms.  There's a single `zxing_add_package ` now for finding pybind11

PR #543, opened 2023-03-29
rare crash fix

lightX2:


axxel:
Thanks for the PR. My understanding of the code is that the check you introduce here will always be evaluated as false. So before I merge that, I'd like understand how it can possibly be true. Can you provide an input image that triggers the crash you are fixing here?

lightX2:
image

axxel:
I don't see how the assertion failure you posted would be affected by the proposed change. This makes it even more important for me to get access to the input image data that triggered this. If you replace the `assert` line from your screenshot with
```
			if( !(line.points().empty() || p != line.points().back()) )
				throw 1;
```
And then replace your `ReadBardcode(...)` call with something along the lines of
```
	try {
		ReadBarcode(...);
	} catch ( int ) {
		// save the ImageView that was passed to a PNG file
	}
```

Then I can start investigating what is happening.

lightX2:
send your email

axxel:
I have not gotten any e-mail from you (I believe). Did you send me something?

axxel:
Closing this now for the lack of feedback from the OP.

jhay06:
this still recurring

axxel:
I still need a sample image that allows me to reproduce the problem.

lightX2:
i think something with not single image, but series.
later image processing is OK.
But we don't see problem for now in app

PR #553, opened 2023-04-22
Start implementing C wrapper

siiky:
This is a sketch implementation to serve as discussion point for https://github.com/zxing-cpp/zxing-cpp/discussions/551. 

An alternative fork can be found [here](https://git.sr.ht/~siiky/zxing-cpp), and an example program using this WIP wrapper [here](https://git.sr.ht/~siiky/zxing-c-test).

Currently, the program sometimes crashes (with a segfault or a floating point exception) and sometimes "succeeds" ("Result is invalid" is printed to `stderr`). I haven't discovered yet why that is because I can't make it crash inside `gdb` :/

For now I started with the reading API only, because that's what I had  to use and know now how to use. It should be enough, at least for a while, to discuss what needs to be discussed.

---

As is custom in C, there's a library prefix used to try to avoid identifier collisions -- I chose `zxing`.

Identifiers follow an easy rule (I think): `ZXing` is lower-cased, and the namespace separator (`::`) is replaced with an underscore (`_`):

 * CamelCase is maintained everywhere;
 * Type names are prefixed with the library prefix (e.g. `zxing_ImageFormat` for `ZXing::ImageFormat`);
 * Enum variants are prefixed with the library prefix as well as the enum name (e.g. `zxing_ImageFormat_RGB` for `ZXing::ImageFormat::RGB`);
 * Instance methods are prefixed with the library prefix as well as the type name (e.g. `zxing_ImageView_format()` for `ZXing::ImageView.format()`);
 * Other functions are only prefixed with the library prefix (e.g. `zxing_ReadBarcode()` for `ZXing::ReadBarcode()`).

I chose this naming scheme simply to make it easy to switch or find the corresponding identifiers from one to the other: someone using the C wrapper should be able to easily find the corresponding C++ identifier and vice-versa.

---

Some things to improve/consider/discuss:

 - [x] Using `void *`
 - [x] Better enum interoperability
 - [ ] Let C users stack allocate if possible?


wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,202 @@
+#include "zxing.h"
+
+#include <ZXing/ImageView.h>
+#include <ZXing/BarcodeFormat.h>
+#include <ZXing/DecodeHints.h>
+#include <ZXing/Result.h>
+#include <ZXing/ReadBarcode.h>
+
+#ifdef __cplusplus
```
axxel, line 9: The #ifdef here is unnecessary as this file is only every compiled by a c++ compiler.

wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,202 @@
...
+ * ...
+ */
+
+/*
+ * ZXing/ReadBarcode.h
+ */
+
+zxing_Result zxing_ReadBarcode (const zxing_ImageView iv, const zxing_DecodeHints hints)
+{
+	const ZXing::ImageView & cppiv = (const ZXing::ImageView &) iv;
```
axxel, line 173: `zxing_ImageView` is a pointer, not a reference. This might have something todo with your crashes?

wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,202 @@
...
+	// TODO
+	//zxing.cpp: In function ‘void* zxing_ReadBarcode(zxing_ImageView, zxing_DecodeHints)’:
+	//zxing.cpp:157:76: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
+	//  157 |         const ZXing::DecodeHints & cpphints = (const ZXing::DecodeHints &) hints;
+	const ZXing::DecodeHints & cpphints = (const ZXing::DecodeHints &) hints;
+	ZXing::Result cppresult = ZXing::ReadBarcode(cppiv, cpphints);
+    ZXing::Result * cppresultptr = new ZXing::Result(cppresult);
```
axxel, line 180: `new ZXing::Result(std::move(cppresult));` should be more performant and save a bunch of string copies.

wrappers/c/zxing.h
```diff
@@ -0,0 +1,143 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * ZXing/ImageView.h
+ */
+
+typedef void * zxing_ImageView;
```
axxel, line 22: using `void *` is a bad idea as it completely losses type safety on the C side. Please have a look at the following sites for inspiration on how this might be handled differently: https://github.com/enricgalceran/cpplib_wrapper and https://www.teddy.ch/c++_library_in_c/

wrappers/c/zxing.h
```diff
@@ -0,0 +1,143 @@
...
+	zxing_ImageFormat_BGR  = 0x03020100,
+	zxing_ImageFormat_RGBX = 0x04000102,
+	zxing_ImageFormat_XRGB = 0x04010203,
+	zxing_ImageFormat_BGRX = 0x04020100,
+	zxing_ImageFormat_XBGR = 0x04030201,
+} zxing_ImageFormat;
+
+zxing_ImageView zxing_ImageView_new (const uint8_t * data, int width, int height, zxing_ImageFormat format, int rowStride, int pixStride);
+void zxing_ImageView_free (zxing_ImageView iv);
+
```
axxel, line 37: `new` and `free` is a wired mix of c++ and c terminology. I would suggest to use ether `new` and `delete` or `create` and `destroy`.

axxel (changes requested):
Thanks for starting to work on this. I hope my comments make sens to you.

wrappers/c/zxing.cpp
siiky: Don't know. However, initially I had this:

```cpp
zxing_Result zxing_ReadBarcode (const zxing_ImageView iv, const zxing_DecodeHints hints)
{
	const ZXing::ImageView * cppiv = (const ZXing::ImageView *) iv;
	const ZXing::DecodeHints * cpphints = (const ZXing::DecodeHints *) hints;
	ZXing::Result cppresult = ZXing::ReadBarcode(cppiv, cpphints);
	// ...
```

With that version the C++ compiler complains:

```
zxing.cpp: In function ‘ZXing::Result* zxing_ReadBarcode(zxing_ImageView, zxing_DecodeHints)’:
zxing.cpp:177:54: error: invalid initialization of reference of type ‘const ZXing::ImageView&’ from expression of type ‘const ZXing::ImageView*’
  177 |         ZXing::Result cppresult = ZXing::ReadBarcode(cppiv, cpphints);
      |                                                      ^~~~~
In file included from zxing.cpp:7:
/home/siiky/.local/include/ZXing/ReadBarcode.h:21:37: note: in passing argument 1 of ‘ZXing::Result ZXing::ReadBarcode(const ImageView&, const DecodeHints&)’
   21 | Result ReadBarcode(const ImageView& buffer, const DecodeHints& hints = {});
      |                    ~~~~~~~~~~~~~~~~~^~~~~~
```

wrappers/c/zxing.h
siiky: Right, using `void*` is one of the things I wanted to avoid, but I didn't know how... It didn't occur to me to use `typedef struct foo * foo;` at the time, nor to conditionally `typedef` to "the right thing". Thanks for the references.

wrappers/c/zxing.cpp
axxel: You need to dereference the two pointers: `ZXing::ReadBarcode(*cppiv, *cpphints)`

wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+extern "C" {
+
+/*
+ * ZXing/ImageView.h
+ */
+
+zxing_ImageView zxing_ImageView_new (const uint8_t * data, int width, int height, zxing_ImageFormat format, int rowStride, int pixStride)
+{
+	ZXing::ImageFormat cppformat = (ZXing::ImageFormat) format;
```
axxel, line 17: Please use a `reinterpret_cast<ZXing::ImageFormat>(format)` here. Same for all the other c-style casts below.

wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+/*
+ * ZXing/ReadBarcode.h
+ */
+
+zxing_Result zxing_ReadBarcode (const zxing_ImageView iv, const zxing_DecodeHints hints)
+{
+	const ZXing::ImageView & cppiv = (const ZXing::ImageView &) iv;
+	const ZXing::DecodeHints & cpphints = (const ZXing::DecodeHints &) hints;
+	return new ZXing::Result(std::move(ZXing::ReadBarcode(cppiv, cpphints)));
```
axxel, line 149: If you pass a temporary to the `Result` constructor, you don't need the `std::move()` ;).

wrappers/c/zxing.cpp
siiky: Oh! That was it, yeah, thanks.

wrappers/c/zxing.h
```diff
@@ -0,0 +1,153 @@
...
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+
+#include <ZXing/ImageView.h>
+#include <ZXing/DecodeHints.h>
+#include <ZXing/Result.h>
+
+typedef ZXing::ImageView * zxing_ImageView;
```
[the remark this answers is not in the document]
siiky: Fair. I hid the pointer initially just out of convenience and to better mimic the C++ equivalent.

wrappers/c/zxing.cpp
siiky: I think it's time now to discuss enums.

Using `reinterpret_cast` in `zxing_ReadBarcode()`/`zxing_ReadBarcodes()` works fine, but for the enums the compiler complains with errors such as these:

```
zxing.cpp: In function ‘zxing_ImageView* zxing_ImageView_new(const uint8_t*, int, int, zxing_ImageFormat, int, int)’:
zxing.cpp:17:40: error: invalid cast from type ‘zxing_ImageFormat’ to type ‘ZXing::ImageFormat’
   17 |         ZXing::ImageFormat cppformat = reinterpret_cast<ZXing::ImageFormat>(format);
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
zxing.cpp: In function ‘zxing_ImageFormat zxing_ImageView_format(zxing_ImageView*)’:
zxing.cpp:48:16: error: invalid cast from type ‘ZXing::ImageFormat’ to type ‘zxing_ImageFormat’
   48 |         return reinterpret_cast<zxing_ImageFormat>(iv->format());
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

---

According to [this](https://stackoverflow.com/questions/20792121) question and answers on SO it's as I suspected: there's no way to "export" C++ enums as C enums.

The alternatives are basically (a) to use plain C enums alone, or (b) to define both C and C++ enums and use a (runtime) translation between the two.

(a) may be the easier alternative but is not backwards compatible. Plus C++ features (such as namespaces and "object-like" enums) can't be used (I think?).

(b) The latter is backwards compatible and requires no changes on the C++ library, but obviously incurs a performance penalty (though C->C++ translation can be minimised to an array indexing operation).

The accepted answer suggests a sort of mix between the two, where both C and C++ enums are defined, with the same values, but the C enum has a typical C prefix, while the C++ enum has no prefix (because of namespaces). The real definition of the enums can be saved in a file of its own so that it can be included either by the C or the C++ headers.

wrappers/c/zxing.cpp
siiky: This may have really been the cause of the segfault. Though I still can't scan a simple QR code, the example program doesn't segfault anymore.

wrappers/c/zxing.h
```diff
@@ -0,0 +1,153 @@
...
+	zxing_ImageFormat_RGBX = 0x04000102,
+	zxing_ImageFormat_XRGB = 0x04010203,
+	zxing_ImageFormat_BGRX = 0x04020100,
+	zxing_ImageFormat_XBGR = 0x04030201,
+} zxing_ImageFormat;
+
+zxing_ImageView * zxing_ImageView_new (const uint8_t * data, int width, int height, zxing_ImageFormat format, int rowStride, int pixStride);
+void zxing_ImageView_delete (zxing_ImageView * iv);
+
+int zxing_ImageView_width (zxing_ImageView * iv);
```
axxel, line 53: please add a `const` qualifier for the getter functions, like `int zxing_ImageView_width (const zxing_ImageView * iv);`

wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,161 @@
...
+	ZXing::Result result = ZXing::ReadBarcode(*cppiv, *cpphints);
+	return new ZXing::Result(result);
+}
+
+zxing_Results * zxing_ReadBarcodes (const zxing_ImageView * iv, const zxing_DecodeHints * hints)
+{
+	const ZXing::ImageView * cppiv = reinterpret_cast<const ZXing::ImageView *>(iv);
+	const ZXing::DecodeHints * cpphints = reinterpret_cast<const ZXing::DecodeHints *>(hints);
+	ZXing::Results results = ZXing::ReadBarcodes(*cppiv, *cpphints);
+	return new ZXing::Results(results);
```
axxel, line 158: either use std::move or a pass a temporary (which is implicitly an r-value reference). I suggest:
```
return new ZXing::Results(ZXing::ReadBarcodes(*cppiv, *cpphints));
```

wrappers/c/zxing.h
siiky: Yup, I thought about that the other day but then forgot.

wrappers/c/zxing.cpp
siiky: Oh I misunderstood your other comment. Fixed it now, thanks.

wrappers/c/zxing.cpp
siiky: The compiler didn't complain about `static_cast`.

Are C++ enum variants guaranteed to be unique sequential integers (unless otherwise explicitly specified)?

```cpp
enum foo {
  variant0, // 0
  variant1, // 1
  variant2, // 2
  // ...
}
```

axxel: Yes they are. There might be differences in the width of the underlying integer type. But the static_cast should take care of that.

wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include <ZXing/Result.h>
+#include <ZXing/ReadBarcode.h>
+
+extern "C" {
+
+/*
+ * ZXing/ImageView.h
+ */
+
+zxing_ImageView * zxing_ImageView_new (const uint8_t * data, int width, int height, zxing_ImageFormat format, int rowStride, int pixStride)
```
axxel, line 15: To fit the style of the rest of the project, would you please remove the space between the typename and the '*'? Or maybe simply let clang-format with the config in the top-level directory take care of that`

wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+/*
+ * ZXing/ReadBarcode.h
+ */
+
+zxing_Result * zxing_ReadBarcode (const zxing_ImageView * iv, const zxing_DecodeHints * hints)
+{
+	const ZXing::ImageView * cppiv = reinterpret_cast<const ZXing::ImageView *>(iv);
+	const ZXing::DecodeHints * cpphints = reinterpret_cast<const ZXing::DecodeHints *>(hints);
+	return new ZXing::Result(ZXing::ReadBarcode(*cppiv, *cpphints));
```
axxel, line 149: I just realized that you don't need these `reinterpret_cast`s here at all, since the typedefs in `zxing.h` take care of that. Simply call `ZXing::ReadBarcode(*iv, *hints)` directly.

wrappers/c/zxing.h
```diff
@@ -0,0 +1,153 @@
+#ifndef _ZXING_C_H
+#define _ZXING_C_H
```
axxel, line 2: I was thinking about the name of the header and thought that `zxing.h` is a bit too generic for my taste in this context. Mabe `zxingc.h` or `zxing-c.h`. I can't quite decide what I prefer. I wanted to use `zxing-cpp` in the context of the python and android module but there a '-' is not allowed or problematic, so I ended up with `zxingcpp`. Do you have a convincing argument for either of them?

axxel (changes requested):
Could you please add your sample code from the other repo as well? Do you think you could also add a `CMakeLists.txt`? If not, I'll try to add one to this PR myself.

wrappers/c/zxing.h
siiky: Basically indifferent to me. Between `zxingc` and `zxing-c` I prefer `zxingc`, mainly because it's shorter, but also because of what you said (`zxingcpp` being used somewhere else) and in case `-` is disallowed for whatever reason.

wrappers/c/zxing.cpp
siiky: Ok nice, this is solved then.

siiky:
Sure, I can add the example program. However, it's still not working, I need to get some time to try to debug it. BTW is it worth adding NULL checks to input arguments?

RE `CMakeLists.txt` I can try too, but I'm not familiar with CMake so it'd probably be easier and faster if you added one yourself ^^'

wrappers/c/zxing.cpp
```diff
@@ -0,0 +1,155 @@
+#include "zxing.h"
+
+#include <ZXing/BarcodeFormat.h>
+#include <ZXing/DecodeHints.h>
+#include <ZXing/ImageView.h>
+#include <ZXing/ReadBarcode.h>
+#include <ZXing/Result.h>
+
+extern "C"
```
axxel, line 9: I updated the .clang-format file to not indent `extern "C"` blocks. Please reapply.

wrappers/c/zxing.h
axxel: I checked what other libraries do. '-' is totally fine in a header file name. -> Please rename this file to `zxing-c.h` and the cpp file to `zxing-c.cpp`. This also fits nicely with your `zxing-c-test.c` choice. ;).

wrappers/c/zxing.h
```diff
@@ -0,0 +1,164 @@
...
+
+/*
+ * <stdbool.h>
+ *  bool
+ *
+ * <stdint.h>
+ *  uint8_t
+ */
```
axxel, line 10: Please remove this comment block and instead add one with copyright and license info like all other source files have. This also applies to the other 2 new files from you.

axxel (changes requested):
Thanks! I think I'm out of ideas on how to further improve this, except simply adding the missing functions.

axxel:
> Sure, I can add the example program. ~However, it's still not working, I need to get some time to try to debug it.~ NVM that, I screwed converting a PNG to JPEG and didn't notice it... The program now reports the result as "valid". Still have to write a function to get the data out, but I can add the program here.

In that regard: note that I added a commit changing the stbi usage to be in line with the other tools. This would then also work for png images.

> RE `CMakeLists.txt` I can try too, but I'm not familiar with CMake so it'd probably be easier and faster if you added one yourself ^^'

The following should work as a minimal starting point (not tested):

```
zxing_add_package_stb()

if (BUILD_READERS)
    add_executable (zxing-c-test zxing-c.cpp zxing-c-test.c)
    target_link_libraries (zxing-c-test ZXing::ZXing stb::stb)
endif()
```

> BTW is it worth adding NULL checks to input arguments?

No.

wrappers/c/zxing.h
siiky: As far as I'm concerned, the code I'm contributing is public domain, you can do with it whatever you want (hopefully good :)

Do you want me to add a copyright notice with your name?

```
/*
* Copyright 2023 Axel Waggershauser
*/
```

axxel: You are the author, so your name is the correct one (in my opinion ;)).

siiky:
> This would then also work for png images.

Yeah, I know STBI works for PNGs &c, I was just trying different options and that's what I committed lol

> you also need to add the line `add_subdirectory (wrappers/c)` at the end of the top-level `CMakeLists.txt`.

Added also a `BUILD_C_WRAPPER` option. (And I'm not really sure how to try the `CMakeLists.txt`, I'm generally content with `Makefile`s ^^)

axxel:
Did I put you off somehow, did I miss an open question your are waiting for the answer or are you simply on vacation? :)

siiky:
Nah, don't worry, I won't run away :) I just have a ton of things atm and this isn't a priority right now.

siiky:
@axxel Sorry I left this hanging all of a sudden. I'm hoping I'll be able to relax a bit around mid-June and have time to work on this again.

siiky:
Hey @axxel, I restarted work on this just now, hoping it would be trivial to add support for the `bytes()` field. The code in this latest commit compiles alright, but the test program is now complaining the result is invalid (it would say valid before). I'm not sure what happened, but will need some time to figure out what's wrong.

In the meantime, could you help with adding the right CMake files to the right places? And then explain how to use them to build the example program? :D

axxel:
I don't see how your latest commit can have any effect on your test program. I'm pretty sure it is something else.

wrappers/c/zxing.h
axxel: I added the copyright comment and used 'siiky', maybe you want to replace that with your real name.

axxel:
I went ahead and added a few more properties of DecodeHints and Result. Also added a ctest case. The test program actually outputs some content now. What else would be required for your original use case?

axxel (approved):

wrappers/c/zxing-c.cpp
```diff
@@ -0,0 +1,244 @@
...
+	{
+		if (!str)
+			return {};
+		try {
+			auto format = BarcodeFormatsFromString(str);
+			return static_cast<zxing_BarcodeFormats>(*reinterpret_cast<BarcodeFormat*>(&format));
+		} catch (...) {
+			return zxing_BarcodeFormat_Invalid;
+		}
+	}
```
siiky, line 84: I saw you replaced `zxing_BarcodeFormatFromString` with `zxing_BarcodeFormatsFromString`. Initially I implemented the singular because I didn't know how to implement the plural (it looked like `zxing_BarcodeFormats` was slightly more complicated than just a `zxing_BarcodeFormat` typedef), but also because `zxing_BarcodeFormatFromString` is more useful to us to prevent (programmer) mistakes -- we only need to scan one format at a time.

Is it ok to have both, or are you against it?

siiky:
Hey, thanks for your work! Sorry I was unavailable once again these few days, I'm going slightly insane with some unexpected events lately...

For our use case at work we only need `zxing_ReadBarcode` and `zxing_Result_bytes`/`zxing_Result_text`, so this wrapper is already feature-complete for us (and looking much better than my original hack).

As it is, is it ready to be merged or is there anything missing for you?

wrappers/c/zxing-c.cpp
axxel: I'm fine with it and just added it...

axxel:
I just added `zxing_Results_*` support which I thought was still a major lack of functionality. Please note that I decided to model "not found" by returning a `NULL` for both single and multi-symbol variants.

I'll be happy to merge it as is and add missing stuff from there when someone speaks up. I intent to eventually simply add the `zxing-c.*` to the core library to be automatically available for all libZXing users. But before I do that, it would be nice to get some feedback about the usability of the API.

siiky:
> note that I decided to model "not found" by returning a NULL for both single and multi-symbol variants

Yup, seems to be the most natural.

> I'll be happy to merge it as is and add missing stuff from there when someone speaks up. I intent to eventually simply add the zxing-c.* to the core library to be automatically available for all libZXing users. But before I do that, it would be nice to get some feedback about the usability of the API.

Makes sense to me. I won't be able to provide any opinion on advanced use-cases, but once I convert our program to use the new C API, I'll report any issues I find.

PR #555, opened 2023-04-23
Update Android wrapper and dependencies

Rupan:
When upgrading to Android Studio Flamingo (2022.2.1), projects are encouraged to upgrade to Gradle 8 (and thus JDK 17).  There is also a new Kotlin release available as well as a build tools update.

The `YUV_422_888` and `YUV_444_888` images formats used in `BarcodeReader` are not supported on Android until SDK version 23.  Simply declaring that the package supports SDK 21 doesn't change that, so update `minSdkVersion` to 23 to prevent undefined behavior.

Comment out the ABI filters to enable ChromeOS support out of the box.  Faster builds are great, but the build time isn't significantly improved (maybe 30s?).  It can always be re-enabled if it significantly hampers productivity.

 Also bump a few dependencies to more recent versions.


wrappers/android/.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,8 @@
+<component name="InspectionProjectProfileManager">
```
axxel, line 1: Please remove this file from the PR.

wrappers/android/.idea/jarRepositories.xml
```diff
@@ -26,5 +26,15 @@
       <option name="name" value="MavenRepo" />
       <option name="url" value="https://repo.maven.apache.org/maven2/" />
     </remote-repository>
+    <remote-repository>
```
axxel, line 29: what is the reason to add these additional repositories?

axxel (changes requested):
Thanks. Looking forward to merging this.

wrappers/android/.idea/jarRepositories.xml
Rupan: This was automatically added by either the IDE update or the switch to Gradle 8.

To be honest I was surprised that the idea project files were committed to begin with.  It would probably be better to remove them and add the path to gitignore.

Rupan:
All requested changes have been made.  I wonder whether deleting the whole `.idea` directory (which is regenerated by the IDE at project open) would be a better solution in this case.

wrappers/android/.idea/jarRepositories.xml
axxel: That is surprising because I did both here as well to better understand why you incorporated what. And this file has not been changed here.

Rupan: I didn't intentionally add this in, it just appeared at some point.  It isn't required for the build to work in any case.  Anyway since it is a concern I've rolled the change back in de4d6527ff04fc551838d7fca165ec9599658814.  Ideally this whole directory would be deleted to avoid such discussions in the future, its all autogenerated anyway.

axxel: Thanks. The file still shows up in the PR, though. There is some whitespace change. Would you remove that completely, please?

Meanwhile, I'll investigate what the best cause of action regarding the whole .idea directory is.

Rupan:
`jarRepositories.xml` has been fully reverted.

wrappers/android/.idea/jarRepositories.xml
Rupan: Fully reverted now.

axxel:
@markusfisch you proposed the last downgrade of the minSDKVersion [here](68b417bca2915b476ad5995b8806bb26bcb57659). Do you have an opinion about the proposed change here?

@Rupan the CI setup apparently does not have a JDK 17 available: https://github.com/zxing-cpp/zxing-cpp/actions/runs/4780113432/jobs/8510757868?pr=555#step:3:26

Rupan:
There appear to be a couple of ways around the noted issues.  The project has (historically) been built successfully with an older JDK, so it might be acceptable to force both the Java and Kotlin versions to 11 or even 8 in `build.gradle`, like this:

`
	compileOptions {
		sourceCompatibility JavaVersion.VERSION_11
		targetCompatibility JavaVersion.VERSION_11
	}

	kotlinOptions {
		jvmTarget = '11'
	}
`

Regarding the change to SDK 23, I'd suggest keeping it.  Allowing lower versions would impact the library functionality.  As an alternative, keeping the exsting level but conditionally initializing the variable like this could work:

`
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
            supportedYUVFormats = arrayOf(ImageFormat.YUV_420_888, ImageFormat.YUV_422_888, ImageFormat.YUV_444_888)
        } else {
            supportedYUVFormats = arrayOf(ImageFormat.YUV_420_888)
        }
`

markusfisch:
Well, according to the latest stats from Google there's a gap of ~2% between API 21 (what covers 99,3% of all devices connecting to Google Play) and API 23 (97,2%). Now 2% doesn't sound like much, and one could argue that API 23 has many security improvements over API 21 and that these 2% are better off getting a newer device anyway.

On the other hand, it would easily be possible to use a runtime check before using the symbols from API 23, `YUV_422_888` and `YUV_444_888`, like this:

```kotlin
diff --git a/wrappers/android/zxingcpp/src/main/java/com/zxingcpp/BarcodeReader.kt b/wrappers/android/zxingcpp/src/main/java/com/zxingcpp/BarcodeReader.kt
index b241cef0..73518e25 100644
--- a/wrappers/android/zxingcpp/src/main/java/com/zxingcpp/BarcodeReader.kt
+++ b/wrappers/android/zxingcpp/src/main/java/com/zxingcpp/BarcodeReader.kt
@@ -20,6 +20,7 @@ import android.graphics.Bitmap
 import android.graphics.ImageFormat
 import android.graphics.Point
 import android.graphics.Rect
+import android.os.Build
 import androidx.camera.core.ImageProxy
 import java.lang.RuntimeException
 import java.nio.ByteBuffer
@@ -67,7 +68,11 @@ class BarcodeReader {
     var options : Options = Options()
 
     fun read(image: ImageProxy): Result? {
-        val supportedYUVFormats = arrayOf(ImageFormat.YUV_420_888, ImageFormat.YUV_422_888, ImageFormat.YUV_444_888)
+        val supportedYUVFormats = mutableSetOf(ImageFormat.YUV_420_888).apply {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                addAll(setOf(ImageFormat.YUV_422_888, ImageFormat.YUV_444_888))
+            }
+        }.toTypedArray()
         if (image.format !in supportedYUVFormats) {
             error("invalid image format")
         }
```

So it's a matter of taste I guess 😉 I would choose the runtime check over increasing the minSDK because I really care about every possible user and don't like leaving someone behind if I can easily avoid that.

Of course, this means a bit more code, and at some point these 2% will shrink to nothing, and then we will want to remove the runtime check and have a little work to do. 

Since the library also depends on CameraX, although this is not absolutely necessary, it is perhaps more consistent to simply increase the minSDK now I think.

Rupan:
@axxel you're the maintainer, so let me know what your preference is (to both topics at hand) and I'll update the PR accordingly.

axxel:
Lets simply increase the min SDK version and "make it compile" on the CI system. If the simplest option is to use the `compileOptions` you mentioned, that is fine by me. I would go for version 11, though, as this is what worked before. I would also suggest to add a comment to those additional lines, to make is clear what their purpose is and we have a hint as to when we can remove them again. Thanks.

Rupan:
It looks like you've already got Java 8 specified in the compile options.  I tried changing that to 11 in the topes it would help, but likely not.  The CI probably needs an updated version of Java.

axxel:
As you expected this did not help. I see the following options moving forward:
1) find a solution for the CI system
2) let this PR rest until ubuntu-latest includes java 17 (maybe the just released 23.04 does that?)
3) revert the gradle 8 update

axxel:
This is (mostly) obsolete now.

PR #567, opened 2023-05-22
wasm wrapper: add function to reader to scan multiple barcode

Trim:
I wrote a new patch as discussed in #565 to update the wasm wrapper to allow to scan multiple barcodes with only one call.

This patch adds two new functions `readBarcodesFromPixmap` and `readBarcodesFromPixmap`. They return a new structure `ReadResults`. All found barcode are stored in the `vector<ReadResult> results` property (their `error` property is always empty). If an error occurs the `results` vector is empty and the `std::string error` property of `ReadResults` is filled with the exception message.

This patch is compatible with the two functions `readBarcodeFromPixmap` and `readBarcodeFromPixmap` which returned only one `ReadResult` with error inside it directly.

In #565 we discussed to add a default value for `maxSymbols`: I did not add it because it wasn't usable from the JavaScript side (an error was raised by the browser saying I had to give 5 arguments and only 4 were found).

About the JavaScript way to use the result: the `register_vector` emscripten function is creating a new interface and some JavaScript boilerplate code should be added to use the result.

See the binding.h emscripten file to view all methods available to interact with a vector: https://github.com/emscripten-core/emscripten/blob/98b618f09bc106d401c0045fa7252f1cd15bd3c0/system/include/emscripten/bind.h#L1894

User can either create C-like for loops with the `.size()` and `.get(int i)` methods or create a JavaScript [Iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator) to wrap the received result.

I've updated the demo_reader to show how to create and use the Iterator way.

The Iterator can be used with two different style: either directly loop on it (see `showScanResults`) or use it to create a native Javascript Array (see `showImageWithResults`).

If you prefer to not show this Iterator way, I can update the demo_reader to use only C-style for loops.

The demo_reader has been improved to show result in an ordered list in text and to add the result index on each barcode found in the image.

### DIFF
--- a/wrappers/wasm/BarcodeReader.cpp
+++ b/wrappers/wasm/BarcodeReader.cpp
@@ -1,14 +1,14 @@
 /*
-* Copyright 2016 Nu-book Inc.
-*/
+ * Copyright 2016 Nu-book Inc.
+ */
 // SPDX-License-Identifier: Apache-2.0
 
 #include "ReadBarcode.h"
 
-#include <string>
+#include <emscripten/bind.h>
 #include <memory>
 #include <stdexcept>
-#include <emscripten/bind.h>
+#include <string>
 
 #define STB_IMAGE_IMPLEMENTATION
 #include <stb_image.h>
@@ -21,7 +21,8 @@ struct ReadResult
 	ZXing::Position position{};
 };
 
-std::vector<ReadResult> readBarcodesFromImageView(ZXing::ImageView iv, bool tryHarder, const std::string& format, const int maxSymbols = 0xff)
+std::vector<ReadResult> readBarcodesFromImageView(ZXing::ImageView iv, bool tryHarder, const std::string& format,
+												  const int maxSymbols = 0xff)
 {
 	using namespace ZXing;
 	try {
@@ -51,7 +52,8 @@ std::vector<ReadResult> readBarcodesFromImageView(ZXing::ImageView iv, bool tryH
 	return {};
 }
 
-std::vector<ReadResult> readBarcodesFromImage(int bufferPtr, int bufferLength, bool tryHarder, std::string format, const int maxSymbols)
+std::vector<ReadResult> readBarcodesFromImage(int bufferPtr, int bufferLength, bool tryHarder, std::string format,
+											  const int maxSymbols)
 {
 	using namespace ZXing;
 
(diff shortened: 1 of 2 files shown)


wrappers/wasm/demo_reader.html
```diff
@@ -20,49 +60,81 @@
...
-		ctx.lineTo(position.topRight.x, position.topRight.y);
-		ctx.lineTo(position.bottomRight.x, position.bottomRight.y);
-		ctx.lineTo(position.bottomLeft.x, position.bottomLeft.y);
-		ctx.closePath();
-		ctx.strokeStyle = "red";
-		ctx.lineWidth = 5;
-		ctx.stroke();
+
+		// Example to use JavaScript Array created from an Iterator
+		const resultArray = Array.from(new ReaderResultVectorIterator(results));
```
axxel, line 86: Creating an array with a copy of the data here, only to then iterate over it in an indexed for loop, does not make much sense to me. Please either use the .size(), .get() API or the `for( ... on ...)` syntax.

wrappers/wasm/demo_reader.html
```diff
@@ -12,6 +12,46 @@
 	zxing = instance; // this line is supposedly not required but with current emsdk it is :-/
 });
 
+/*
```
axxel, line 15: This iterator seems a bit over the top, just to replace a `for( int ... )` with a `for( ... on ... )` loop in this simple demo, but I generally like abstractions and maybe someone learns something from it for a 'real' application. Keep it or leave it, as you prefer.

axxel (changes requested):
Thanks.

Trim:
Thanks for the review. I've just added commits to:

* use `vector<ReaderResult>` instead of a new structure
* use the new functions to implement original ones
* apply clang-format on `BareCodeReader.cpp`
* update the demo_reader to just use `.size()` and `.get()`

wrappers/wasm/BarcodeReader.cpp
```diff
@@ -55,40 +62,56 @@ ReadResult readBarcodeFromImage(int bufferPtr, int bufferLength, bool tryHarder,
...
+	return readBarcodesFromImageView({reinterpret_cast<uint8_t*>(bufferPtr), imgWidth, imgHeight, ImageFormat::RGBX}, tryHarder,
+									 format, maxSymbols);
 }
 
 ReadResult readBarcodeFromPixmap(int bufferPtr, int imgWidth, int imgHeight, bool tryHarder, std::string format)
 {
 	using namespace ZXing;
-	return readBarcodeFromImageView({reinterpret_cast<uint8_t*>(bufferPtr), imgWidth, imgHeight, ImageFormat::RGBX}, tryHarder, format);
+	auto results = readBarcodesFromPixmap(bufferPtr, imgWidth, imgHeight, tryHarder, format, 1);
+	return results.front();
```
axxel, line 89: Same here.

axxel (changes requested):
Two one-line fixes remaining. Looking very good but I have not tested the code yet. I trust that you did ;).

Trim:
> Two one-line fixes remaining. Looking very good but I have not tested the code yet. I trust that you did ;).

Thanks for the review, I've just fixed it.

I've just tested manually with the old and new demo_reader page and images available in the test/samples/multi-1 folder :)

axxel:
Thanks a lot. I decided to squash the 6 commits into 1.

axxel:
It looks like immediately after I merged this, someone discovered an issue with the new code: #568. Could you maybe have a look at it?

PR #573, opened 2023-06-09
fix obtaining pattern in case of negative  pixstride (rotation of 90 …

fedormsv:
fix obtaining pattern in case of negative  pixstride (rotation of 90 or 180 degrees from original layout):
if pixstride is negative, end pointer will be smaller than begin and loop will never be executed


axxel:
One line patches are my favorite ;). Thanks!

PR #588, opened 2023-07-14
WASM: bytes in `ReadResult`

Sec-ant:
Safely return bytes as `Uint8Array` in `ReadResult` via `emscripten::val::global`

should fix the issue mentioned in https://github.com/zxing-cpp/zxing-cpp/pull/584 and https://github.com/zxing-cpp/zxing-cpp/commit/8a0c458dc2284dcaa1a1393510fdf5e8f7735041

### DIFF
--- a/wrappers/wasm/BarcodeReader.cpp
+++ b/wrappers/wasm/BarcodeReader.cpp
@@ -7,6 +7,7 @@
 #include "ReadBarcode.h"
 
 #include <emscripten/bind.h>
+#include <emscripten/val.h>
 #include <memory>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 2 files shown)


wrappers/wasm/BarcodeReader.cpp
```diff
@@ -49,9 +44,18 @@ std::vector<ReadResult> readBarcodes(ImageView iv, bool tryHarder, const std::st
 		std::vector<ReadResult> readResults{};
 		readResults.reserve(results.size());
 
+		thread_local const emscripten::val Uint8Array = emscripten::val::global("Uint8Array");
+
 		for (auto&& result : results) {
-			readResults.push_back({ToString(result.format()), result.text(), std::string(result.bytes().asString()),
-								   ToString(result.error()), result.position(), result.symbologyIdentifier()});
+			ByteArray bytes = result.bytes();
```
axxel, line 50: Very nice! I have one tiny improvement to suggest: this line makes a copy of result.bytes(), which is unnecessary. Either use `result.bytes().data()` below or change this line to `const ByteArray& bytes = result.bytes()`. I have not tested your changes myself, will do.

Sec-ant: Thanks, fixed now.

PR #596, opened 2023-07-30
Fix QT 6 QML ComboBox Crash & Enable Shape

LeonnardoVerol:
Sadly, this is not the Image Crash, just the QML one
PS: GitIgnore Improved a bit

image


axxel:
Thanks for your contribution. Much appreciated. There is something off with the math in your `mapPointToItem` function, though. See the attached screenshot. The red quadrilateral is supposed to exactly surround the QRCode.
#596-wrong-math

LeonnardoVerol:
You are right, I managed to Fix the math, with this code:

```
function mapPointToItem(point)
{
    const xPointPercentage = 100 * point.x / videoOutput.sourceRect.width;
    const yPointPercentage = 100 * point.y / videoOutput.sourceRect.height;

    const xMapped = (xPointPercentage * videoOutput.contentRect.width / 100) + videoOutput.contentRect.x;
    const yMapped = (yPointPercentage * videoOutput.contentRect.height / 100) + videoOutput.contentRect.y;

    return Qt.point(xMapped, yMapped);
}
```

But I ended up swapping that code with the QT5 Logic which accounts for more stuff like orientation etc. 

PR Updated

LeonnardoVerol:
BTW, The QT5 Logic is an adaptation from the Github QT5 Source code. If That`s a problem, I can revert back to my simple math

axxel:
Thanks for the fix. I took the liberty to condense the 4 new functions into one. I would like to ask you to do the following: create a new PR with only the final version of the qml-file, not including the .gitignore changes.

LeonnardoVerol:
Will do

PR #597, opened 2023-07-30
Fix QT 6 QML Crash & Enable QML Shape

LeonnardoVerol:
New PR without gitignore update

Let me know if you meant a new clean branch without all those commits


axxel:
I meant a 'clean' one but I simply used the squash-and-merge feature. Thanks for your contribution.

Issue #592, opened 2023-07-21
Support for non-ISO8859_1 characters in DataMatrix generation missing


wwzh2015:
How to Generate QR code that contain chinese word? But contain English word is exactly.

e.x:
int main(int argc, char* argv[])
{
	CharacterSet encoding = CharacterSet::UTF8;
	BarcodeFormat format = BarcodeFormat::QRCode;

	int width = 500, height = 500;
	int margin = 10;    //空白
	int eccLevel = -1;	//纠错级别
	
	std::string text = "公共号1hgreh346346w+_=@#$%| a|";
	std::string filePath = "qr1.png";

	if (!ParseOptions(argc, argv, &width, &height, &margin, &encoding, &eccLevel, &format, &text, &filePath)) {
		PrintUsage(argv[0]);
		return -1;
	}

	try {
		auto writer = MultiFormatWriter(format).setMargin(margin).setEncoding(encoding).setEccLevel(eccLevel);
		
		auto bitmap = ToMatrix<uint8_t>(writer.encode(TextUtfEncoding::FromUtf8(string_To_UTF8(text)), width, height));

		auto ext = GetExtension(filePath);
		int success = 0;
		if (ext == "" || ext == "png") {
			success = stbi_write_png(filePath.c_str(), bitmap.width(), bitmap.height(), 1, bitmap.data(), 0);
		}
		else if (ext == "jpg" || ext == "jpeg") {
			success = stbi_write_jpg(filePath.c_str(), bitmap.width(), bitmap.height(), 1, bitmap.data(), 0);
		}

		if (!success) {
			std::cerr << "Failed to write image: " << filePath << std::endl;
			return -1;
		}
	}
	catch (const std::exception& e) {
		std::cerr << e.what() << std::endl;
		return -1;
	}

	std::system("Pause");
	return 0;
}

axxel:
If call `ZXingWriter -encoding utf8 qrcode "公共号1hgreh346346w+_=@#$%| a|" test.png` and then `ZXingReader test.png` I get exactly the input string back. What is your problem?

I suggest you replace `TextUtfEncoding::FromUtf8(string_To_UTF8(text))` with `text` and try again.

wwzh2015:
> replace `TextUtfEncoding::FromUtf8(string_To_UTF8(text))` with

My problem is generate fail when the QR code contain chinese word, english word is successful.

How can successful generate QR code that contain chinese word!

axxel:
[...]

PR #599, opened 2023-08-02
MultiFormatWriter: fix encoding binary std::string

markusfisch:
While `MultiFormatWriter::encode(std::wstring)` can encode binary data just fine, `MultiFormatWriter::encode(std::string)` corrupts the data by trying to interpret it as UTF-8.

This may also break immediately because not all byte combinations are allowed in UTF-8.

So when `_encoding` is set to `CharacterSet::BINARY`, the `std::string` needs to be converted to a `std::wstring` of unsigned values, because `TextEncoder::GetBytes(std::wstring)` calls `ToUtf8()` which cannot handle negative values.

For example, without this commit, try:

	auto writer = MultiFormatWriter(BarcodeFormat::QRCode)
		.setEncoding(CharacterSet::BINARY).
	auto bitmap = writer.encode(std::string("\x7e\x7f\x80\x81"), 200, 200);

Which will result in: "ValueError: Unexpected charcode".

For a QRCode, `MultiFormatWriter::encode(std::string("\x7e\x7f\x80\x81"))` will run:

1. MultiFormatWriter::encode(std::wstring)
2. Writer::encode(std::wstring)
3. QREncoder::Append8BitBytes(std::wstring)
4. TextEncoder::FromUnicode(std::wstring)
5. TextEncoder::GetBytes(std::wstring) <-- here it will call `ToUtf8(str)`
6. TextEncoder::GetBytes(std::string) where `str` will now be `0x7e, 0x7f, 0xff, 0xbf, 0xbe, 0x80, 0xff, 0xbf, 0xbe, 0x81`

Please note ZXingWriter is **NOT** affected by this!
This works just fine:

	$ printf "\\x7e\\x7f\\x80\\x81" > file
	$ example/ZXingWriter -binary QRCode file out.png

Because:
1. it's calling `MultiFormatWriter::encode(std::wstring)` and
2. builds the `std::wstring` from `uint8_t`

### DIFF
--- a/core/src/MultiFormatWriter.cpp
+++ b/core/src/MultiFormatWriter.cpp
@@ -67,9 +67,24 @@ MultiFormatWriter::encode(const std::wstring& contents, int width, int height) c
 	}
 }
 
+static std::wstring ToUInt8WideString(const std::string& s) {
+	std::wstring ws(s.begin(), s.end());
+	// Keep only the first byte of `wchar_t` and remove the sign bits
+	// from the `char` conversion. Without this, any `char` > 127 will
+	// be negative in `wchar_t`, too, which breaks later.
+	for (int i = 0, l = ws.length(); i < l; ++i) {
+		ws[i] &= 0xff;
+	}
+	return ws;
+}
+
 BitMatrix MultiFormatWriter::encode(const std::string& contents, int width, int height) const
 {
-	return encode(FromUtf8(contents), width, height);
+	return encode(
+		_encoding == CharacterSet::BINARY
+			? ToUInt8WideString(contents)
+			: FromUtf8(contents),
+		width, height);
 }
 
 } // ZXing


axxel:
Your fix/workaround is exactly what I did in `ZXingReader` and also suggested in https://github.com/zxing-cpp/zxing-cpp/pull/557#issuecomment-1527859569. I see that there is a point to be made to make this more convenient but for one, your change is technically breaking the API (`encode(std::string)` takes UTF8 input) and it is also still not self-explanatory and a bit of a "hack" given the requirement to pass `BINARY` as the encoding character set. I'd like to make sure this use case is properly handled in the new API (see https://github.com/zxing-cpp/zxing-cpp/issues/332#issuecomment-1666612061) but am wondering if your problem could not also be solved by simply doing this on the client side (see the two examples above) until a proper solution is available?

markusfisch:
Okay, I look forward to the new API then 😉

axxel:
> DataMatrix seems to especially problematic here, probably because of this [transformation](https://github.com/zxing-cpp/zxing-cpp/blob/master/core/src/datamatrix/DMHighLevelEncoder.cpp#L882).

Exactly. This has indeed come up already earlier this year (see [here](https://github.com/zxing-cpp/zxing-cpp/issues/592#issuecomment-1646851749)).

> So before taking action I thought it would be wise to check your opinion on that.

Very wise ;). I have to admit that my progress on the new Writer API has been stalled recently. I have an early prototypish hack lying around but there was a lack of time and there are still quite a few open questions regarding the types used in the API (as mentioned in #332)

If you need this 'now' I see two options:
 1. you find a quick (and dirty?) hack to make the existing DataMatrix encoder work with binary data.
 2. I try to throw something out there behind the `EXPERIMENTAL` build option soonish and you can be the first to alpha-test it ;)

markusfisch:
Thanks for the quick reply!

I think I'll try to hack something together then as it is a bit urgent, but of course, I would always happily be the first to alpha-test your solution! 😉

If you maybe already have a concrete idea, please tell me so 😉

markusfisch:
@axxel 
Here's what I did to make it work in case you want to merge it: https://github.com/zxing-cpp/zxing-cpp/pull/628

PR #616, opened 2023-09-11
Python Multithreading Support via GIL

billmccartney:
Enabled parallel Python support by disabling the GIL during ReadBarcodes. This allows multiple threads to decode different images at the same time in Python.
pybind11 notes are here https://pybind11.readthedocs.io/en/stable/advanced/misc.html#global-interpreter-lock-gil

### DIFF
--- a/wrappers/python/zxing.cpp
+++ b/wrappers/python/zxing.cpp
@@ -90,6 +90,8 @@ auto read_barcodes_impl(py::object _image, const BarcodeFormats& formats, bool t
 	}
 
 	const auto bytes = image.data();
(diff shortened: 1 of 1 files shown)


axxel:
Thanks for your contribution. Question: do you have a use case where this actually increases your throughput? If so, how big are your input images?

axxel:
Thanks for your benchmarking information. Makes your change look obviously very worthwhile. :). If you care about performance, make sure you only look for symbologies (`BarcodeFormats`) that you are actually interested in. It sounds like you have a rather standardized set of input images. I'd be interested to see a sample if possible, maybe I can provide more tips on how to improve the performance for that type of input.

PR #618, opened 2023-09-12
Expose return_errors parameter to Python wrapper APIs.

gregtws:
Creating this draft PR based on discussion in https://github.com/zxing-cpp/zxing-cpp/discussions/617. I've tried to keep the API compatible with defaults reflecting existing behavior.  I'm not super familiar with pybind and if there were specific doc patterns you'd prefer so hopefully what I've done is reasonable.

One challenge is since existing APIs return `None` when there's no match (regardless of errors), with `return_errors=True` we have a somewhat awkward mode where it could return `None` (nothing at all found) or a result with errors (something found, but errors occurred).  This forces callers in this mode to check twice for a clean scan result via `if result is not None and result.valid`.  I'm not super excited about this, but seemed like the least bad way without changing the API completely (like a separate function).

By default errors will not be returned (so API compatible with defaults).  When errors are requested, results with errors will only be returned when possible barcodes were found.

### DIFF
--- a/wrappers/python/zxing.cpp
+++ b/wrappers/python/zxing.cpp
@@ -37,7 +37,8 @@ std::ostream& operator<<(std::ostream& os, const Position& points) {
 }
 
 auto read_barcodes_impl(py::object _image, const BarcodeFormats& formats, bool try_rotate, bool try_downscale, TextMode text_mode,
-						Binarizer binarizer, bool is_pure, EanAddOnSymbol ean_add_on_symbol, uint8_t max_number_of_symbols = 0xff)
+						Binarizer binarizer, bool is_pure, EanAddOnSymbol ean_add_on_symbol, bool return_errors,
+						uint8_t max_number_of_symbols = 0xff)
 {
 	const auto hints = DecodeHints()
 		.setFormats(formats)
@@ -47,7 +48,8 @@ auto read_barcodes_impl(py::object _image, const BarcodeFormats& formats, bool t
 		.setBinarizer(binarizer)
 		.setIsPure(is_pure)
 		.setMaxNumberOfSymbols(max_number_of_symbols)
-		.setEanAddOnSymbol(ean_add_on_symbol);
+		.setEanAddOnSymbol(ean_add_on_symbol)
+		.setReturnErrors(return_errors);
 	const auto _type = std::string(py::str(py::type::of(_image)));
 	py::buffer buffer;
 	ImageFormat imgfmt = ImageFormat::None;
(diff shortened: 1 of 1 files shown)


wrappers/python/zxing.cpp
```diff
@@ -287,8 +311,13 @@ PYBIND11_MODULE(zxingcpp, m)
 		":type ean_add_on_symbol: zxing.EanAddOnSymbol\n"
 		":param ean_add_on_symbol: Specify whether to Ignore, Read or Require EAN-2/5 add-on symbols while scanning \n"
 		"  EAN/UPC codes. Default is ``Ignore``.\n"
+		":type return_errors: bool\n"
+		":param return_errors: Set to True to return the barcodes with errors as well (e.g. checksum errors).\n"
+		" Default is False."
 		":rtype: zxing.Result\n"
-		":return: a zxing result containing decoded symbol if found, None otherwise"
```
axxel, line 291: please leave the `return` doc string as is.

wrappers/python/zxing.cpp
```diff
@@ -324,9 +354,12 @@ PYBIND11_MODULE(zxingcpp, m)
 		":type ean_add_on_symbol: zxing.EanAddOnSymbol\n"
 		":param ean_add_on_symbol: Specify whether to Ignore, Read or Require EAN-2/5 add-on symbols while scanning \n"
 		"  EAN/UPC codes. Default is ``Ignore``.\n"
+		":type return_errors: bool\n"
+		":param return_errors: Set to True to return the barcodes with errors as well (e.g. checksum errors).\n"
+		" Default is False.\n"
 		":rtype: zxing.Result\n"
-		":return: a list of zxing results containing decoded symbols, the list is empty if none is found"
```
axxel, line 328: please leave the `return` doc string as is.

wrappers/python/zxing.cpp
```diff
@@ -287,8 +311,13 @@ PYBIND11_MODULE(zxingcpp, m)
 		":type ean_add_on_symbol: zxing.EanAddOnSymbol\n"
 		":param ean_add_on_symbol: Specify whether to Ignore, Read or Require EAN-2/5 add-on symbols while scanning \n"
 		"  EAN/UPC codes. Default is ``Ignore``.\n"
+		":type return_errors: bool\n"
+		":param return_errors: Set to True to return the barcodes with errors as well (e.g. checksum errors).\n"
```
axxel, line 315: maybe add "see Result.error" at the end here?

axxel:
> we have a somewhat awkward mode where it could return `None` (nothing at all found) or a result with errors (something found, but errors occurred).

I don't see anything awkward with this approach at all. That is exactly how the c++ API works and if the caller is not interested in erroneous symbols, then they can simply leave the `return_errors` parameter alone.

gregtws:
Made fixes, pushed a rebased version of the branch.

PR #621, opened 2023-09-22
Ignore .idea

okarmazin:
.idea should never be checked in. Applied globally to ensure it cannot creep in in any subproject.

This can be thought of as a completion of https://github.com/zxing-cpp/zxing-cpp/commit/52a5f2e7501ff9017536958a2056dec4a1a2dceb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ CMakeLists.txt.user
 *.d
 *.a
 compile_commands.json
+.idea


axxel:
Since this is already taken care of in the wrapper/android/.gitignore, I'd consider this issue solved. Are you using AndroidStudio anywhere "outside" of the wrapper/android directory?

tiny-dancer:
I beleive .idea is a configuration file for all jetbrains ide's, including clion

benjohnde:
Depends a bit on the usage though, won't do any harm.

axxel:
Closed for lack of feedback from the author.

PR #622, opened 2023-09-22
Correctly handle supported image formats of Android APIs < 23

okarmazin:
WIP: Rebase on master after #619 

The library declares minSdk 21, but `BarcodeReader` accesses ImageFormat constants that were added in API 23. This PR guards the access to these constants with a version check.

Also, the [explicit API mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors) was enabled for the Kotlin compiler. This flag instructs the compiler to enforce visibility modifiers and return types be explicitly specified (if they would be public by default).


axxel:
Just to make I understand the situation regarding the explicit mode: Everything in a kotlin class is public unless specified private, right? So if there is no `private` visibility modifier, the symbol is (explicitly) public. What is there to gain when littering the code with a bunch of `public` strings? Preventing an ignorant developer from accidentally adding public symbols to the API?

okarmazin:
Yes, unfortunately Kotlin visibility is implicitly `public` by default, unless specified otherwise.

Explicit API mode helps prevent two errors:

1) Makes it harder to accidentally publish implementation details in your public API. The requirement to specify visibility makes you think about your symbols and if they are actually supposed to be accessible by library consumers.


2) Prevents type inference from changing your return types by mistake. Explicit API mode requires that return types be specified explicitly instead of relying on type inference. It's the same as if you disallowed `auto` in public API in C++.

axxel:
As you can possibly infer from my choice of words in my last comment, I'm not a fan of 'unnecessary' verbosity but I'm also not particularly emotional about the Kotlin code base in general and you clearly have more knowledge and experience here... Thanks for taking the time to engage in the discussion.

PR #624, opened 2023-10-05
Add required framework versions to MACOSX builds

alexmanzer:
With Xcode15 it is required, that frameworks must have in their Info.plist two (previously not needed) values: CFBundleShortVersionString and CFBundleVersion.

If they are missing, a build will be successful, but no upload to the AppStore is possible, it will fail with this error codes: Asset validation failed (90057)
Asset validation failed (90056)

Screenshot 2023-09-22 at 00 29 37

see https://stackoverflow.com/a/52298841 and
https://cmake.org/cmake/help/v3.0/prop_tgt/MACOSX_FRAMEWORK_INFO_PLIST.html


axxel:
Thanks a lot.

PR #625, opened 2023-10-05
Build wheels for Python 3.12.

dainnilsson:
This updates the python-build workflow to use the latest version of cibuildwheel and build wheels for the newly released Python 3.12.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,7 +140,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11']
+        python-version: ['3.12']
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,7 +154,7 @@ jobs:
     - name: Install dependencies
       working-directory: wrappers/python
       run: |
-        python -m pip install --upgrade pip
+        python -m pip install --upgrade pip setuptools
         python -m pip install numpy pillow
 
     - name: Build module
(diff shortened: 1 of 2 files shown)


axxel:
Thanks. I have two comments:

1) Could you also change the line https://github.com/zxing-cpp/zxing-cpp/blob/a6a181fba91ca4aa5f2b97f7873d9af6f3027d65/.github/workflows/ci.yml#L143 accordingly? Disclaimer, I don't know if this would actually work a.t.m. i.e. whether or not the CI environment already supports python 3.12

2) I decided in the past to only "support" / deliver prebuild wheels for most current 3 versions, which would mean dropping 3.9. Any thoughts on that practice?

dainnilsson:
1. Yeah, absolutely! I can confirm that 3.12 is supported by setup-python already. I'll push a new commit for that.
2. Certainly not a problem for me. I suppose if you wanted to maximize support then all officially supported python versions might be nice to support (3.8 and up) but that would obviously mean 5 concurrent versions at the moment. If nobody has "complained" about the lack of 3.8 support thus far, then I'd guess nobody would miss 3.9 either.

axxel:
Thinking about this some more I now believe that we should do the following:
 1. have `python-build` keep 3.9 (or maybe even add 3.8 if that is still officially supported)
 2. have `CI` only build with one python version, say 3.12

Reasoning:
 * I've never seen that only a single python version from `CI` failed.
 * The `python-build` is currently separate because it takes so much time, so it is only run near the next release. If it takes even longer, that does not really make much of a difference to me.
 * That said, for people who just want to download wheels from PyPi, it makes a substantial difference if they have to setup a c++ build environment just to build the wheel from the source package.

dainnilsson:
Your reasoning sounds fine to me, I'll update the PR!

dainnilsson:
I've now made the above changes, and tested to make sure both the ci and python-build workflows run successfully. I had to add an explicit step to install `setuptools` , which I believe was previously automatically added with pip, but no longer is.

PR #628, opened 2023-10-10
DataMatrix: use charset for encoding

markusfisch:
So UTF8 and binary contents are encoded correctly.


axxel:
Thanks. That was less invasive than I expected.

markusfisch:
That's what I thought! 😉

Issue #623, opened 2023-09-24
Make a pod available for the iOS wraper or a way to use in a podspec


AbijahKaj:
Hello!
Is there a way to use the iOS wrapper in a `podspec` file? Or is there an available pod of the wrapper? I only found [this one](https://github.com/zxingify/zxingify-objc), but it seems outdated and it's written in objc.

Thanks in advance.

axxel:
There is currently no `podspec` file available/included in `zxing-cpp` but @benjohnde who maintains the repo you linked to just recently [mentioned](https://github.com/zxing-cpp/zxing-cpp/pull/619#issuecomment-1730923196) that he wants to work on 'our' iOS wrapper code. So maybe he would be in the best possible position to provide a CocoaPods integration? :)

benjohnde:
Hi there @AbijahKaj, you can nowadays use CocoaPods alongside with SPM. Is there any particular reason you want to use CocoaPods? I can tackle a podspec in the upcoming week, with the recent PR #630 you should be able to inject the dependency with ease by just providing this repository in Xcode (Package Dependency Window).

benjohnde:
btw you can already give it a try by specifying the branch `feat/ios-wrapper-improvement` and temporarily using my fork `https://github.com/benjohnde/zxing-cpp` as package url.

AbijahKaj:
Hey @benjohnde !
Thanks for the message, I'm making a react native module, so a Podspec is the only way to go, I tried some not-so-clean hacks to add SPM but it's not being consumed by my module.

Thanks I'll try your branch.

axxel:
As explained by @alexmanzer [here](https://github.com/zxing-cpp/zxing-cpp/pull/630#issuecomment-1766046411), the whole CocoaPods universe is about to go under soonish? I have to admit that I usually tend towards his "legacy guys" reaction and would question the introduction of some already outdated distribution solution. (I will for example require c++20 support in the upcoming `WriteBardcode` API, or maybe I should even be consequent and go for C++23, looking at my calendar ;)).

On the other hand, I also appreciate if the user base of this library grows... sigh ;).

AbijahKaj:
Hi @axxel !
As you probably know, Cocoapods is also used in many other massively used tools/frameworks, so I don't think it's going anywhere anytime soon. Just look at how many people use react native or flutter, if you want to add an iOS native module to those, the only way to go is Cocoapods.

[...]

Issue #344, opened 2022-06-17
Sensitivity to cropping


hgaiser:
For my task I am limited in the available space for a marker (roughly 1cm x 1cm only). I am running some tests to see how viable different markers are at this size, in combination with the required workspace area. Below is a crop of the area where I am expecting markers to be found:

![cropped2](https://user-images.githubusercontent.com/716138/174255093-755dc01b-2ef0-4e58-986e-5f049430dbcc.png)

From this image, 10/10 markers are correctly detected. If I use the exact same image but crop it further (so pixel density and information should remain the same) it will sometimes detect less markers than before. For example on this crop:

![cropped](https://user-images.githubusercontent.com/716138/174255128-47b3ab15-89f9-4d50-9b2b-a77e11fbcd7b.png)

It didn't detect markers 2 and 6 and it decoded marker 1 as 10.

Am I close to the limits of what this algorithm can detect? Or is there something strange going on here? Any information is appreciated, because it would help me decide the direction for my project :).

Some extra information: I am using Micro QR code because in my (brief) experiments it seemed to produce the most stable results. I used data matrices before but the detection rate and cropping sensitivity was significantly worse there. What I require is to identify roughly 20 different markers.

axxel:
Interesting problem. There are three things I'd like to say:
 1. these are all invalid MicroQRCode symbols because they don't have a 2 modules wide quite zone (the black border is too close to the symbols (I believe, if you changed that it would work).
 2. I don't know why the cropping makes any difference here. That is weird.
 3. I will try to change something, such they hopefully work even though they are not valid symbols.

hgaiser:
Ahh interesting! I didn't know the quiet zone had to be two modules wide, I will surely try that out. I will probably be checking this again next Thursday, I will let you know what I find 👍 . Thank you for the reply.

oehhar:
Am 18.06.2022 um 13:06 schrieb axxel:
> Interesting problem. There are three things I'd like to say:
> 
>  1. these are all invalid MicroQRCode symbols because they don't have a
>     2 modules wide quite zone (the black border is too close to the
>     symbols (I believe, if you changed that it would work).

AFAIK, the standard quiet zone is 3 modules, at least for QR-Code. That 
is often seen as a disadvantage compared to DMRE/DM, as the quiet zone 
is only one pixel.

Enjoy,
Harald

hgaiser:
Seems to be 4 for QR and 2 for MicroQR:

[...]

PR #630, opened 2023-10-14
iOS Wrapper improvements

benjohnde:
The old wrapper had a few disadvantages which are solved by the following PR:

- `Package.swift` resides now in root, enabling people to include `zxing-cpp` on iOS by just adding this repository as package dependency in Xcode
- `unsafeFlags` are not used anymore, thus there is no need for injecting the framework manually in a hacky way
- Creation of an 'umbrella framework' is avoided
- Demo project was improved on a minor level
- The cmake file was cleaned-up by removing the iOS bits


benjohnde:
@axxel ready for review and release, potentially 😏

benjohnde:
@alexmanzer I recognized your PR (https://github.com/zxing-cpp/zxing-cpp/pull/624), feel free to give this one a review. I removed the duplication of having a separate project to build the XCFramework which is then used by SwiftPM. One can now use `zxing-cpp` as package dependency and use the `ZXingCppWrapper` more easily.

Package.swift
```diff
@@ -9,23 +9,26 @@ let package = Package(
...
+            publicHeadersPath: "."
         ),
         .target(
             name: "ZXingCppWrapper",
             dependencies: ["ZXingCpp"],
-            path: "Sources/Wrapper",
+            path: "wrappers/ios/Sources/Wrapper",
             publicHeadersPath: ".",
-            cxxSettings: [
-                .unsafeFlags(["-stdlib=libc++"]),
```
benjohnde, line 26: Short remark: not necessary anymore due to dynamic build. I didn't need it for the static variant either though.

Package.swift
```diff
@@ -9,23 +9,26 @@ let package = Package(
...
-                .unsafeFlags(["-std=gnu++17"])
+            linkerSettings: [
+                .linkedFramework("CoreGraphics"),
+                .linkedFramework("CoreImage"),
+                .linkedFramework("CoreVideo")
             ]
         )
-    ]
+    ],
+    cxxLanguageStandard: CXXLanguageStandard.gnucxx17
```
axxel, line 33: It might be a good idea to switch to c++20 here. This will allow position-independent and multi-symbol DataMatrix detection (#344). But this comes at a performance cost, that said, my android app does not feel any different. This could follow later as well, just wanted to mention it now that I saw this line.

wrappers/ios/demo/demo.xcodeproj/project.pbxproj
```diff
@@ -27,16 +42,16 @@
 		388BF034283CC49E005CE271 /* Base */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; name = Base; path = Base.lproj/LaunchScreen.storyboard; sourceTree = "<group>"; };
 		388BF036283CC49E005CE271 /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
 		388BF043283CE0AC005CE271 /* README.md */ = {isa = PBXFileReference; lastKnownFileType = net.daringfireball.markdown; path = README.md; sourceTree = "<group>"; };
+		52A975492ADADA1B002D6BD8 /* zxing-cpp */ = {isa = PBXFileReference; lastKnownFileType = wrapper; name = "zxing-cpp"; path = "/Users/john/git/private/zxing-cpp"; sourceTree = "<absolute>"; };
```
axxel, line 45: This line definitively needs to change, even if the `path` part should be functionally irrelevant.

wrappers/ios/demo/demo/ViewController.swift
```diff
@@ -35,11 +40,13 @@ class ViewController: UIViewController {
             self.captureSession.addInput(cameraInput)
             let videoDataOutput = AVCaptureVideoDataOutput()
             videoDataOutput.setSampleBufferDelegate(self, queue: self.queue)
-            videoDataOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: Int(kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange)]
+            videoDataOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: kCVPixelFormatType_32BGRA]
```
axxel, line 43: I have a technical counter argument for this change: first of all, looking at the setup of the `ImageView` where only `ImageFormat::Lum` is used, I would assume this is simply a bug. Second of all, it is simply a waste of CPU resources to transform RGB data to grey scale inside zxing-cpp when we can have the latter right from the start.

axxel (changes requested):
Thanks a lot for the work done and the interest in merging our efforts. Since I'm not the author of most of the involved code here, I'll also ask those if they have any comments on your PR.

Package.swift
```diff
@@ -9,23 +9,26 @@ let package = Package(
     products: [
         .library(
             name: "ZXingCppWrapper",
-            type: .static,
+            type: .dynamic,
             targets: ["ZXingCppWrapper"])
     ],
```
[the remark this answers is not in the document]
benjohnde: Not really, probably they were missing some core libraries. With the old variant, you had to also add these frameworks to your app (CoreGraphics, CoreImage, CoreVideo). I can't see any benefit here in using a static library, it will increase your overall app size without any benefit at all.

Not quite sure how the original author used it, but there were a few downsides which avoided one from submitting an app to the App Store (umbrella framework) or even to build the app (unsafeFlags).

Package.swift
benjohnde: Ah nice catch! According to the [documentation](https://developer.apple.com/documentation/packagedescription/cxxlanguagestandard/cxx20), that should work easily. I will still test it and feedback afterwards.

core/CMakeLists.txt
```diff
@@ -482,25 +482,6 @@ endif()
...
-        MACOSX_FRAMEWORK_SHORT_VERSION_STRING ${PROJECT_VERSION}
-        MACOSX_FRAMEWORK_BUNDLE_VERSION ${PROJECT_VERSION}
-        CMAKE_XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH NO
-        #MACOSX_FRAMEWORK_INFO_PLIST Info.plist
-        PUBLIC_HEADER "${PUBLIC_HEADERS}"
-        XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "iPhone Developer"
-        XCODE_ATTRIBUTE_ENABLE_BITCODE "NO"
-    )
-endif()
-
```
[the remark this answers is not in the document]
benjohnde: This is obsolete! But I am glad you are asking these questions:

These additional bits were necessary to craft Frameworks merged into an XCFramework (containing the static built library - actually some renamed `*.a` file).

Framework is a library (either dynamically or statically linked) for one particular platform. It contains a bit of Apple-related meta information, that's all. XCFramework is a collection of multiple Frameworks to work (depends on how you built it, but that's the idea) on every platform (w.r.t. Apple) - an universal framework so to speak.

An `.a` (static) or in case of Apple `.dylib` (dynamic) file gets generated by the Swift Package Manager (which you can easily use) by `swift build` in the root folder (depending where the `Package.swift` file is located).

So there is no real downside in this approach, you can still build the file if you just need the library without any additional foo.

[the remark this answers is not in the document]
benjohnde: Ah forgot to mention, as we can now include `zxing-cpp` as a direct dependency with the Xcode dependency management, Xcode takes care on building and including `zxing-cpp` in the app / project.

wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
```diff
@@ -3,9 +3,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #import "ZXIBarcodeReader.h"
-#import "ZXing/ReadBarcode.h"
-#import "ZXing/ImageView.h"
-#import "ZXing/Result.h"
+#import "ReadBarcode.h"
```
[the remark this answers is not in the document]
benjohnde: Yes

wrappers/ios/demo/demo.xcodeproj/project.pbxproj
benjohnde: OH DAMN IT, I reviewed every single change sigh :) commit incoming.

axxel:
To those who had their hands on any of the iOS/macOS wrapper code before: would you care to comment on this PR?

There are two changes that I'd like to specifically point to:
 * `build-release.sh` script removal, which might also loose the separate `iphonesimulator` Framework?
 * the removal of the 'core' ZXingCpp framework (see `core/CMakeLists.txt`)

@parallaxe, @Alex-MSFT, @alexmanzer, @ChristianNorbertBraun, @ryantrem

wrappers/ios/demo/demo/ViewController.swift
benjohnde: I was basically referring to that method here https://github.com/zxing-cpp/zxing-cpp/blob/master/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm#L56 and thought it would be fine using BGRA.

The reason to change that is that modern devices natively output 32BGRA. So using YUV is an additional conversion on the phone and - if my knowledge is not completely wrong - is lossy. But maybe that doesn't do any harm, I can easily revert it back.

benjohnde:
> To those who had their hands on any of the iOS/macOS wrapper code before: would you care to comment on this PR?
> 
> There are two changes that I'd like to specifically point to:
> 
> * `build-release.sh` script removal, which might also loose the separate `iphonesimulator` Framework?
> * the removal of the 'core' ZXingCpp framework (see `core/CMakeLists.txt`)
> 
> @parallaxe, @Alex-MSFT, @alexmanzer, @ChristianNorbertBraun, @ryantrem

Just to clarify on these two points:
- the respective simulator frameworks get automatically built by the compiler using the Swift PM
- the additional core framework was not necessary in resulted in an umbrella framework

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
```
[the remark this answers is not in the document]
benjohnde: Yes, exactly!

For version control, the tags of the repository are used, cf. https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#handling-version-specific-logic.

There were some discussions in allowing custom directory layouts (i.e. where the `Package.swift` has to be located), cf. https://github.com/apple/swift-package-manager/issues/5339, https://github.com/apple/swift-package-manager/issues/5768.

benjohnde:
@axxel I will wait feedback from the other guys and then also adjust the gitlab action. I've overlooked that one, sry!

core/CMakeLists.txt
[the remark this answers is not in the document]
benjohnde: Just to clarify that: you can do both, use it in your app or just build solely the library.

wrappers/ios/demo/demo/ViewController.swift
axxel: I see, so the 'bug' part of my comment was b.s. But the alternative path then explicitly asks the OS to convert the input into an `kCGColorSpaceGenericGray` image, which is then fed into `ReadBarcodes`. It would be possible to pass `BGRA` pixels directly but they would then only be converted to grey scale data internally. And having an OS level function do that is much more likely to be optimal in terms of CPU usage.

All that said, I could imagine a situation where converting `BGRA` to greyscale is in-fact cheaper than asking for `kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange` if the original raw data was in fact already `BGRA` because the two chroma channels are then computed for nothing. But if you are running on some hardware where the 'raw' data you get out of your sensor/preprocessor is Yuv, then simply passing the Y-channel is obviously the most effective option. And I understood that this is actually the most prevalent situation in the mobile phone space?

benjohnde:
Additional remark: if there is need for a solely published XCFramework, I would rather use the new configuration with some tool like https://github.com/unsignedapps/swift-create-xcframework.

wrappers/ios/demo/demo/ViewController.swift
benjohnde: I changed it back, there was no need to do the `Int()` cast.

benjohnde: @axxel Still not quite sure how your internal binarizer works with RGB images and if that makes really a difference but I was kind of right:

https://developer.apple.com/documentation/corevideo/kcvpixelformattype_420ypcbcr8biplanarvideorange

Lum in the YUV format only ranges from 16 to 235.

benjohnde:
CI was adjusted in [8f7f0e8](https://github.com/zxing-cpp/zxing-cpp/pull/630/commits/8f7f0e8d8b697a06af7f77c73bb5618e16362965), was a low-hanging fruit.

wrappers/ios/demo/demo/ViewController.swift
parallaxe: As far as I remember, @ChristianNorbertBraun and I checked the native format of some iPhones when developing the code, and they supported `kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange` as native format. From our understanding, this is the most efficient and broadly supported format. It may have changed for newer iPhones though, but I assume that it still true.

parallaxe:
First off, thanks @benjohnde for this PR.
Secondly, it would be great to also get opinions from the community about this PR, as this is a breaking change how the framework is build and used on iOS.

I discussed the consequences of the changes on our side with @alexmanzer. Our overall conclusion is: This is a very _welcome pull request_. Some of the decisions we made at the beginning are no longer relevant for ourself (i.e. having a build script that creates an xcframework that can then be distributed with Cocoapods - we are switching over to Swift Package Manager, and this PR makes this easier).

**But there is one change that would break our requirements**: The change from a _static_ library to a _dynamic_ one. We build frameworks for customers that use ZXing internally. But this is an internal detail the customer should not care about. Our philosophy is to hand them over one framework, without any external dependencies. But with a dynamic library, they have to add ZXing as additional dependency. It may seem like a little detail, but for us, this is important. 

Also, from what I know, there are no real disadvantages of static libraries vs. dynamic ones.
@benjohnde you wrote that static frameworks `will increase your overall app size` - this should not be the case for most of the apps out there. Using a dynamic framework will copy all of the framework into your app, no matter what portion of the framework you use (except this has changed meanwhile - I'm no longer an iOS developer and am not up to date from this years changes). Using a static framework, the compiler will strip out code that is not used by the app / framework that links against it - so the overall app size may in fact be smaller.
Though, there might be the case where a dynamic framework reduces the overall size. If the app contains extensions that also use ZXing, they may share all of the dynamic libraries, instead of linking ZXing statically into the app and every extension. But I'm not even sure about that, and [looking at the list of possible extensions](https://developer.apple.com/app-extensions/), it most probably will only be a corner case of the corner case.

@benjohnde Do I overlook something that regards to dynamic vs. static? Would it be ok for you to change it to static? I'm not sure if Swift Package Manager may offer the possibility for the consumer of the package to choose between static or dynamic - if there is a way, it should be the best of both worlds.

benjohnde:
That is true to a certain degree, as only Apple frameworks are being used, it does probably not make a huge difference but normally dynamic libraries are smaller, as you don't link (store) other libraries in that one file. I won't be very detailed here as there is enough literature out there, one quick link would probably be https://theswiftdev.com/deep-dive-into-swift-frameworks/ or https://belkadan.com/blog/2022/02/Dynamic-Linking-and-Static-Linking/ (besides the cheesy title, good read though). I can certainly feel the need for a static library sometimes, we also have that requirement for one of our customer (not related to zxing-cpp though).

My idea would then be (I should have implemented it like that right from the start) to omit `.dynamic` and `.static`. Then it gets built either dynamically or statically, depending on the target environment which is probably what you guys need, cf. [library(name:type:targets:)
](https://developer.apple.com/documentation/packagedescription/product/library(name:type:targets:)). Is that something we should experiment with? :)

> "as this is a breaking change how the framework is build and used on iOS."

Still don't know how you have used the library so far, for us the built framework was not compliant to the App Store guidelines. Have you ever distributed that via the App Store (or TestFlight) or used some hack to get it injected properly? The produced library was not usable for us in an app which was sent to Apple. Maybe we did a mistake there. I always got the following two issues:

- unsafeFlags are being used in a dependency, cf. https://github.com/apple/swift-package-manager/pull/1884
- umbrella framework, cf. [Don’t Create Umbrella Frameworks
](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/CreationGuidelines.html#//apple_ref/doc/uid/20002254-104843) and [Umbrella frameworks are not supported on iOS, watchOS, or tvOS](https://developer.apple.com/library/archive/technotes/tn2435/_index.html#//apple_ref/doc/uid/DTS40017543-CH1-PROJ_CONFIG-APPS_WITH_DEPENDENCIES_BETWEEN_FRAMEWORKS)

benjohnde:
> Using a dynamic framework will copy all of the framework into your app, no matter what portion of the framework you use (except this has changed meanwhile - I'm no longer an iOS developer and am not up to date from this years changes). Using a static framework, the compiler will strip out code that is not used by the app / framework that links against it - so the overall app size may in fact be smaller.

Just realised, maybe you have mixed up both terms. Or I did a mistake here. You are describing the comparison between dynamic frameworks (Apple will strip code depending on the target device / platform) and static frameworks (the library contains everything to run independently on frameworks, available on the target).

alexmanzer:
> Still don't know how you have used the library so far, for us the built framework was not compliant to the App Store guidelines. Have you ever distributed that via the App Store (or TestFlight) or used some hack to get it injected properly? The produced library was not usable for us in an app which was sent to Apple. Maybe we did a mistake there. I always got the following two issues:
> 
> * unsafeFlags are being used in a dependency, cf. [[PackageGraph] Disallow unsafe build flags in dependencies apple/swift-package-manager#1884](https://github.com/apple/swift-package-manager/pull/1884)
> * umbrella framework, cf. [Don’t Create Umbrella Frameworks
>   ](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/CreationGuidelines.html#//apple_ref/doc/uid/20002254-104843) and [Umbrella frameworks are not supported on iOS, watchOS, or tvOS](https://developer.apple.com/library/archive/technotes/tn2435/_index.html#//apple_ref/doc/uid/DTS40017543-CH1-PROJ_CONFIG-APPS_WITH_DEPENDENCIES_BETWEEN_FRAMEWORKS)

Yes, we have some apps in the AppStore that use zxing-cpp, among others [CodeScan](https://apps.apple.com/en/app/code-scan-liest-jeden-qrcode/id1554812545) and [DigitalProductPassport](https://apps.apple.com/en/app/digital-product-passport/id1643687835) . 
(With Xcode15 there was a new problem with the app upload, see here: https://github.com/zxing-cpp/zxing-cpp/pull/624 but with the change it works again with Xcode15).

You are right, the thing with the unsafeFlags was also a thorn in our side, that only worked if you either select [the branch or a commit directly](https://forums.swift.org/t/override-for-unsafeflags-in-swift-package-manager/45273/2). The versioning as it is actually intended with SPM (via tag) is unfortunately not possible with unsafe flags. Therefore in general I am very pleased with your changes. 🙏


But the second argument (with the "umbrella framework") is not valid at all, from my perspective.
The "old way" creates two .frameworks which then are assembled into one .xcframework. And this .xcframework is used in our iOS apps. There is no umbrella framework at all needed for apps, that uses zxing-cpp.


**Maybe as a general comment to this PR:**
At the moment we also have an extra repo `zxing-cpp-spm` where we push the built `.xcframework` along with the `Package.swift` (in root) and `Sources/Wrapper/` so we can get it easier via Xcode SPM.
But in general having the Package.swift in the root of the Repo would solve this also. 💪🏻
(In the past it was not so easy to place a file in the root folder here, so we had to go this workaround. 🤷🏻‍♂️)

benjohnde:
Ouh nice!

> You are right, the thing with the unsafeFlags was also a thorn in our side, that only worked if you either select [the branch or a commit directly](https://forums.swift.org/t/override-for-unsafeflags-in-swift-package-manager/45273/2). The versioning as it is actually intended with SPM (via tag) is unfortunately not possible with unsafe flags. Therefore in general I am very pleased with your changes. 🙏

Thanks, I wasn't aware of that! We still get the warning 'Umbrella framework' though.

> But the second argument (with the "umbrella framework") is not valid at all, from my perspective.
The "old way" creates two .frameworks which then are assembled into one .xcframework. And this .xcframework is used in our iOS apps. There is no umbrella framework at all needed for apps, that uses zxing-cpp.

As the unsafeFlag issue existed, I created a new XCFramework with the Wrapper files which then umbrella'd the other one. So I resulted in `ZXingCppWrapper.xcframework` which contained `ZXingCpp.xcframework`. So maybe that was totally on me and not that you guys had in mind.

Package.swift
```diff
@@ -9,23 +9,25 @@ let package = Package(
     products: [
         .library(
             name: "ZXingCppWrapper",
-            type: .static,
```
benjohnde, line 12: Short remark: this should lead to either generate it statically or dynamically depending on the target, cf. [library(name:type:targets:)](https://developer.apple.com/documentation/packagedescription/product/library(name:type:targets:)).

benjohnde:
You may have to add the framework to `Embed framework` section in Xcode, depending on the target and how the dependency gets compiled. I just tried it out! Feel free to give it another round of review, feedback and / or test it. Thanks for all the feedback @axxel @alexmanzer @parallaxe!

axxel:
Apart from the still failing CI build (see error message about the SPM version) I don't have any objections anymore. I'm very happy to see this constructive discussion from all the Apple "stakeholders". I'd be interested to hear from @parallaxe / @alexmanzer that the latest changes are in-fact addressing their remaining concerns.

@benjohnde I'd probably prefer to "squash" your commits to get rid of a bit of cleanup/revert noise.

alexmanzer:
> As the unsafeFlag issue existed, I created a new XCFramework with the Wrapper files which then umbrella'd the other one. So I resulted in ZXingCppWrapper.xcframework which contained ZXingCpp.xcframework. So maybe that was totally on me and not what you guys had in mind.

Yeah, ok, that than is an umbrella framework. But as mentioned before, that was not needed at all. 

> You may have to add the framework to `Embed framework` section in Xcode, depending on the target and how the dependency gets compiled. I just tried it out! Feel free to give it another round of review, feedback and / or test it. Thanks for all the feedback @axxel @alexmanzer @parallaxe!

I will play around with it again later today and try it out.
What is not quite clear to me yet: At which point I then define whether the lib is then built statically or dynamically when this info is flown out of the Package.swift. But I will find out then. 🙂

axxel:
Also, since @benjohnde has not commented on #623 yet, please give this a quick thought, just in case the mentioned podspec question has any influence on the decisions made here.

alexmanzer:
> Also, since @benjohnde has not commented on #623 yet, please give this a quick thought, just in case the mentioned podspec question has any influence on the decisions made here.

Hm, overall it's a question of how we/you want to proceed with iOS integration. 
From my perspective, Cocoapods was the most widespread dependency manager in the iOS world for over 10 years. We also still have it in use on some older projects as well.
However, it must be said that the entire iOS developer world is currently moving towards Swift Package Manager (SPM), as a native package manager simply makes many things easier/better. Cocoapods is a monster in terms of functionality and complexity under the hood. 
However, CocoaPods is not yet dead and SPM still has to come as far with some features as CocoaPods has come in the last 10 years.

Therefore, there are clearly two hearts beating in my chest:
For new projects I would always recommend SPM, but there are still enough projects out there that use CocoaPods. For them it would be nice if they could also use zxing-cpp.
So it's also a political question of saying:
* fu__ off "legacy guys", use SPM! or
* we are here for everyone, also for the older projects.

However, in the short term (= next days/weeks) I can't take the time to create a *.podspec and test through all the configurations, even though that's probably not the biggest thing in the world.

In general, CocoaPods should also have all the possibilities to build a framework from source during installation. 
(I have seen in recent years podspecs [like this one] (https://github.com/Legoless/OpenCV-Dynamic/blob/master/OpenCV-Dynamic.podspec)  where the lib/framework is also created on `pod install`).
So @benjohnde's changes should not make CocoaPod integration impossible at all.
Maybe a little more complicated.

Ultimately, however, this would also mean having to have another file (*.podspec) at root level.
And I know that this is a sensitive topic. 🫣

benjohnde:
> Apart from the still failing CI build (see error message about the SPM version) I don't have any objections anymore.
> @benjohnde I'd probably prefer to "squash" your commits to get rid of a bit of cleanup/revert noise.

Sure, squash that one! I will fix the CI in a sec.

I do totally agree on what @alexmanzer just stated, there are currently three "camps", CocoaPods, Carthage (still people are using that though) and SPM.

For CocoaPods, what was described here is a possibility to go (using the Framework in a Pods reference). I would rather still go with the same approach like the `Package.swift`: describe what needs to be built as CocoaPods is a wonderful tool for that. I did so with the old Objective-C port, cf. https://github.com/zxingify/zxingify-objc/blob/master/ZXingObjC.podspec. I can tackle that later next week for #623.

> At which point I then define whether the lib is then built statically or dynamically when this info is flown out of the Package.swift.

Depends on your usage, if your library (which uses zxing-cpp) is linked statically, then your deps also have to be linked statically (at least should be, otherwise you cannot guarantee your static library is working "statically", i.e. not relying on other libraries installed on your system). I am very open to have a discussion or little pair-programming session on that, if there are issues arising because of my proposed changes!

Lastly, if you really need that explicit `.static` param, in my thinking that should probably reside in some fork of yours than 'forcing' everybody else to use this dependency as a static build.

parallaxe:
@benjohnde Thanks for the changes, sounds good to me! Like @alexmanzer wrote, we have to figure out how to configure the static/dynamic-part at build time, but hopefully there will be a solution.

Regarding your comment on static / dynamic libraries
> Just realised, maybe you have mixed up both terms. Or I did a mistake here. You are describing the comparison between dynamic frameworks (Apple will strip code depending on the target device / platform) and static frameworks (the library contains everything to run independently on frameworks, which are available on the target side).

Phuh, it's a complex topic, and Apple has it's own share of complexity on its own. So it would be no wonder if we are talking past each other at some point. 😉 I don't know if it is *that* important at this point to get this clear, but I give it a shot by outlining my understanding of the topic:
- At this point, we should differentiate between a static / dynamic _library_ (.a / .dylib) and static / dynamic _framework_ (.framework / .xcframework). Breaking it down, a library is a binary that contains the executable code and exported symbols; a framework contains the library and bundles it with additional resources and meta informations. Frameworks make it easier to use a library, as its containing meta informations enable IDEs like Xcode to enable autocompletion & co, and you don't need to add headers / modulemaps manually.
- A library can contain the code for one architecture (the default) or multiple architectures (called universal or fat library).
- A framework may contain a universal / fat library, or a single architecture library. A xcframwork bundles multiple single architecture frameworks, where each framework is put into a directory for its specific architecture.
- When an iOS app is build for multiple architectures and uploaded to the AppStore (independently of the used libraries, frameworks, ...), the AppStore splits it into multiple apps and strips them down, for the specific platform and architecture (which can be seen in the AppStore, i.e. Screenshot 2023-10-17 at 12 17 45)
- The difference is: When building the app and linking against a static library / framework, the compiler will eliminate dead code. This includes the code of the static library. When building the app and linking against a dynamic library / framework, it will end up as-is on the customers iPhone, without eliminating the parts of the code the app does not use.

benjohnde:
@parallaxe static / dynamic framework (.framework / .xcframework)? `XCFramework` is just simply the collection of multipleFramework`. Dynamic and static can be applied to both, library and or frameworks. With the rest you've written above I do agree. A universal framework is created by using lipo to merge multiple created library files (of different archs) into one.

benjohnde:
@parallaxe maybe my knowledge is just wrong, I will re-read some articles (e.g. https://bpoplauschi.github.io/2021/10/25/Advanced-static-vs-dynamic-libraries-and-frameworks.html).

I had the assumption that dynamic should be favoured over static in apps. That was probably wrong.

parallaxe:
Dynamic frameworks are the default for apps and in general. My assumption is for this is, that this enables a faster iteration at development time (if all needed libraries are statically linked into the app, the linking step could take longer than the compilation step itself). Though, for ZXing with a binary size of <2MB per architecture, this should not be noticeable.
Besides that, I think static libraries are not as bad as its reputation.

benjohnde:
Sry I forgot to push, [19241b6](https://github.com/zxing-cpp/zxing-cpp/pull/630/commits/19241b6c4ed8e269d0d1ec4ededa05bbc8bc432c) is up. @axxel ci should hopefully turn green 😏

benjohnde:
@alexmanzer I had the same issue with `.dynamic`, the library needs to be embed. I mixed up both variants, my bad :)
@axxel Check, I will try to fix that as well

axxel:
> @axxel I hope [2f0d820](https://github.com/zxing-cpp/zxing-cpp/pull/630/commits/2f0d8203aaa7d4dd1ca1d2fdcae5379f4852879a) does some justice w.r.t. the ci job. Can I also manually run the test?

Seems to have had no effect. I thought you can simply manually trigger a ci job on the Actions page, but I could not find it now. hm. Maybe you need to somehow activate GA on your fork and then any push from you triggers a ci run?

alexmanzer:
@benjohnde Just tried some things here to find the problem with the build:

If I execute `xcodebuild build -scheme demo -sdk "iphonesimulator"` on my machine with Xcode15, everythings runs fine.
In one of our older build machines, that have still Xcode 14.3.1 (like the GitHub runners) I get the same problem as in the CI job:
```
/Users/ci-macos-01/tmp/zxing-cpp/wrappers/ios/demo/demo.xcodeproj: error: Missing package product 'ZXingCppWrapper' (in target 'demo' from project 'demo')
/Users/ci-macos-01/tmp/zxing-cpp/wrappers/ios/demo/demo.xcodeproj: error: Missing package product 'ZXingCppWrapper' (in target 'demo' from project 'demo')
** BUILD FAILED **
```

If I try to open the demo project with Xcode 14.3.1 it also throws this error:
image

So it is really a problem of the already updated demo project.

alexmanzer:
@benjohnde Ok, I got it working again with Xcode14.
Unfortunately I can't commit anything here, but this git patch fixes the problem:

[zxing-cpp-bc06b10-Fix demo project for Xcode 14 again.patch](https://github.com/zxing-cpp/zxing-cpp/files/12939488/zxing-cpp-bc06b10-Fix.demo.project.for.Xcode.14.again.patch)

benjohnde:
Ah, I did a mistake. I tried to use the macos-13 runner but forgot to update to Xcode 15. The provided patch solves both, Xcode 14 and Xcode 15 usage. Guess we are good to fuel that into the master ⛽ 🚀 !

alexmanzer:
> The provided patch solves both, Xcode 14 and Xcode 15 usage. Guess we are good to fuel that into the master ⛽ 🚀 !

🚀🚀🚀

So, thanks to all your work @benjohnde together with some other iOS improvements (e.g. also this open [pull request](https://github.com/zxing-cpp/zxing-cpp/pull/635)) the iOS usage of zxing-cpp really becomes very powerful and I think there is no reason to use the old zxingify-objc anymore. 💪🏻

benjohnde:
@parallaxe I have to admit that my knowledge was wrong. As far as I understood it, dynamic frameworks are useful for development (if linking statically is time consuming) and environments, in which you want to change the underlying framework without compromising the target (i.e. in which the framework is being used) - like hot-swapping frameworks 😏 without re-deploying the whole application. It could also be useful, if the downside of linking statically would result in a dramatic increased app size. Apart from that, it seems that static is more useful in an app context.

axxel:
Thanks to all of you for this pretty intense collaboration! This is the first time two parties have been working together on something in this code base, where I was neither of those two parties. :)

alexmanzer:
@axxel Thanks for merging into the master! 🙏

Merely for the sake of interest: When do you usually set a new version tag to this repo?
Because with this PR (and a new version tag) it is then actually finally possible to use the versioning system of SPM as it is intended. 

(It would be cool to include the change from the other [pull request](https://github.com/zxing-cpp/zxing-cpp/pull/635), that would be really helpful for us.)

axxel:
> When do you usually set a new version tag to this repo?

Usually right before the next release. I understand that this might cause confusion when just grabbing HEAD and tag it with the latest release number even if the code changed. But even if I started introducing some `-alpha` like naming scheme applied directly after a release, it would still not be the best idea to always simply grab HEAD when you build your app, right? So I would imagine there must be a way to specify in your SPM config which tag you'd like to download?

alexmanzer:
@axxel When importing an SPM package into Xcode the package manage will care about the versions, **but only** if the repo has "semantic version tag", like 2.1.0.

If you add a package in Xcode, you get this dialog:
Screenshot 2023-10-18 at 15 26 30

And the options for "Dependency Rules" are:

Screenshot 2023-10-18 at 15 26 39

So, Xcode SPM can only care about updates, when there is a version tag.
Sure, there is possibility to select a branch or a commit, but it's always better to reference to a defined version IMO.

And the fact that it's only possible with this PR here that you can even use the SPM like this would of course strengthen a new version tag. But of course, this mainly affects the iOS community
 But since maybe lots of zxingify-objc users will end up here soon, it might be worth it. 🙂

Issue #614, opened 2023-09-08
QR Code Model 1 not supported


sayzzy-nt:
First of all, thank you for providing such a nice library.
This QR code is not recognized, but I don't know what the problem is.

The smartphone app recognized it, so I don't think the QR code is broken.

![errorQR1](https://github.com/zxing-cpp/zxing-cpp/assets/136289338/91f2cb7f-0bdd-4722-8f5c-1b69e9becace)

axxel:
I do believe that this symbol is in fact invalid. It is correctly detected but could not be decoded due to a checksum error. I checked the sample grid if there was an issue there but the grid placement is perfectly fine. I also tried to decode this image with a couple different online scanners and all agree that there is no valid qrcode to be found.

Which "smartphone app" were you using and what was the content of the detected qrcode?

sayzzy-nt:
The app I was using was "QRQR"
And the string read is "766400488514071012112935"

I have attached the app URL below.
https://www.denso-wave.com/en/system/qr/product/reader.html

axxel:
Thanks for the information. I found out what the issue is: the image you provided contains an (old/outdated) QR Code Model 1 (as opposed to the Model 2). According to the ISO specification, "Model 1 of QR Code, as defined in AIM ITS 97-001, is the form of the symbology originally used for a number of early or closed systems applications but is not recommended for use in new or open systems applications [...]". Model 1 codes are simply not implemented. If you need support for those and have some money to throw at the problem, please get in touch with me via zxingcpp (at) gmail.com.

sayzzy-nt:
thank you for your reply.

I'm sorry if my understanding is wrong.
I believe that the difference between Model 1 and Model 2 of QR codes is the presence or absence of alignment patterns.
However, this library may be able to read her QR code without alignment pattern.
I don't know what the difference is.
I hope you'll tell me.

I have pasted the QR code read by "zxing" below.

![2023-09-11 193837](https://github.com/zxing-cpp/zxing-cpp/assets/136289338/65b5443b-3633-4ae1-a8c5-264edbf89448)

axxel:
That is not entirely correct. Model 1 has no alignment patterns in any of the available sizes (or versions in QR speak). Model 2 Version 1 has no alignment patterns but higher versions have them. The most relevant difference is (quote) "symbol character placement follows different rules", i.e. the way the black/white squares are to be interpreted is different.

sayzzy-nt:
[...]

PR #633, opened 2023-10-17
Support QR Code Model1

sayzzy-nt:
- #614 

Supported QR code model 1
Although the source code is not very clean, you should be able to read all QR code model 1 with this.


core/src/qrcode/QRDecoder.cpp
```diff
@@ -316,17 +321,28 @@ DecoderResult DecodeBitStream(ByteArray&& bytes, const Version& version, ErrorCo
...
 		return FormatError("Invalid version");
-	const Version& version = *pversion;
+	const Version& preversion = *pversion;
 
-	auto formatInfo = ReadFormatInformation(bits, version.isMicroQRCode());
+	auto formatInfo = ReadFormatInformation(bits, preversion.isMicroQRCode(), isModel1);
 	if (!formatInfo.isValid())
 		return FormatError("Invalid format information");
 
+	if (isModel1)
```
axxel, line 335: This split between reading version and reading format info is obviously not suitable anymore when those two inter-depend in the presence of Model1. I'll think about how I'd like this to be cleanup after merging.

core/src/qrcode/QRBitMatrixParser.h
```diff
@@ -20,20 +20,20 @@ class FormatInformation;
...
  */
-FormatInformation ReadFormatInformation(const BitMatrix& bitMatrix, bool isMicro);
+FormatInformation ReadFormatInformation(const BitMatrix& bitMatrix, bool isMicro, bool& isModel1);
 
 /**
  * @brief Reads the codewords from the BitMatrix.
  * @return bytes encoded within the QR Code or empty array if the exact number of bytes expected is not read
  */
-ByteArray ReadCodewords(const BitMatrix& bitMatrix, const Version& version, const FormatInformation& formatInfo);
+ByteArray ReadCodewords(const BitMatrix& bitMatrix, const Version& version, const FormatInformation& formatInfo , bool isModel1);
```
axxel, line 36: `ReadCodewords` can then get the `isModel1` info from the `formatInfo`.

core/src/qrcode/QRVersion.cpp
```diff
@@ -304,23 +399,35 @@ Version::Version(int versionNumber, const std::array<ECBlocks, 4>& ecBlocks)
...
+const Version* Version::FromNumber(int versionNumber, bool isMicro, bool isModel1)
 {
 	if (versionNumber < 1 || versionNumber > (isMicro ? 4 : 40)) {
 		//throw std::invalid_argument("Version should be in range [1-40].");
 		return nullptr;
 	}
-	return &(isMicro ? AllMicroVersions() : AllVersions())[versionNumber - 1];
+
+	if (isMicro)
+	{
```
axxel, line 410: Please follow the `{` placement style of the rest of the file.

axxel (changes requested):
Thanks a lot for your cooperation. Please also add the following test-image as `test/samples/qrcode-2/qr-model-1.png` and update `BlackboxTestRunner.cpp` accordingly.
qr-model1

axxel (approved):

axxel:
As you can see, I thought the build failure can be fixed easily via the web-gui of github but it took me 4 tries to get everything. sigh. Now it builds but one test fails. If could have look, that would be very nice. If not, I will then have to properly download your PR locally and debug it.

Also, please replace the png you added with the one, I provided earlier. I made an effort to make this file as small as possible (230 bytes vs 3.4kB). The test images use up way to much space and I try to limit the effect of any new addition.

PR #635, opened 2023-10-17
iOS wrapper: Add functionality of encoding binary data into Barcodes

alexmanzer:
This is the iOS addition for the previous pull request from @markusfisch (https://github.com/zxing-cpp/zxing-cpp/pull/628) to be able to use the functionality of encoding binary data into Barcodes like DataMatrix.

Usage from swift will then be like:
```
let rawData = Data(base64Encoded: "VEVTVDEyMzQ=")!

let zxWriter = ZXIBarcodeWriter()
let cgImage = try? zxWriter.writeBytes(rawData, width: 500, height: 500, format: .DATA_MATRIX)
```


wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
```diff
@@ -14,6 +14,13 @@ NS_ASSUME_NONNULL_BEGIN
                      height:(int)height
                      format:(ZXIFormat)format
                       error:(NSError **)error;
+
+-(nullable CGImageRef)writeData:(NSData *)data
```
axxel, line 18: I'm not really happy with the name `writeData`. First of all: am I right in assuming that ObjectiveC does not support function overloading (based on the type of parameter), so this function has to have a different name than the existing one? The first best alternative that comes to my mind is `writeBytes` (see `class Result`). It would then make a lot of sens to rename the other to `writeText`.

The next question would be: how disruptive would it be to later change the name / parameter list of this function. I'm asking because I'm working on a new `WriteBarcode` API that is supposed to be analogous to the `ReadBarcode` API, meaning it would get passed an `EncodeHints` struct, similar to `DecodeHints`. Introducing something new now that effectively prohibits me from doing 'the right thing' later is what I'd like to avoid.

wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.mm
```diff
@@ -4,10 +4,10 @@
 
 #import <CoreGraphics/CoreGraphics.h>
 #import "ZXIBarcodeWriter.h"
-#import "ZXing/MultiFormatWriter.h"
-#import "ZXing/BitMatrix.h"
 #import "ZXIFormatHelper.h"
 #import "ZXIErrors.h"
+#import "ZXing/MultiFormatWriter.h"
+#import "ZXing/BitMatrix.h"
```
axxel, line 10: This is causing a conflict with the just merged PR. Please fix.

wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.mm
```diff
@@ -18,8 +18,20 @@
...
+std::wstring NSDataToStringW(NSData *data) {
+    std::wstring s;
+    const unsigned char *bytes = (const unsigned char *) [data bytes];
+    size_t len = [data length];
+    wchar_t buf[len];
+    for (int i = 0; i < len; ++i) {
+        buf[i] = bytes[i];
+    }
+    s.assign(buf, len);
+    return s;
```
axxel, line 30: Minor cosmetic: a somewhat cleaner version without the `buf` would be along the lines of:
```
			std::wstring s;
			for (int i = 0; i < len; ++i)
				s.push_back(bytes[i]);
```

wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.mm
```diff
@@ -18,8 +18,20 @@
...
+}
+
 #ifdef  DEBUG
-std::string ToString(const BitMatrix& matrix, char one, char zero, bool addSpace, bool printAsCString)
+std::string ToString(BitMatrix& matrix, char one, char zero, bool addSpace, bool printAsCString)
```
axxel, line 34: Why did you remove the `const` qualifier?

axxel (changes requested):
Thanks.

alexmanzer:
@axxel From my point of view, everything is now done.
If you agree, I would suggest a squash commit to avoid having all the fedback rounds and corrections on the master. But I'm almost sure that's what you would have done anyway.

axxel:
@benjohnde are you using the Writer-API of the iOS Wrapper? If so, am I right that you would find the amount of work caused by the `write` -> `writeText` renaming acceptable? ;)

benjohnde:
@axxel only using read :)

Issue #204, opened 2021-03-13
Releasing python wrapper on pypi


antoinehumbert:
Hi,

I'd like to know if you have any plan for officially releasing wheels of the zxing-cpp python wrapper on pypi.org ?

I can see that wrapper is considered experimental in the readme file. What is missing to consider it stable for official release ? How can one help / contribute in order to achieve the goal of official releases (if this is in your plan) ?

Thanks for your answer

axxel:
Thanks for your offer to help improving the project. There is currently no plan to officially release wheels on pypi.org. That said, I would personally welcome that. I don't have any experience in doing that and it would not happen without help from the community.

I'd consider the wrapper experimental mostly because I don't see the current API being 'future proof'. I'd generally prefer if the different wrappers and the main-line C++ API were consistent and ideally equally capable. But it also makes sense to me to use an API that feels natural/native in the respective language. The read function for example has a set of parameters that naturally split into 2 sets: a) the image (the data and its layout/format) and b) the (optional) decoding parameters. The latter is typically modeled as an object in Android, which is why we have this `DecodingHints` class here and I consider it a good approach compared to a single function with a ton of parameters.

The current python wrapper goes the single parameter route though, but is currently limited to the most basic set of 'hints'. The question would be how to go about that in a pythonic way. Optional named parameters or a dictionary? Suggestions?

Looking forward to your feedback.

antoinehumbert:
Well, I see what you mean. May be it would be better for the zxing-cpp project to have wrappers as close as possible as the c++ API in order to make it easier to maintain when something changes in the C++ code.

However, providing APIs that are more "language friendly" is also probably better for adoption by "final" users.

So, there are probably two API levels to propose : the "C++ main-line" one and the "pythonic" one. The first being in the wrappers of zxing-cpp. The latter may be in the same project, or another one : I'm in doubt about that.

I'll try to make some proposals for that in a pull request. Just one question before starting some wrok : are you attached to the use of pybind11 for the binding, or would it be acceptabe to use cython ? I know the first one is usually prefered by cpp developpers and the latter by python developpers.

axxel:
I'd prefer the following:
1) have one python module/API that feels "pythonic" and supports all current features of the `ReadBarcode` API
2) stick to pybind11
3) anticipate the following future improvements:
  * more options controlling the detection
  * return results for multiple symbols from one call
  * extend the `Result` with more information (like e.g. a `flipped/mirrored` flag)

[...]

PR #637, opened 2023-10-18
iOS: Enable CocoaPods by providing a podspec

benjohnde:
## General

I did not use the provided modulemap as Swift PM works differently than CocoaPods. To reuse that, I had to change the definition `module ZXingCppWrapper` to `framework module ZXingCppWrapper` which then caused a lot of changes and in the end broke the swift package. I decided to go with the pods-generated modulemap (which is enabled by default) and provided the public headers in the spec.

This solves #623.

Feedback is very welcome @AbijahKaj @alexmanzer @parallaxe @axxel!

## Release
For a potential release I would introduce a proper github-action in this PR (probably something like https://github.com/marketplace/actions/cocoapods-action).

As far as I am aware, the release has to happen before the spec can be validated as normally the spec accesses the version via tag, maybe there is a different variant possible.

## Testing

To test that via:
```shell
pod spec lint --verbose --allow-warnings
```

Warnings should be ignored, they have nothing to do with the actual podspec.

One need to replace the source with a local repo, e.g.:
```ruby
s.source = {
  :git => 'file:///Users/{whoami}/git/zxing-cpp'
}
```

### DIFF
--- a/zxing-cpp.podspec
+++ b/zxing-cpp.podspec
@@ -12,7 +12,7 @@ Pod::Spec.new do |s|
     :git => 'https://github.com/zxing-cpp/zxing-cpp.git',
     :tag => "v#{s.version}"
   }
-  s.module_name = 'ZXingCppWrapper'
+  s.module_name = 'ZXingCpp'
   s.platform = :ios, '11.0'
   s.frameworks = 'CoreGraphics', 'CoreImage', 'CoreVideo'
   s.library = ['c++']
--- a/zxing-cpp.podspec
+++ b/zxing-cpp.podspec
@@ -1,0 +1,29 @@
+Pod::Spec.new do |s|
+  s.name = 'zxing-cpp'
+  s.version = '2.1.0'
+  s.summary = 'C++ port of ZXing'
+  s.homepage = 'https://github.com/zxing-cpp/zxing-cpp'
+  s.author = 'axxel'
+  s.license = {
+    :type => 'Apache License 2.0',
+    :file => 'LICENSE'
+  }
+  s.source = {
+    :git => 'https://github.com/zxing-cpp/zxing-cpp.git',
+    :tag => "v#{s.version}"
+  }
+  s.module_name = 'ZXingCppWrapper'
+  s.platform = :ios, '11.0'
+  s.frameworks = 'CoreGraphics', 'CoreImage', 'CoreVideo'
+  s.library = ['c++']
+  s.pod_target_xcconfig = {
+    'CLANG_CXX_LANGUAGE_STANDARD' => 'c++20'
+  }
+  s.source_files = 'core/src/**/*.{h,c,cpp}',
+                   'wrappers/ios/Sources/Wrapper/**/*.{h,m,mm}'
+  s.public_header_files = 'wrappers/ios/Sources/Wrapper/Reader/{ZXIBarcodeReader,ZXIResult,ZXIPosition,ZXIPoint,ZXIDecodeHints}.h',
+                          'wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h',
+                          'wrappers/ios/Sources/Wrapper/{ZXIErrors,ZXIFormat}.h'
+  s.private_header_files = 'core/src/**/*.h'
+  s.exclude_files = 'wrappers/ios/Sources/Wrapper/UmbrellaHeader.h'
+end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
       with:
         ref: ${{github.ref}}
 
-    - name: Build the swift package
+    - name: Build the swift package //  pod lib lint --allow-warnings
       shell: sh
       working-directory: ${{runner.workspace}}/${{github.event.repository.name}}
       run: swift build
@@ -89,6 +89,11 @@ jobs:
       shell: sh
       working-directory: ${{runner.workspace}}/${{github.event.repository.name}}/wrappers/ios/demo
       run: xcodebuild build -scheme demo -sdk "iphonesimulator"
+
+    - name: Validate the Pod
+      shell: sh
+      working-directory: ${{runner.workspace}}/${{github.event.repository.name}}
+      run: pod lib lint --allow-warnings
 
   build-android:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
       with:
         ref: ${{github.ref}}
 
-    - name: Build the swift package //  pod lib lint --allow-warnings
+    - name: Build the swift package
       shell: sh
       working-directory: ${{runner.workspace}}/${{github.event.repository.name}}
       run: swift build


benjohnde:
> Hi @benjohnde !
> 
> Thanks for this addition!
> 
> I tested by adding it in a fresh iOS project it works fine. But I stil have trouble using it in a podspec as subspec (as there's no published version yet).
> 
> Having a published version on cocoapods will help a lot.

That's actually good news :) Then I have to find out if there exists a better way to lint the spec (lint does actually more than just linting, it validates the spec by compiling the Pod, creating a demo project and integrating the pod). I couldn't find a meaningful way to lint the spec (for some pre-release check) against my local host without replacing the `source`.

benjohnde:
Done, CI should turn green in the end 🤞 !

@axxel it is still some decision whether you want to release the CocoaPod integration officially.

If so, after an official release one would need to adjust the `s.version = '2.1.0'` (according to the new version) and release the Pod by running `pod trunk push`, cf. https://guides.cocoapods.org/making/getting-setup-with-trunk.html. Then it gets published on CocoaPods.org and is officially available.

I am not super familiar with github-actions but there should be a way to automate that after a new version was tagged / published - somehow we should think about that for published android versions as well.

Nonetheless, this Pod is working and releasable.

axxel:
Thanks for the continued effort. I will merge this but have a couple of open questions:

1) So the `s.version` string has to be kept in sync manually on every release?
2) I'm not sure I understood the validation situation: the file is 'valid' as is or we'll only know after the first release and if it is not, we'll then need to fix it and make another release (like 2.3)?
3) I'd like to bring up the naming subject again, now that a new namespace (cocoapods.org) is entered. I'm talking about the `ZXingCpp` vs `zxingcpp` vs `zxing-cpp` vs `ZXingCppWrapper` situation. For reference, please checkout [point 7 here](https://github.com/zxing-cpp/zxing-cpp/pull/337#issuecomment-1136363104) and the followup discussion as well as [point 1 here](https://github.com/zxing-cpp/zxing-cpp/issues/204#issuecomment-821893053). If it were not for the limitation of not allowing '-' in some contexts, I'd prefer if it was called `zxing-cpp` (most of the time). Bottom line: we have some rather incoherent naming mess right now and I'd like to not make that even more random. If I compare the cocoapods.org situation with pypi.org, then calling the entity `zxing-cpp` even though the thing you need to later include/add/link is called `zxingcpp` (Python) / `ZXingCppWrapper` (CocoaPods) has some consistency to it. I imagine on the Android side the same question will show up when looking for a name of the thing on MavenCentral, etc.? The Android module itself is called `zxingcpp` again.
That also makes me wonder whether the term `ZXingCppWrapper` is actually still required, now that the core code and wrapper code ended up in one single framework? Why not simply call it `ZXingCpp` before potentially a lot more people start using it?

As the part of the naming question is affecting all iOS users, let me ping the rest of the bunch again: @alexmanzer @parallaxe @ryantrem

zxing-cpp.podspec
```diff
@@ -0,0 +1,29 @@
...
+  s.pod_target_xcconfig = {
+    'CLANG_CXX_LANGUAGE_STANDARD' => 'c++20'
+  }
+  s.source_files = 'core/src/**/*.{h,c,cpp}',
+                   'wrappers/ios/Sources/Wrapper/**/*.{h,m,mm}'
+  s.public_header_files = 'wrappers/ios/Sources/Wrapper/Reader/{ZXIBarcodeReader,ZXIResult,ZXIPosition,ZXIPoint,ZXIDecodeHints}.h',
+                          'wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h',
+                          'wrappers/ios/Sources/Wrapper/{ZXIErrors,ZXIFormat}.h'
+  s.private_header_files = 'core/src/**/*.h'
+  s.exclude_files = 'wrappers/ios/Sources/Wrapper/UmbrellaHeader.h'
```
axxel, line 28: Can you explain why this one header is excluded?

alexmanzer:
> Thanks for the continued effort. I will merge this but have a couple of open questions:
>    That also makes me wonder whether the term `ZXingCppWrapper` is actually still required, now that the core code and wrapper code ended up in one single framework? Why not simply call it `ZXingCpp` before potentially a lot more people start using it?

@axxel  I'm right there with you. 👍🏻
This has always disturbed my inner Monk a bit, because as an end user of the lib I want just to use ZxingCpp and be able to write `import ZxingCpp` in my sources instead of caring about an "Wrapper". 
It's not a big change, but in my opinion it's an important one in terms of consistency.

I have created a separate PR, which is independent of this CocoaPods topic here:
https://github.com/zxing-cpp/zxing-cpp/pull/638

benjohnde:
Absolutely d'accord with renaming it to ZxingCpp!!!

zxing-cpp.podspec
benjohnde: 1. Regarding `s.version`: yes!

benjohnde:
> 1. So the `s.version` string has to be kept in sync manually on every release?

Yes, sadly! Not sure if there exists another way of dealing with that honestly!

> 2. I'm not sure I understood the validation situation: the file is 'valid' as is or we'll only know after the first release and if it is not, we'll then need to fix it and make another release (like 2.3)?

axxel:
I just merged the renaming, so the podspec needs a final update and we are good to go, right?

benjohnde:
@axxel rename is pushed

ryantrem:
> Thanks for the continued effort. I will merge this but have a couple of open questions:
> 
> 1. So the `s.version` string has to be kept in sync manually on every release?
> 2. I'm not sure I understood the validation situation: the file is 'valid' as is or we'll only know after the first release and if it is not, we'll then need to fix it and make another release (like 2.3)?
> 3. I'd like to bring up the naming subject again, now that a new namespace (cocoapods.org) is entered. I'm talking about the `ZXingCpp` vs `zxingcpp` vs `zxing-cpp` vs `ZXingCppWrapper` situation. For reference, please checkout [point 7 here](https://github.com/zxing-cpp/zxing-cpp/pull/337#issuecomment-1136363104) and the followup discussion as well as [point 1 here](https://github.com/zxing-cpp/zxing-cpp/issues/204#issuecomment-821893053). If it were not for the limitation of not allowing '-' in some contexts, I'd prefer if it was called `zxing-cpp` (most of the time). Bottom line: we have some rather incoherent naming mess right now and I'd like to not make that even more random. If I compare the cocoapods.org situation with pypi.org, then calling the entity `zxing-cpp` even though the thing you need to later include/add/link is called `zxingcpp` (Python) / `ZXingCppWrapper` (CocoaPods) has some consistency to it. I imagine on the Android side the same question will show up when looking for a name of the thing on MavenCentral, etc.? The Android module itself is called `zxingcpp` again.
>    That also makes me wonder whether the term `ZXingCppWrapper` is actually still required, now that the core code and wrapper code ended up in one single framework? Why not simply call it `ZXingCpp` before potentially a lot more people start using it?
> 
> As the last part of the naming question is affecting all iOS users, let me ping the rest of the bunch again: @alexmanzer @parallaxe @ryantrem

I see this was already merged, but I'll add some thoughts regardless. One question for me is whether it is more ideal to have a single podspec, or one per "layer." For example, it's really important at a CMakeLists.txt level to have core separate from specific language bindings ("wrappers"). This makes it easy to use the core library in C++ code that is compiled for different platforms (e.g. Android, iOS, etc.) but doesn't use platform specific languages (e.g. Java, Objective-C, etc.). I wonder a little bit if we'd want the same thing here, where there is a podspec for language bindings that depends on a podspec for core. If a consumer can depend on a single podspec that has both core and language bindings, but can safely ignore the language bindings, then maybe the podspec situation is fundamentally different from the CMakeLists.txt situation, and a single podspec is ok.

benjohnde:
> I see this was already merged, but I'll add some thoughts regardless. One question for me is whether it is more ideal to have a single podspec, or one per "layer." For example, it's really important at a CMakeLists.txt level to have core separate from specific language bindings ("wrappers"). This makes it easy to use the core library in C++ code that is compiled for different platforms (e.g. Android, iOS, etc.) but doesn't use platform specific languages (e.g. Java, Objective-C, etc.). I wonder a little bit if we'd want the same thing here, where there is a podspec for language bindings that depends on a podspec for core. If a consumer can depend on a single podspec that has both core and language bindings, but can safely ignore the language bindings, then maybe the podspec situation is fundamentally different from the CMakeLists.txt situation, and a single podspec is ok.

Love these thoughts, I initially created two subspecs, Core & Wrapper. Then I skipped that approach as I weren't sure if one would want the Core without the Obj-C wrapper code. I can provide a recommendation @ryantrem if that's of interest for the react-native realm!

axxel:
Preparing the 2.3.0 release, I came by the cocoapods.org page again and was surprised that searching for neither `zxing` nor `zxing-cpp` revealed anything about out project. Does anyone know what is going on? I would have hoped the [project page](https://cocoapods.org/pods/zxing-cpp) would be discoverable via search.

PR #638, opened 2023-10-19
Rename ZXingCppWrapper to ZXingCpp

alexmanzer:
As discussed [here](https://github.com/zxing-cpp/zxing-cpp/pull/637#issuecomment-1769255088) this is IMO more intuitive way to name the package.


benjohnde:
Love it!

PR #641, opened 2023-10-19
Split Podspec into subspecs

benjohnde:
@ryantrem following your remark in #637 this is a first draft of a potential Pod with subspecs.

One could either use:

```
pod 'zxing-cpp/Core'
```

or

```
pod 'zxing-cpp'
```

which is basically an alias due to the declared dependency for

```
pod 'zxing-cpp/Wrapper'
```

### DIFF
--- a/wrappers/ios/README.md
+++ b/wrappers/ios/README.md
@@ -1,3 +1,27 @@
 # ZXingCpp iOS Framework
 
-Add the iOS wrapper either as a local Swift Package by adding it as a dependency to your app or add the repository, the Package.swift file is automatically picked.
+## Usage
+
+For general usage, please compare the source code provided in the demo project.
+
+### Swift PM
+
+As this repository provides a `Package.swift` on root level, you can add `zxing-cpp` including wrapper code by adding a Package Dependency in Xcode.
+
+An alternative way is to check this repository out and add it as a local Swift Package by adding it as dependency to your app.
+
+### CocoaPods
+
+As an alternative way, you can also rely on [CocoaPods](https://cocoapods.org/pods/zxing-cpp). Just add the Pod as dependency, for instance:
+
+```
+pod 'zxing-cpp'
+```
+
+If you just need the core without the wrapper code, you can also rely on:
+
+```
+pod 'zxing-cpp/Core'
+```
+
+The module to be imported is named `ZXingCpp`.
--- a/zxing-cpp.podspec
+++ b/zxing-cpp.podspec
@@ -4,6 +4,7 @@ Pod::Spec.new do |s|
   s.summary = 'C++ port of ZXing'
   s.homepage = 'https://github.com/zxing-cpp/zxing-cpp'
   s.author = 'axxel'
+  s.readme = 'https://raw.githubusercontent.com/zxing-cpp/zxing-cpp/master/wrappers/ios/README.md'
   s.license = {
     :type => 'Apache License 2.0',
     :file => 'LICENSE'
--- a/zxing-cpp.podspec
+++ b/zxing-cpp.podspec
@@ -19,6 +19,8 @@ Pod::Spec.new do |s|
     'CLANG_CXX_LANGUAGE_STANDARD' => 'c++20'
   }
 
+  s.default_subspec = 'Wrapper'
+
   s.subspec 'Core' do |ss|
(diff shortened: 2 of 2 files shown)


benjohnde:
@axxel before this Pod gets published we could also rename the file to `ZXingCpp.podspec` which is probably more sound with the Swift / iOS naming but in the end more a personal preference.

Do you want the Pod to be published as `zxing-cpp` or `ZXingCpp` on CocoaPods? Both is possible, your call! :)

ryantrem (approved):

axxel:
> which is basically an alias due to the declared dependency for

Interesting, I don't see any dependency of the root entity on the `Wrapper`. Is `pod zxing-cpp` simply pulling all subspecs?

> Do you want the Pod to be published as `zxing-cpp` or `ZXingCpp` on CocoaPods?

In my [comment](https://github.com/zxing-cpp/zxing-cpp/pull/637#issuecomment-1769255088) I argued that there would be some consistency if we used the term `zxing-cpp` (see the PyPi situation). I also thought that from an iOS developers perspective who might use SPM (via the github url) and/or CocoaPods (via the `pod` command), it makes sense that both access paths use the same term (`zxing-cpp`). On the other hand, if the CocoaPods namespace obviously settled for PascalCase and the `s.module_name` is `ZXingCpp` then maybe that internal consistency trumps the other?

Lets vote? (@benjohnde @ryantrem @alexmanzer @parallaxe): 
👍 = `zxing-cpp`
👎 = `ZXingCpp`

benjohnde:
@axxel good catch! `s.default_subspec = 'Wrapper'` was missing, my bad! Now my initial post is valid. That's the way CocoaPods recommends to do it: https://guides.cocoapods.org/syntax/podspec.html#default_subspecs

> A Pod should make available the full library by default. Users can fine tune their dependencies, and exclude unneeded subspecs, once their requirements are known.

CocoaPods uses the name of the spec which should be the same as the filename as module name. That's why I added the module name as it is different.

The main obstacle to talk about is how the module should be named on CocoaPods, I don't really have an opinion. A weak statement would be that `zxing-cpp` is probably the better choice as the name of this project and repo is the same.

alexmanzer:
> Lets vote? (@benjohnde @ryantrem @alexmanzer @parallaxe): 👍 = `zxing-cpp` 👎 = `ZXingCpp`

Personally I really like `zxing-cpp` because it's more consistent to this whole repo/project.
But in the iOS world, there is unfortunately another convention regarding the naming of frameworks/libs, so for iOS I would stick with `ZXingCpp`, as you can see here:
image

benjohnde:
Hey @alexmanzer, as it is committed as of now, the Pod name would be `zxing-cpp` but the module is named `ZXingCpp`.

So you would require the Pod `zxing-cpp` but import `ZXingCpp` as module. Not quite sure whether this makes a difference in your decision but just to paint the picture in a complete way.

axxel:
With 3 votes for `zxing-cpp` as the CocoaPods name that seems settled. I understand that this is ready for merging, right?

@benjohnde maybe you can give one final look at the multitude of places where a name shows up and make sure they are not somehow contradicting each other: `zxing-cpp.podspec`: `s.name`, `s.module_name`, `Package.swift`: Package-, Library-, Taget-name, `demo.xcodeproj/*`

benjohnde:
Awesome!

@axxel I will look into that after the weekend to the latest and also provide a meaningful README section, which is subject to be reviewed by @alexmanzer :)

Do you want to have this placed in the root README or should I use the wrappers/ios/README?

benjohnde:
@axxel I can point to a subfolder for the Pod readme btw https://guides.cocoapods.org/syntax/podspec.html#readme depends on your flavour!

benjohnde:
@axxel done. Next steps are:
- merge
- prepare new release (i.e. adjust version number and maybe tag of source in the zxing-cpp.podspec file)
- create new release
- Then I will push the Spec to CocoaPods
- that's it

gitlost:
Thanks, that works for me.

axxel:
> Thanks, that works for me.

? Was that comment supposed to go somewhere else?

gitlost:
Ha yeah, sorry, had repo wrong, both this and other one.

PR #644, opened 2023-10-24
ios: specify EC level/margin for generation

markusfisch:
So we can specify a certain error correction level (and margin).

### DIFF
--- a/wrappers/ios/Sources/Wrapper/UmbrellaHeader.h
+++ b/wrappers/ios/Sources/Wrapper/UmbrellaHeader.h
@@ -10,6 +10,7 @@
 #import "Reader/ZXIPosition.h"
 #import "Reader/ZXIPoint.h"
 #import "Reader/ZXIDecodeHints.h"
+#import "Writer/ZXIEncodeHints.h"
 #import "Writer/ZXIBarcodeWriter.h"
 #import "ZXIErrors.h"
 #import "ZXIFormat.h"
--- a/wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
+++ b/wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
@@ -36,9 +36,21 @@ const int PDF417_ERROR_CORRECTION_8 = 8;
                            format:(ZXIFormat)format
                             width:(int)width
                            height:(int)height
+                            error:(NSError *__autoreleasing  _Nullable *)error;
+
+-(nullable CGImageRef)writeString:(NSString *)contents
+                           format:(ZXIFormat)format
+                            width:(int)width
+                           height:(int)height
                            margin:(int)margin
                           ecLevel:(int)ecLevel
                             error:(NSError *__autoreleasing  _Nullable *)error;
+
+-(nullable CGImageRef)writeData:(NSData *)data
+                         format:(ZXIFormat)format
+                          width:(int)width
+                         height:(int)height
+                          error:(NSError *__autoreleasing  _Nullable *)error;
 
 -(nullable CGImageRef)writeData:(NSData *)data
                          format:(ZXIFormat)format
--- a/wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
+++ b/wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.h
@@ -7,18 +7,45 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+const int AZTEC_ERROR_CORRECTION_0 = 0;
+const int AZTEC_ERROR_CORRECTION_12 = 1;
+const int AZTEC_ERROR_CORRECTION_25 = 2;
+const int AZTEC_ERROR_CORRECTION_37 = 3;
+const int AZTEC_ERROR_CORRECTION_50 = 4;
+const int AZTEC_ERROR_CORRECTION_62 = 5;
+const int AZTEC_ERROR_CORRECTION_75 = 6;
+const int AZTEC_ERROR_CORRECTION_87 = 7;
+const int AZTEC_ERROR_CORRECTION_100 = 8;
+const int QR_ERROR_CORRECTION_LOW = 2;
+const int QR_ERROR_CORRECTION_MEDIUM = 4;
+const int QR_ERROR_CORRECTION_QUARTILE = 6;
+const int QR_ERROR_CORRECTION_HIGH = 8;
+const int PDF417_ERROR_CORRECTION_0 = 0;
+const int PDF417_ERROR_CORRECTION_1 = 1;
+const int PDF417_ERROR_CORRECTION_2 = 2;
+const int PDF417_ERROR_CORRECTION_3 = 3;
+const int PDF417_ERROR_CORRECTION_4 = 4;
+const int PDF417_ERROR_CORRECTION_5 = 5;
+const int PDF417_ERROR_CORRECTION_6 = 6;
+const int PDF417_ERROR_CORRECTION_7 = 7;
+const int PDF417_ERROR_CORRECTION_8 = 8;
+
 @interface ZXIBarcodeWriter : NSObject
 
 -(nullable CGImageRef)writeString:(NSString *)contents
+                           format:(ZXIFormat)format
                             width:(int)width
                            height:(int)height
-                           format:(ZXIFormat)format
+                           margin:(int)margin
+                          ecLevel:(int)ecLevel
                             error:(NSError *__autoreleasing  _Nullable *)error;
 
 -(nullable CGImageRef)writeData:(NSData *)data
+                         format:(ZXIFormat)format
                           width:(int)width
                          height:(int)height
-                         format:(ZXIFormat)format
+                         margin:(int)margin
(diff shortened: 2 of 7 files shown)


benjohnde:
Not a big deal but you could add a wrapper function to provide defaults and I would not change the signature as this would break existing code. On the other hand, I am not sure how many people rely on the library and on these functions @alexmanzer.

markusfisch:
Good point! I was already playing with the idea of adding a wrapper function, but wanted to keep this initial request as simple as possible. But I think you're right, and so here are two wrapper functions now.

About the function signature - I know this was a bold move - but since we just introduced these functions, there's probably no existing code out there that would break, apart from our own 😉 

I moved `format` up, because `width` and `height` are (in theory) optional parameters too (you can give 0 for both to get the smallest possible barcode). So `format` and `contents`/`data` are the only mandatory arguments, and I felt they should be next to each other, just as the are in (our flavour of) the [Android wrapper](https://github.com/markusfisch/zxing-cpp/blob/master/wrappers/aar/zxingcpp/src/main/java/de/markusfisch/android/zxingcpp/ZxingCpp.kt#L261). So, of all times, this is probably the moment to make this change 😉

benjohnde:
> About the function signature - I know this was a bold move - but since we just introduced these functions, there's probably no existing code out there that would break, apart from our own 😉

Haha, I actually liked the change - nothing wrong with the bold move though.

> I moved `format` up, because `width` and `height` are (in theory) optional parameters too (you can give 0 for both to get the smallest possible barcode). So `format` and `contents`/`data` are the only mandatory arguments [...]

True, I can relate. Just wanted to mention that overall. LGTM, the release notes should probably contain a gentle hint if @axxel agrees with the overall change.

axxel:
In general, I'm totally fine with those changes as things stand right now. That said I'd like to refer to the age old writer API discussion (again, I know), especially [this and following](https://github.com/zxing-cpp/zxing-cpp/issues/332#issuecomment-1630804296). I wished I had made more progress there by now but I have not. Bottom line: I'd like to introduce an `EncodeHints` struct along the lines of
```
struct EncodeHints {
  BarcodeFormat format;
  int/float scale; // AND/OR widthHint, heightHint
  int ecLevel; // the current interpretation of the numbers 0-8 is a mess and will not last
  int margin; // currently meaning pixels but should rather mean modules (imho)
  ... // to be expanded
}
```
See [here](https://github.com/zxing-cpp/zxing-cpp/issues/332#issuecomment-1638239345) and [here](https://github.com/zxing-cpp/zxing-cpp/issues/332#issuecomment-1641740838) for details about the `scale/width` thingy.

This whole size issue might convince me to separate the process into two steps (similar to what libzint is doing):
1. Generate an abstract "symbol" result that knows how many modules it contains, i.e. it's minimum size in pixels.
2. Supply member functions or free functions that allow to render this symbol into a bitmap of a certain size or maybe an SVG.

The SVG route would not have this size issue at all and I'm wondering whether that would not be the best result type to begin with? iOS should have some splendid options for rendering some SVG content in arbitrary sizes on the screen, right?

Maybe start with `writeToSvgFromString` or `writeSVGString` or however this aught to look to make the Swift auto overloading feature work? Even if we end up with the mentioned two-step process, those names would not collide and could stay around for backward compatibility.

markusfisch:
Well, while I can see why you want to get rid of width and height in favor of a simple scale parameter, as well as cleaning up ecLevel and margin, and even that it would be better to just return a clean 1x1 BitMatrix, I think all of this is probably a bit beyond this small extension of the iOS wrapper 😬 To change these things would affect all wrappers, of course, and we would like to have a way to specifiy these things now. 

So I would propose to tackle these changes in a second, bigger step. Of course, that would mean breaking changes, but that's what it means for all other wrappers anyway.

>  I'd like to introduce an EncodeHints struct…

Something like [this](https://github.com/kurzdigital/zxing-cpp/commit/5daae2c9c4f2957a75701e3c8cec016ed8d7e01e)?
If you like this, I would merge it in this branch.

> iOS should have some splendid options for rendering some SVG content in arbitrary sizes on the screen, right?

Unfortunately not 🙈  iOS doesn't support SVG directly. You can only bundle it in the assets of an app, but there are no native functions to work with SVG on iOS yet.

> This whole size issue might convince me to separate the process into two steps 

This is very similar to what I already do in [my flavor](https://github.com/markusfisch/zxing-cpp/blob/master/wrappers/aar/zxingcpp/src/main/java/de/markusfisch/android/zxingcpp/ZxingCpp.kt#L215) of the Android wrapper, as you know. So I totally agree with that. But I think this would be another step for the iOS wrapper.

axxel:
> Of course, that would mean breaking changes

My idea to bring this up now was to try to prevent (or limit) breaking changes in the iOS wrapper considering that this functionality is not really used by anyone yet (except you, it seems).

If there is no SVG support in iOS, how do you currently render the result? Do you have an area where you want to render the code into and then request just that size in pixels? Is there any other vector graphic API / data format that one could use or would it be required to "manually" draw primitives directly in some form of rendering API?

Thanks for the suggestion with the ZXIEncodeHints. Please merge that as you suggested. Looks like a good starting point.

My two-step idea has similarities to your Android code but as far as I saw, the difference would be that the first stop would not have anything like a `scale` or `width` only the rendering part.

One thing that I'd definitively like to add/change before a release is that the width/height parameters are only going to be a hint and the returned image would then contain just the symbol with requested quiet zones (note the current implementation already interprets `margin` as exactly that) but no extra white background just to produce a specific size. I had a short look at the `Inflate` helper and will later hack something with inverted sign of `width` to switch between the current and the new behavior, so `MultiFormatWriter` can be used for both until we have a generally new `WriteBarcode` style API on the c++ level.

alexmanzer:
> At the moment, we don't use any vector formats on iOS for barcodes. But this will change at some point, of course. I don't really know what we will use for that - that's a question @alexmanzer can maybe answer.

Current state:
- iOS still can't handle vector images like SVGs in (UIKit's) UIImage or SwiftUI Images. There are a lot of 3rd party libs (like [SVGKit](https://swiftpackageindex.com/SVGKit/SVGKit)) for this, but I'm not the biggest fan of even more external dependencies.
- UIImage, CIImage, CGImage are all pixel based Classes/Implementations
- There are workarounds without external dependencies like "use a webview, because they can render SVG's", but this is also not a real solution IMO
- The same is valid for PDF vector images
- Yes, I know, since [Xcode12 you can use SVG images](https://www.avanderlee.com/xcode/svg-image-assets/) in an app bundle. But that is not the same as downloading/creating a SVG and show it in an ImageView. This is still not possible.

For iOS development, SVG is still a "foreign body" and you need external libs to use it. It's sad - but that's the way it is.

Please correct me, if I'm wrong @benjohnde, @parallaxe and all the other iOS Devs out there.

axxel:
Thanks @alexmanzer to chime in from your vacation ;). I'll happily trust your expertise regarding SVG and iOS. But then let me repeat my earlier question: If there is no SVG support in iOS, how do you currently render the result? Do you have an area where you want to render the symbol into and then request just that size in pixels?

alexmanzer:
Yes, at least "something like that".

Example:
You have an UIImageView that can display a pixel image.
Since there are devices with different pixel densities (iPhone 11 has x2 and 11 Pro has x3), you would theoretically need multiple images with different resolutions for a pixel-perfect display.
So if a UIImageView in the app is 200x200pt (pt = points), then you would need a 400x400px image for 2x and a 600x600px image for 3x.

Long Story Short:
In practice, from my experience, this is how you do it (if pixel-perfect rendering is not crucial to life):
You just use an image with high enough resolution for the best possible display. And since you have enough computing power anyway, you don't care that iOS has to downscale the image for 2x devices. For a QR code with e.g. 64x64 modules it doesn't really matter in real life, because a single module is rendered with e.g. 9x9px.

(With the scaling mode you should be careful not to use a mode that interpolates bilinearly, but only scales up-down according to e.g. nearest neighbor. But even this is not "decisive for the war" in most real world applications.)

alexmanzer:
So from a pure iOS point of view I would say that it is either fine:
- to specify the needed image size
- or you just get a bitmap with 1x = 1 Module, which you scale up yourself.

By the way, in my opinion this is the most effective way to handle e.g. a QR code as an image anyway. A PNG with 64x64px (for 64x64 modules) is most likely much smaller than a SVG, where this structure is "described".
But that's another discussion, I believe. 😉

For this PR, I think it would be fine to just include the pixels needed in height and width.

axxel:
Thanks for the backgound info. I just wanted to make sure the user of the API is aware that just because they specified a `width` of 200 does not mean they'll get it. The resulting size might be lower or even higher if the symbol size is in modules is larger than 200.

As for the SVG vs Bitmap discussion: once we switch to libzint, we'll have the option to ask for text rendered into the result (think EAN13). And then SVG makes a lot of sens. In-fact rendering the symbol with a module size of 1px will produce a barely readable text.

markusfisch:
Thanks for merging! 👍 

> As for the SVG vs Bitmap discussion: once we switch to libzint, we'll have the option to ask for text rendered into the result (think EAN13). And then SVG makes a lot of sens. In-fact rendering the symbol with a module size of 1px will produce a barely readable text.

Ah, now I understand why you're focussed on SVG! Makes sense!

PR #646, opened 2023-10-27
ios: remove unnecessary setter functions

markusfisch:
These setters are already defined by `@property`.

### DIFF
--- a/wrappers/ios/Sources/Wrapper/Writer/ZXIEncodeHints.mm
+++ b/wrappers/ios/Sources/Wrapper/Writer/ZXIEncodeHints.mm
@@ -30,24 +30,4 @@
     return self;
 }
 
--(void)setFormat:(ZXIFormat)format {
-    self.format = format;
-}
-
--(void)setWidth:(int)width {
-    self.width = width;
-}
-
--(void)setHeight:(int)height {
-    self.height = height;
-}
-
--(void)setEcLevel:(int)ecLevel {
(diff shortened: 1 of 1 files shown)


axxel:
Ahh, so those setters were invisible (and therefore dead code) after all. ;). This makes #645 obsolete, right?

markusfisch:
No, no, this PR is about `ZXIEncodeHints` while #645 is about `ZXIDecodeHints` 😉 

The setters in `ZXIDecodeHints`, the ones that I corrected in #645, are actually necessary because they set the native ZXing hints, see [here](https://github.com/zxing-cpp/zxing-cpp/blob/master/wrappers/ios/Sources/Wrapper/Reader/ZXIDecodeHints.mm#L48) for example.

The setters in `ZXIEncodeHints` are just setting the property, which is already covered by `@property`.

axxel:
Thanks for your patience in explaining the world to me ;). So the renaming is just a cosmetic change that is not observable from code using these properties and the spelling in terms of lower vs. upper case is irrelevant to the ObjectiveC compiler?

markusfisch:
No problem, Objective-C is a bit weird from a C/C++ perspective 😉 

> So the renaming is just a cosmetic change that is not observable from code using these properties and the spelling in terms of lower vs. upper case is irrelevant to the ObjectiveC compiler?

Yes, oddly enough, setting a property is case insensitive in Objective-C, while methods are case sensitive.

PR #647, opened 2023-10-27
feat: add read only for android analyser

adrienbusin:
Adding param closeImageAfterRead

 if true, image will be read and closed after analysis (Default, no regression).
 else this function will only read the image but it will not close image
  
This could be useful (it's my case) if you want to analyse image with other analyser in parallel and close image manually

### DIFF
--- a/wrappers/android/app/src/main/java/com/example/zxingcppdemo/MainActivity.kt
+++ b/wrappers/android/app/src/main/java/com/example/zxingcppdemo/MainActivity.kt
@@ -27,9 +27,11 @@ import android.media.ToneGenerator
 import android.os.Bundle
 import android.os.Environment
 import android.view.View
+import androidx.annotation.OptIn
 import androidx.appcompat.app.AppCompatActivity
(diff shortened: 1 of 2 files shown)


axxel:
Thanks for reaching out. I had a look at the code overall and the documentation of `ImageProxy` and came to the conclusion that what I was doing there (closing inside `read`) is just wrong from a conceptual point of view. Would you please update your PR to simply not close inside read and fix the demo-app to do that outside, just like for the Java path already.

This is a breaking change but the right thing to do in my opinion. Any comments from @markusfisch, @benjohnde?

markusfisch:
Yes, I also think this is the right thing to do 👍 

(And another reason why I don't like having that `ImageProxy` dependency, but you know that already 😉)

axxel:
I decided to make this 2-line commit myself instead of asking you to cleanup your PR. I hope you don't mind. Thanks for the collaboration.

adrienbusin:
Ok ! thanks 🙂

PR #659, opened 2023-11-10
Android wrapper improvements

markusfisch:
And finally some improvements we haven't taked about yet:

* [make properties of DecodeHints mutable](https://github.com/zxing-cpp/zxing-cpp/commit/3451f7bf6a03a076a1d9798d09b27f585eaa5c85)
* [android: remove double code when creating enum](https://github.com/zxing-cpp/zxing-cpp/commit/17353dce018fa79c3a9392336224cd44efd48546)


wrappers/android/zxingcpp/src/main/java/com/zxingcpp/ZXingCpp.kt
```diff
@@ -0,0 +1,152 @@
...
+
+	public data class Position(
+		val topLeft: Point,
+		val topRight: Point,
+		val bottomLeft: Point,
+		val bottomRight: Point,
+		val orientation: Double
+	)
+
+	public data class Result(
```
axxel, line 91: There are some inconsistencies that I'd like to address:
 * You added `returnErrors` to `DecodeHints` but there is no corresponding `error` member here.
 * Some members are 'nullable', others are not. Whether or not this result describes a valid or erroneous symbol we always have a format, position, orientation and time (and arguably a sequence size and index). Everything else is optional / depending on the validity of the result. I understand that this optionality is canonically described with the `?` in Kotlin. Not sure whether a `Boolean?` would people make rise their eyebrows?

wrappers/android/zxingcpp/src/main/java/com/zxingcpp/ZXingCpp.kt
```diff
@@ -0,0 +1,152 @@
...
+		val ecLevel: String?,
+		val symbologyIdentifier: String?,
+		val sequenceSize: Int,
+		val sequenceIndex: Int,
+		val sequenceId: String,
+		val readerInit: Boolean,
+		val lineCount: Int
+	)
+
+	public fun read(image: ImageProxy, decodeHints: DecodeHints): List<Result>? {
```
axxel, line 108: Since a `List` can inherently be empty, I wonder whether `List<Restult>` would be the better return type here. I can imagine that the built in syntactic sugar of Kotlin to perform conditional operations based on the value of a nullable/optional would make it more convenient from a client perspective to use the optional as you did here. Just saying that from a conceptual point of view it does make sens to have two different ways of specifying 'nothing found'. If we stick with the optional, we should probably add a comment somewhere that the returned list object is either `null` or non-empty, otherwise you would have to check for nullness and emptiness?

axxel (changes requested):
@markusfisch thank you very much for your time spent on this.

@benjohnde If you have an opinion on the two comments I made regarding the use of optional types, please let us know.

wrappers/android/zxingcpp/src/main/java/com/zxingcpp/ZXingCpp.kt
markusfisch: > You added returnErrors to DecodeHints but there is no corresponding error member here.

Ah, my bad, of course. I've just added a commit to add an `error` member to `Result`. For the sake of simplicity I just return the `msg` string, but of course could add an Error structure with the corresponding type, if you wish so.

> Some members are 'nullable', others are not…

Well, I didn't change the existing members (to keep the changes as minimal as possible), and `time` and `postion` were optional before. But that really doesn't make much sense because they're always set on the [C++ side](https://github.com/zxing-cpp/zxing-cpp/pull/659/files#diff-b0fe4e53e8e58acbf0322650c3f08f1237c7d4a80738eb759a5293426224046fR199).

Now, that is also true all other members, which is why I have no optionals in the Result of my [wrapper](https://github.com/markusfisch/zxing-cpp/blob/master/wrappers/aar/zxingcpp/src/main/java/de/markusfisch/android/zxingcpp/ZxingCpp.kt#L99) (except of `gtin`, which we don't want here). I think this is the best match of the native Result, where `Result.text()` always returns a `std::string` object, and never `NULL`. So I thought the Kotlin/JVM Result should also behave that way and have an empty `String` instead of an optional `String?`.

Also, we can't really make something like `readerInit` a `Boolean?` because the native `Result.readerInit()` is just a `bool`. So `Result.readerInit` on the JVM will always be non-null, too.

So my suggestion would be to remove all optionals in Result 😉

wrappers/android/zxingcpp/src/main/java/com/zxingcpp/ZXingCpp.kt
markusfisch: > Since a List can inherently be empty, I wonder whether List<Restult> would be the better return type here. 

Hm, yes, that's true! From a client's point of view it probably makes more sense to always return `List<Result>` and cut the nullness case. One thing less to worry about.

The reason why I chose to return `null` (if there are no results) is, that this saves us from allocating an empty `ArrayList` in the hot path. But that's not a big deal, of course, and also probably more of a leaking abstraction. So if you prefer `List<Result>` over `List<Result>?` I will happily change it.

Syntactically it's almost the same to test for nullness or emptiness in Kotlin.

>  If we stick with the optional, we should probably add a comment somewhere that the returned list object is either null or non-empty, otherwise you would have to check for nullness and emptiness?

Agreed, otherwise a comment would make sense 👍 Maybe just before [read()](https://github.com/zxing-cpp/zxing-cpp/pull/659/files#diff-40bfa6be62aab191f820c872aa4229fcaabe9efdc96a6578ee56a2d071d3c175R108)?

axxel: I believe `List<Result>` is the most natural return type. Thanks.

wrappers/android/zxingcpp/src/main/java/com/zxingcpp/ZXingCpp.kt
axxel: If we can agree that `text : String?` is the correct model, then `sequenceId` should also be optional.

wrappers/android/zxingcpp/src/main/java/com/zxingcpp/ZXingCpp.kt
markusfisch: I've changed the return type to `List<Result>` in cf3cb80.

wrappers/android/zxingcpp/src/main/java/com/zxingcpp/ZXingCpp.kt
markusfisch: There's now a proper `Error` struct, `sequenceId` is nullable, and `Result.text` will now be `null` when the `Result` is not valid (as are the other nullable members of `Result`).

> But in Kotlin I got the impression that having a String? immediately communicates to the user that this thing can be nonexisting and there are language features to nicely deal with that.

Yes, exactly.

> readerInit could be set to null based on Result::isValid() is we really wanted to. I'm not convinced that this is required, just saying ;).

I skipped that for now 😉 If you want, I can change it to `Boolean?`, too. Although it might make more sense to add a `valid` member instead of making all members optional.

axxel:
Progress. Awesome. :)

PR #661, opened 2023-11-14
ios: use a default error text instead of crashing

markusfisch:
Because `e.what()` returns `nil` sometimes, what is causing a crash when initializing the dictionary.

### DIFF
--- a/wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.mm
+++ b/wrappers/ios/Sources/Wrapper/Writer/ZXIBarcodeWriter.mm
@@ -76,10 +76,10 @@ std::wstring NSDataToStringW(NSData *data) {
         return [self inflate:&bitMatrix];
     } catch(std::exception &e) {
         if (error != nil) {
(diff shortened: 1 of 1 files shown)


axxel:
Do you see any chance to reproduce the issue? According to https://stackoverflow.com/a/1038504 `e.what()` can only return `nullptr` if a custom class returns that (or the iOS stdlib implementation is buggy). If there is an issue we can fix somewhere in the writer implementation, I'd prefer that.

markusfisch:
No, unfortunately we don't have a sample or any more info 🤷‍♂️  But we're seeing this crash happening to our users in our crash logging system.

axxel:
I checked all `throw` statements and this looks like a bug in the libc++ implementation on iOS. weird.

axxel:
It is so nice if unrelated compilation issues (see recent wasm build failures) that randomly show up can also randomly disappear again... ;)

markusfisch:
Well, in my experience, such random issues unfortunately always come back 😉 

(And I was wondering about the wasm build failures already, too).

PR #664, opened 2023-11-17
ios: improve exception handling

markusfisch:
Make `read()` throw so C++ exceptions can be caught in Swift, so an app can't crash silently anymore when a C++ exception is thrown while trying to read a barcode.

This also adds a function to properly set a NSError from a C++ exception to avoid code duplication in ZXIBarcodeReader/Writer.

### DIFF
--- a/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.h
+++ b/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.h
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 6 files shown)


axxel:
For my understanding: setting this `NSError*` to something non-nil will 'throw' an exception on the Swift side?

markusfisch:
Yes, exactly! 

Here are the details on [how Cocoa error parameters are converted to Swift throwing methods](https://developer.apple.com/documentation/swift/about-imported-cocoa-error-parameters). Also check out how [Cocoa Errors are handled in Swift](https://developer.apple.com/documentation/swift/handling-cocoa-errors-in-swift).

PR #666, opened 2023-11-20
android: manually convert Format set to String

markusfisch:
Because `Set.toString()` doesn't exist on all Android versions.


axxel:
I was looking at your code, wanted to suggest a few things and ended up simply doing it at bit differently. May I ask which Android version is lacking the `toString()` function?

markusfisch:
Ah, nice! I like your implementation better, too! 👍 

Unfortunately, I can't really tell what version excactly, but I ran into this on an old device running Android 7 (API level 24). According to the documentation `Set.toString()`, which is inherited from [AbstractCollection.toString()](https://developer.android.com/reference/java/util/AbstractCollection#toString()), should exist since API 1.

I _guess_ `Set.toString()` was added with Android 8.0 (API level 26) which [updated Java language support](https://developer.android.com/about/versions/oreo/android-8.0#java) and brought support for "several additional OpenJDK Java APIs".

PR #675, opened 2023-11-29
qrdecoder: fix `DecodeAlphanumericSegment()` percent/FNC1 loop

gitlost:
Blast from the past...


axxel:
Thanks. Looking at the diff I tried to see what the issue with the old code was. My best guess: the loop termination was wrong resulting in garbage at the end of the string?

gitlost:
Didn't examine it too closely to be honest but it was to do with the
`erase()` lopping off the rest of the string. If you run the unit test
without the change it results in ""9112%" instead of "9112%<GS>2012".

PR #679, opened 2023-12-05
fix: wrong path for consumer-rules

adrienbusin:
Since android path change from com.zxingcpp to zxingcpp, consumer file is not updated


axxel:
Thx!

PR #681, opened 2023-12-05
Add support for Rectangular Micro QR Code (rMQR), pure only

gitlost:
- Result: allow this or other to have `lineCount` 1 and swop calcs accordingly; "rssexpanded-2/4_03.png" causes `assert(lineCount() == 1)` to trigger; also makes equals symmetrical
- Add support for Rectangular Micro QR Code (rMQR), pure only


gitlost:
Just trying it out now...

axxel:
The `BarcodeFormat::rMQR` does not quite fit into the list IMHO. I noticed that the spec speaks of "Rectangular Micro QR Code (rMQR)" (which supports your choice) but the Denso marketing term is [rMQR Code](https://www.denso-wave.eu/en/denso-products/rmqr-code) which would translate to `rMQRCode` or maybe even `RMQRCode` and fits better into the list of existing names, in my opinion.

What is your take on that? I can easily change that later myself.

gitlost:
I'm easy - the term "rMQR" is used by the ISO/IEC document in the title so
it has some "official" status...

gitlost:
Ok the conversation's gotten out of sync but just made the change to
"RMQRCode" locally and yes, it does fit better in the list of existing
names. I can push the change if you like...

axxel:
I changed it myself while adding all the wrapper copies of that as well (which are quite a few nowadays... ;))

I did a few quick tests with the android demo app and it actually works better than expected with real input. Very cool.

Issue #662, opened 2023-11-14
Adding support for DX film edge barcode


Merinorus:
# Adding support for DX film edge barcode
Hello ZXing team,

I'm working on a personal project where I would like to be able to scan DX film edge barcodes.

## DX film edge barcodes?

These are the kinds of barcodes we find on the edge of 35mm negative films, like at the bottom of the following image:
![Pasted image 20231114195712](https://github.com/zxing-cpp/zxing-cpp/assets/3023499/fda7de46-7c0b-4c63-b7e4-2767290fb4bd)

They refer to the [DX encoding system](https://en.wikipedia.org/wiki/DX_encoding).

In this barcode, there are two numbers:
- the film model (manufacturer and type of film)
- the current image number (+ a flag to indicate whether we are in the middle or the end of the image)

## So... Why would I need to read them?

More and more people are involved in digitizing negatives at home. The film model number is important to know, so we can apply the right color treatment to the pictures.

Some people use dedicated film scanners, which already read this code and can apply the appropriate color treatment automatically.

Other people, including me, use a DSLR to digitize the films. This is done in two main steps:
- Take a picture of the negative with a proper setup (backlight, macro lens...): with a lot of setups, only the picture itself is digitized, not the barcode.
- Apply an appropriate color treatment on the computer to retrieve more or less the original colors, to obtain the "positive". Darktable, Lightroom, Negative Lab Pro, etc. are some tools that can help in this way. Some tools can help in guessing the color treatment but, generally, it's better to know the film model.

This is why I'm interested in adding the support for DX film edge barcode. Some industrial products with closed-source software already do this kind of job to apply a color treatment, but the number itself is not displayed and I didn't find any open-source solution to do it.

## How to read them

DX edge barcodes have some characteristics in common with QR codes:
- patterns to indicate the beginning and the end of the barcode,
- a "clock" to help read the signal with the machine,
- the signal itself (2 dimensions in the QR codes, 1 dimension in the DX film edge barcode).

An expired [Kodak patent](https://patents.google.com/patent/US4965628A/en) explains how to read them ([pdf version](https://patentimages.storage.googleapis.com/d6/13/7d/51ea3bebaa8c99/US4965628.pdf)).

Here is an image extracted from the patent that explains the pattern. It does not explain exactly how to decode it but it gives an idea. If you are interested, I can post how to decode it step by step.

[...]

PR #684, opened 2023-12-10
Add support for DX Film Edge read

Merinorus:
# Description
Add support for reading DX Edge barcodes on 35mm films.
Implements #662.

# To do
- [ ] Unit tests?
- [ ] More black-box tests? Especially with multiple codes on the same image
- [ ] Check the `SymbologyIdentifier`
- [ ] Ensure the output format is convenient
- [ ] Find the best values to improve detection without any false positive
- [ ] Disable by default?

# Demo

Most recent films include the half-frame number (number of the frame, followed by an "A" if we're in the middle of the frame):

https://github.com/zxing-cpp/zxing-cpp/assets/3023499/582914c5-a0a5-4ca5-9aff-14a23ca64898

Oldest films do not include this number:

https://github.com/zxing-cpp/zxing-cpp/assets/3023499/83a8c79e-e508-4ebf-b5a9-af15216e031b


axxel:
Thanks for this very complete and comprehensive PR :). Bad news first: I will not merge this as is. The main reason is that it slows down the library considerably. This new detector takes more time than all other linear codes combined. That is not your fault, though. It is caused mainly by the `FindLeftGuard` call with such a long fixed pattern, which we don't have anywhere else. But there is ample opportunity to improve the situation.

The first thing we should do is to only look for the code where we can reasonably expect it in a real-world scenario. If I understand the use case correct, then swiping with a phone over a film strip, like you did in your two videos, is not really something one would do? I thought the input would be way more constraint:
 1. near perfect lighting, i.e. no over or underexposure because then the image would be useless anyway
 2. always a landscape oriented scan with the film-center close to the image center
 3. always scanning from the front-side of the film with the normal orientation (barcode at the bottom of the image, not like in your `1.jpg` sample)
 4. maybe even generally only one film frame per image?

Can you confirm or dispute any/all of those assumptions?

Then there are a few more 'cosmetic' changes that I would like see before merging the code. Would you prefer to read a detailed review and do those changes yourself or rather have me do them (or some of them) and simply add them to your PR?

gitlost:
Something that might be considered is to introduce "opt-in" barcode
formats, i.e. one's that aren't scanned for by default but must be
explicitly requested...

axxel:
@gitlost we are on the same page here, see [this comment](https://github.com/zxing-cpp/zxing-cpp/issues/662#issuecomment-1811515769) ;)

gitlost:
Cool!

axxel:
I spent an unexpected amount of time yesterday to fix the performance issue. After my initial guess turned out to be wrong, it took me ages until I figured out where exactly all the time was spent (basically calling the `FindLeftGuard` multiple time for the same row was the problem). During this debugging I considerably changed the algorithm. And while I was at it, I also simplified the code and used existing tools instead of your new implementation (e.g. `toDecimal`). And I changed a few names (like "clock signal" -> clock track"). In the end, I touched almost each and every line. I saw no point in detailing all this to you. To make both of our contributions clearly visible in the project history I decided to merge your PR (with the compressed sample images to not bloat the repo unnecessarily) and then later add my changes.

With the current performance the additional DXFilmEdge detector slows down the whole process only about 1%. I see no point in not including it during the standard detection.

From your ToDo list, only the "check the output format" is open in my opinion. Maybe there is a hint in the Kodak patent at how to write down those 3 numbers in a string?

Thanks a lot for your work on this and welcome as a new contributor.

PR #697, opened 2024-01-07
Rust Wrapper

SteveCookTU:
This wrapper follows the implementation of the C wrapper.

With Rust <-> CPP limitations, some implementations are not available such as `ReadBarcodes`.

This aims to address #696 

Todo:
- [x] Add documentation
- [x] Add README
- [x] Use builder pattern for `ReaderOptions`


axxel:
After having had a look at your pro/cons arguments (thanks for that!), my experience with the bindgen wrapper and the current autocxx code (which has an issue with the `format()` member of `Result`), I can say that I can rule out the autocxx version. Could you please commit your cxx patch to this PR, this helps me with evaluating it on my side? Thanks.

SteveCookTU:
I'll be able to bring the cxx branch to the PR. The main issue in general for the format function is needing a Rust implementation for `BarcodeFormat` to be able to use bitflags in the Rust API. The issue will still be there using cxx for the same reason. 

We can use the enum from C++ as a source of truth but ints would still need to be passed in for the `setFormats` function. This can be accomplished by just implementing `BitOr` on the enum from C++ and using the same shim function.

axxel:
This work is now superseeded by the bindgen based approach that is merged and released on crates.io.

PR #700, opened 2024-01-12
fix out_of_range exception when edge coord inside image but not coords

simeosch:
std::out_of_range exception occurs with some images when try to detect qr codes, here is an image that generate the exception
qr-detect-fail-8863445


axxel:
Thanks for reporting this. The issue was more general, so I fixed it properly myself.

PR #717, opened 2024-02-08
ios: make sure errorDescription is not nil

markusfisch:
Since `[NSString stringWithUTF8String: message]` can still return `nil`. So it's important to check `errorDescription` before creating the dictionary to avoid an `NSInvalidArgumentException`.

Which we saw in our crash logs.

### DIFF
--- a/wrappers/ios/Sources/Wrapper/ZXIErrors.mm
+++ b/wrappers/ios/Sources/Wrapper/ZXIErrors.mm
@@ -13,9 +13,14 @@ void SetNSError(NSError *__autoreleasing _Nullable* error,
     NSString *errorDescription = nil;
     if (message && strlen(message) > 0) {
         try {
-            errorDescription = [NSString stringWithCString: message encoding: NSASCIIStringEncoding];
+            errorDescription = [NSString stringWithUTF8String: message];
         } catch (NSException *exception) {
-            errorDescription = @"Unknown ObjC error";
+            try {
+                errorDescription = [NSString stringWithCString: message
+                                                      encoding: NSASCIIStringEncoding];
+            } catch (NSException *exception) {
+                errorDescription = @"Unknown ObjC error";
+            }
         }
     }
     if (errorDescription == nil) {
--- a/wrappers/ios/Sources/Wrapper/ZXIErrors.mm
+++ b/wrappers/ios/Sources/Wrapper/ZXIErrors.mm
@@ -13,7 +13,7 @@ void SetNSError(NSError *__autoreleasing _Nullable* error,
     NSString *errorDescription = nil;
     if (message && strlen(message) > 0) {
         try {
-            errorDescription = [NSString stringWithUTF8String: message encoding: NSASCIIStringEncoding];
+            errorDescription = [NSString stringWithCString: message encoding: NSASCIIStringEncoding];
         } catch (NSException *exception) {
             errorDescription = @"Unknown ObjC error";
         }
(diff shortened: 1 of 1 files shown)


wrappers/ios/Sources/Wrapper/ZXIErrors.mm
```diff
@@ -10,14 +10,17 @@ void SetNSError(NSError *__autoreleasing _Nullable* error,
...
-    NSString *errorDescription = @"Unknown C++ error";
+    NSString *errorDescription = nil;
     if (message && strlen(message) > 0) {
         try {
             errorDescription = [NSString stringWithUTF8String: message];
         } catch (NSException *exception) {
             errorDescription = @"Unknown ObjC error";
         }
     }
+    if (errorDescription == nil) {
```
axxel, line 21: The condition in line 14 must have been true for the new line 21 to evaluate to true. But the only reason I can imagine why `NSString stringWithUTF8String` would return `nil` in that case is that `message` did not contain a valid UTF8 sequence. Do you see any other possible cause?

Then I'd be very interested to learn about the exact content of `message` and fix that one instead of 'burrying' this problem in the underlying code underneath the "Unknown C++ error" message. Can you get some information about the contents of `message`?

markusfisch: Yes, and indeed we have just found such a message: this can happen when a user inputs invalid data when generating a barcode. Then, the error message contains an invalid UTF8 sequence.

For example: try encoding "¥" as Code128 (nonsense, of course), and the error message is something like `Bad character in input: \xa5` for which `NSString stringWithUTF8String` returns `nil`.

Now, we would fix this by using `NSASCIIStringEncoding` for `NSString stringWithUTF8String`. In this case, `NSString stringWithUTF8String` returns the message just fine. Of course, this can produce invalid UTF8 sequences on the iOS side, but these are still better than having the generic "Unknown C++ error" message. I will push a corresponding commit.

axxel: Ahh, the writer API... I always keep forgetting it even exists ;). I understand your intention to improve the situation with interpreting the message as ASCII data but in my opinion that is still not the correct fix for the bug. I never thought about the character set of the error messages (except for the C-API where I explicitly specified (at least in my mental model) that every `char*` returned is UTF8 encoded, including any error messages. So the correct fix for this particular crash should be to replace [this line](https://github.com/zxing-cpp/zxing-cpp/blob/609ea8d78b65fa31070d1bb79c847562c33d9cbc/core/src/oned/ODCode128Writer.cpp#L156) with
```c++
throw std::invalid_argument("Bad character in input: " + ToUtf8(contents.substr(i, 1)));
```
Now, there might be more cases in the source code like this one. To not crash in those circumstances, I'd suggest a slightly different approach in the wrapper: use `NSString stringWithUTF8String` first and if that returns `nil` then call the ASCII version and maybe even hint or log an "internal error", so we might find any remaining issues at a later time.

axxel: I believe your latest fix wont work, since `stringWithUTF8String` apparently does not throw anything but silently returns `nil`. If it would throw, then you would have gotten "Unknown ObjC error" before instead of `nil`.

markusfisch: Ah, of course, you're right 🙈 I should take a break 😉 I've just pushed another commit.

Sorry for the mess.

axxel: Your 4th attempt still fails to achieve the goal. Looks like it is not your day today ;). I decided to push a change that I believe to be correct myself. Thanks for your effort, nonetheless.

PR #718, opened 2024-02-08
publish-android.yml: fix typo of ci name

ISNing:


axxel:
Thx. A one character PR, that is uncommon ;).

ISNing:
Just because I'm current working on binding this onto kotlin/native and noticed this, I believe the second pr will soon be available after I tested ci building

axxel:
Interesting, never heard of kotlin/native before... reading the introduction makes me anticipate another use for the C-API coming up. ;).

PR #730, opened 2024-02-17
cmake: Make build reproducible across different build directories

EchterAgo:
Without this, the usage of the __FILE__ macro leaves the build directory in the binary. When building the Python extension with build isolation enabled this would lead to random paths in the binary.


EchterAgo:
This mostly works, but the path of `ZXAlgorithms.h` still remains in the binary:

```
$ strings wrappers/python/build/lib.linux-x86_64-cpython-311/zxingcpp.cpython-311-x86_64-linux-gnu.so | grep $HOME
/home/ago/src/zxing-cpp/core/src/ZXAlgorithms.h
```

This seems to be from the precompiled headers, which somehow include the file from `zxing-cpp/core/src/ZXAlgorithms.h` instead of `zxing-cpp/wrappers/python/core/src/ZXAlgorithms.h`.

Anyone have an idea why this is? I checked the include paths and it has only the wrapper path. However when I copy the root `core` directory to the wrapper and delete `ZXAlgorithms.h` in the wrapper directory, the correct path ends up in the binary, so it does seem like some path issue to me.

EchterAgo:
When inspecting `wrappers/python/build/temp.linux-x86_64-cpython-311/ZXing/CMakeFiles/ZXing.dir/cmake_pch.hxx.gch.d` you can see that it seems to mix include paths:

```
 /home/ago/src/zxing-cpp/wrappers/python/core/src/ByteArray.h \
 /home/ago/src/zxing-cpp/wrappers/python/core/src/CharacterSet.h \
 /home/ago/src/zxing-cpp/wrappers/python/core/src/GTIN.h \
 /home/ago/src/zxing-cpp/core/src/ZXAlgorithms.h \
 /home/ago/src/zxing-cpp/core/src/Error.h /usr/include/c++/13/algorithm \
...
 /home/ago/src/zxing-cpp/wrappers/python/core/src/TextUtfEncoding.h \
 /home/ago/src/zxing-cpp/wrappers/python/core/src/ZXConfig.h \
 /home/ago/src/zxing-cpp/wrappers/python/core/src/Content.h \
 /home/ago/src/zxing-cpp/core/src/ReaderOptions.h \
 /home/ago/src/zxing-cpp/wrappers/python/core/src/DecodeHints.h \
 /home/ago/src/zxing-cpp/wrappers/python/core/src/ImageView.h \
 /home/ago/src/zxing-cpp/wrappers/python/core/src/Point.h \
```

EchterAgo:
This seems to only happen with `target_precompile_headers`.

EchterAgo:
Ah, I get it, PCH starts including `BarcodeFormat.h` from the wrapper directory, that file has a `#include "Flags.h"`, which in turn incldues `BitHacks.h`.

This matches the `cmake_pch.hxx.gch` file:

```
ZXing/CMakeFiles/ZXing.dir/cmake_pch.hxx.gch: \
 /home/ago/src/zxing-cpp/wrappers/python/build/temp.linux-x86_64-cpython-311/ZXing/CMakeFiles/ZXing.dir/cmake_pch.hxx.cxx \
 /usr/include/stdc-predef.h \
 /home/ago/src/zxing-cpp/wrappers/python/build/temp.linux-x86_64-cpython-311/ZXing/CMakeFiles/ZXing.dir/cmake_pch.hxx \
 /home/ago/src/zxing-cpp/wrappers/python/core/src/BarcodeFormat.h \
 /home/ago/src/zxing-cpp/core/src/Flags.h \
 /home/ago/src/zxing-cpp/core/src/BitHacks.h /usr/include/c++/13/cassert \
<snip>
```

So it seems like the include path for PCH dependencies is wrong?

EchterAgo:
GCC actually resolves the symlinks:

```
[pid 195830] readlink("/home", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] readlink("/home/ago", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] readlink("/home/ago/src", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] readlink("/home/ago/src/zxing-cpp", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] readlink("/home/ago/src/zxing-cpp/wrappers", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] readlink("/home/ago/src/zxing-cpp/wrappers/python", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] readlink("/home/ago/src/zxing-cpp/wrappers/python/core", "../../core", 1023) = 10
[pid 195830] readlink("/home/ago/src/zxing-cpp/core", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] readlink("/home/ago/src/zxing-cpp/core/src", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] readlink("/home/ago/src/zxing-cpp/core/src/Flags.h", 0x7ffda37f38d0, 1023) = -1 EINVAL (Invalid argument)
[pid 195830] openat(AT_FDCWD, "/home/ago/src/zxing-cpp/core/src/Flags.h", O_RDONLY|O_NOCTTY) = 5
[pid 195830] newfstatat(5, "", {st_mode=S_IFREG|0664, st_size=4452, ...}, AT_EMPTY_PATH) = 0
[pid 195830] read(5, "/*\n* Copyright 2020 Axel Waggers"..., 4452) = 4452
[pid 195830] close(5)                   = 0
```

I'm still not quite sure how this then winds up in the binary, but it is definitely through the `cmake_pch.hxx.gch`

EchterAgo:
Updated to find the project root.

core/CMakeLists.txt
```diff
@@ -44,6 +44,32 @@ else()
...
+    while(NOT "${current_root}" STREQUAL "")
+        if(EXISTS "${current_root}/LICENSE")
+            break()
+        endif()
+        get_filename_component(current_root ${current_root} DIRECTORY)
+    endwhile()
+    set(${result_variable} ${current_root} PARENT_SCOPE)
+endfunction()
+
+find_project_root(PROJECT_ROOT)
```
axxel, line 60: I think (and hope) that what you are looking for is simply `CMAKE_SOURCE_DIR`. This function does not sit well with me.

EchterAgo: I'll check, thanks for the hint.

EchterAgo: `CMAKE_SOURCE_DIR` is also `/home/ago/src/zxing-cpp/wrappers/python` when building the Python wrapper. 

I want to strip out the path prefix from the `__FILE__` log messages, but when building the Python wrapper some precompiled headers end up with `/home/ago/src/zxing-cpp` prefix due to GCC resolving the symlinks.

I'll think about alternative approaches.

EchterAgo: Why are you symlinking core in the wrapper instead of just calling `add_subdirectory` on the non-symlinked path?

I just tried and that seems to work great. I'll push an update with this.

EchterAgo:
Updated to not loop over the parent directories.

I noticed that the `.editorconfig` file does not match some of the files in the repo, for example for .py it specifies tab indent, but the `setup.py` file is space indented. I think we should change one of them to make it consistent. Should I raise an issue for that or do a PR?

core/CMakeLists.txt
EchterAgo: I think this is resolved.

axxel: > Why are you symlinking core in the wrapper instead of just calling `add_subdirectory` on the non-symlinked path?

Because this is part of the fix that made sure the c++ source is part of the python source package. Hence your update will break this.

Which directory is your top-level cmake dir? Are you using setup.py to build your module?

axxel:
> I noticed that the `.editorconfig` file does not match some of the files in the repo, for example for .py it specifies tab indent, but the `setup.py` file is space indented. I think we should change one of them to make it consistent. Should I raise an issue for that or do a PR?

Please raise an issue. I'll have to look into it.

core/CMakeLists.txt
EchterAgo: > Because this is part of the fix that made sure the c++ source is part of the python source package. Hence your update will break this.

Sorry, I missed that. I've restored the symlink and just make it resolve the path if it is a symlink.

> Which directory is your top-level cmake dir? Are you using setup.py to build your module?

I'm not sure, I am using pip to install the package using `pip install wrappers/python`

EchterAgo:
Updated, `sdist` works again.

wrappers/python/CMakeLists.txt
```diff
@@ -17,8 +26,11 @@ if (NOT hasParent)
     option (BUILD_WRITERS "Build with writer support (encoders)" ON)
     option (BUILD_READERS "Build with reader support (decoders)" ON)
     set(BUILD_DEPENDENCIES "AUTO")
-    if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/core)
```
axxel, line 20: Looking at this line, it occurred to me that is does not make any sense in current HEAD because `core` always exists. Then I realized: your function is not necessary at all. Simply do the `IS_SYMLINK` check on `core` (which will fail in an sdist build) and if it is a symlink, then hardcode the `../../core` path. Less lines of code to worry about.

EchterAgo: Will change, but maybe not today.

EchterAgo: Updated, I still kept the EXISTS check but made a variable `CORE_PATH` that is set conditionally based on whether `./core`  is a symlink.

axxel:
@EchterAgo I just noticed that this change broke my debugging environment because the IDE can't find the sources anymore and displays assembly instead. A quick fix is to only apply the file-prefix-map flag in a release build. I guess this would still satisfy your requirements?

EchterAgo:
Maybe try `-ffile-prefix-map=${CMAKE_CURRENT_SOURCE_DIR}=.` (note the trailing dot).

I think just using `-fmacro-prefix-map` instead of `-ffile-prefix-map` should also solve this. File prefix map sets the prefix map for all options, `-fmacro-prefix-map`, `-fdebug-prefix-map`, `-fprofile-prefix-map` and `-fcanon-prefix-map`. See https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#:~:text=args%20cc1%20%E2%80%A6%E2%80%99.-,%2Dffile%2Dprefix%2Dmap,-%3Dold%3D

PR #733, opened 2024-02-18
Example for displaying QR codes in QML

UpdogCoder:
Hi!
I've done some work on my project and was using the ZXing-C++ together with QML to display the QR codes. Because of that I wanted to add a basic example of doing that since the examples contain only reading the barcodes in QML.

Logic for generating the barcode is based on the methods from Qt Writer example by Alex and main goal was to show how to expose all of the generated stuff to QML.

Additionally added not in README about usage with Qt to indicate that it is possible to use ZXing-C++ directly with QML.


axxel:
Thanks for your contribution. Here are my thoughts:
* please replace the space with tab based indentation (see .editorconfig / .clangformat)
* the indentation in the CMakeLists.txt is off
* you mentioned a change in some README but that is not part of the PR
* would you be able to add a setting for the BarcodeFormat to make the sample a bit more usable?

axxel:
There was no feedback from the OP and this PR is based on now outdated code.

PR #734, opened 2024-02-21
Install Barcode.h as it's used in the installed ReadBarcode.h

vkrause:
Fixes compilation for consumers using installed headers.


axxel:
Ups... thx.

Issue #737, opened 2024-02-26
IOS wrapper does not use the provided options


ninio:
In the demo project of the iOS wrapper, any settings passed to the reader are ignored.

E.g. if you update the options in `viewDidLoad` method of the `ViewController`, they are not used when calling the reader.

Example:
```
class ViewController: UIViewController {
    /* ... */
    override func viewDidLoad() {
        super.viewDidLoad()
        
        self.reader.options.tryHarder = false;
         /* ... */
    }
```

axxel:
That sounds embarrassing... ;). I've never seen the iOS code in action myself, so I can't confirm nor contradict your statement. Maybe @benjohnde, @markusfisch, @alexmanzer or @parallaxe can say something about it.

That said, I recently noticed something fishy with regards to the `ReaderOptions` handling myself. Could you maybe try a simple change in the code and see if that helps? I suggest to replace [this line](https://github.com/zxing-cpp/zxing-cpp/blob/b3aff4a98b03e056a244ca385a8221c50d67e352/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm#L140) with
```ObjectiveC
        Barcodes results = ReadBarcodes(imageView, self.options.cppOpts);
````
And if that does not compile, then there might be some change required in `ZXIReaderOptions.h` to get outside access to [cppOpts](https://github.com/zxing-cpp/zxing-cpp/blob/b3aff4a98b03e056a244ca385a8221c50d67e352/wrappers/ios/Sources/Wrapper/Reader/ZXIReaderOptions.mm#L9). I don't know ObjectiveC.

ninio:
The `ReaderOptions` (`cppOpts`) are once created when initializing the reader (and its options) (`ZXIReaderOptions.mm:16`), but when calling the `reader.read`, a new `ReaderOptions` instance is created (`ZXIBarcodeReader.mm:124`)

I can (try to) create a PR which has the ZXIReaderOptions store only the properties (without a `self.cppOpts`), and do the options ObjC->cpp translations only in the `ReaderOptionsFromZXIReaderOptions`. 

Is this a viable approach, or am I missing something?

axxel:
> The `ReaderOptions` (`cppOpts`) are once created when initializing the reader (and its options) (`ZXIReaderOptions.mm:16`), but when calling the `reader.read`, a new `ReaderOptions` instance is created (`ZXIBarcodeReader.mm:124`)

[...]

PR #738, opened 2024-02-26
ios: directly use cppOpts from ZXIReaderOptions

markusfisch:
And remove the now incomplete `ReaderOptionsFromZXIReaderOptions` transfer method.


ninio:
This PR fixes #737

PR #745, opened 2024-03-10
kn: Add cmake building directory as a root of searching header

ISNing:


axxel:
I meant only your change, sorry if that was unclear.

ISNing:
@axxel Is it now right?

PR #763, opened 2024-04-10
Add support for detecting and reading Aztec Runes

gormster:
This pull adds support for Aztec Runes, as described in Annex A of ISO/IEC 24778:2008.

Runes are identical to the Core Symbol of a Compact aztec code, but with every other bit of the data message inverted to distinguish them from regular Mode Messages. In other words, all we need to do to detect a rune is (a) fail to decode a regular Mode Message and then (b) XOR each 4-bit word in the data message with `0b1010` and try again. If it decodes this time, then we've got a rune.

Let me know if there's any changes you'd like to see!


test/samples/aztec-1/rune-1.png
axxel, line 1: Since I care about the size of the repository binary blob, please replace the file with the following minimized version:
rune-1

test/samples/aztec-1/rune-64.png
axxel, line 1: Same here:
rune-64

axxel (changes requested):
Thanks for the contribution! Since you asked for requests for changes, here they come... ;).

core/src/aztec/AZDecoder.cpp
```diff
@@ -346,9 +346,22 @@ DecoderResult Decode(const BitArray& bits)
 	return DecoderResult(std::move(res)).setStructuredAppend(sai);
 }
 
+DecoderResult DecodeRune(const DetectorResult& detectorResult) {
+	Content res;
+	res.symbology = {'z', 'C', 3};
+
+	res.push_back(detectorResult.runeValue());
```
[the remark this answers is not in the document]
gormster: huh… that's kinda crazy, but I guess that is technically what the spec says.

core/src/aztec/AZDetector.cpp
```diff
@@ -291,7 +292,21 @@ static int ModeMessage(const BitMatrix& image, const PerspectiveTransform& mod2P
...
+	bool decodeResult = ReedSolomonDecode(GenericGF::AztecParam(), words, numECCodewords);
+
+	if ((!decodeResult) && compact) {
+		// Is this a Rune?
+		for (int i = numCodewords - 1; i >= 0; --i) {
+			words[i] ^= 0b1010;
+		}
+		decodeResult = ReedSolomonDecode(GenericGF::AztecParam(), words, numECCodewords);
+
+		if (decodeResult)
```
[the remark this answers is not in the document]
gormster: The `if` here tells us that it definitely _is_ a rune, otherwise it's just a failure to decode the core symbol. While that doesn't matter right now – we check it again and `return -1` right afterward – it would be better not to set isRune = true if that's not actually necessarily the case.

Like, imagine you're debugging a symbol that fails to decode by setting a breakpoint here, and you see that `isRune` is set to true, you would probably be misled into thinking that the code is being detected as a rune for some reason - but it's not.

Unless you want to stick the `return -1` in the if block? Like

```cpp
// line 310
		if (!decodeResult)
			return -1;

		isRune = true;
	}
```

though I think I'd like to make it explicit that `isRune` being set is directly caused by `decodeResult` being true.

```cpp
// line 310
		if (decodeResult)
			isRune = true;
		else
			return -1; // Failed to decode as either normal or rune
	}
```

core/src/aztec/AZDetector.cpp
```diff
@@ -409,7 +427,7 @@ DetectorResults Detect(const BitMatrix& image, bool isPure, bool tryHarder, int
 		if (!bits.isValid())
 			continue;
 
-		res.emplace_back(std::move(bits), radius == 5, nbDataBlocks, nbLayers, readerInit, mirror != 0);
+		res.emplace_back(std::move(bits), radius == 5, nbDataBlocks, nbLayers, readerInit, mirror != 0, isRune ? modeMessage : 0);
```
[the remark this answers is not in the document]
gormster: Yeah I thought about this – I do specifically call out in the header file that `runeValue()` is only valid if `nbDataBlocks == 0` – but you're probably right that a belt-and-braces approach is better. This does mean I'll have to change to type to mean it's possible to represent invalid rune values, but oh well.

> see also AZDetector.h

what am I looking at in AZDetector.h, sorry?

test/samples/aztec-1/rune-1.png
gormster: whoa that's a lot smaller. what's the difference? I can see metadata is gone but surely that wasn't adding 4kb?

core/src/aztec/AZDetector.cpp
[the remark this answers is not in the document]
axxel: Fair enough. I respect the effort of making a sound technical argument. :)

core/src/aztec/AZDetectorResult.h
```diff
@@ -19,6 +19,7 @@ class DetectorResult : public ZXing::DetectorResult
 	int _nbLayers = 0;
 	bool _readerInit = false;
 	bool _isMirrored = false;
+	int _runeValue = 0;
```
axxel, line 22: Sorry, I meant this line when I wrote "See AZDetector.h"...

test/samples/aztec-1/rune-1.png
axxel: I usually throw any new image I add to the repository at this site: https://www.iloveimg.com/compress-image

gormster:
all done! should probably flag up that while this patch does also include support for encoding runes, it's mainly done to add it to the battery of Encode/Decode tests, and I haven't really done anything to integrate it into the rest of the project. But it's there to build off for a future patch. (Which I will probably also end up contributing, lol.)

(and it's a good thing I did, because I would never have found the issue with runes 68 and 223 otherwise!)

axxel:
Thanks a lot. About the writer part: I did not look too hard at that, e.g. I did not suggest to use the appropriate 'str -> int' conversion (inverse of the `ToString(runeValue(), 3)` call). Please note that the writing code is going to be deprecated/removed in the long run and replaced by a `libzint` backend. See #332 et al. Bottom line: there is not much point putting effort into this.

That said, it is currently not possible to specify the use a Rune instead of a standard Aztec symbol with the new interface.

gormster:
> Bottom line: there is not much point putting effort into this.

Noted! I will put zero more effort into this. Cheers!

PR #764, opened 2024-04-10
kn: Add support for experimental writer api

ISNing:


ISNing:
I think you must give this a look as I think this is not the best practice, if you have any idea about modeling this, please tell me:
https://github.com/zxing-cpp/zxing-cpp/pull/764/files#diff-522c2c5ffea68705c4dd11a7c6100002687b8fb541c5bc1e84b1da025c6a1932R16-R53
https://github.com/zxing-cpp/zxing-cpp/pull/764/commits/5f1be8fcf44a959338d203a35a7c38c2c90ad082#diff-522c2c5ffea68705c4dd11a7c6100002687b8fb541c5bc1e84b1da025c6a1932R73-R89

axxel:
> I think you must give this a look as I think this is not the best practice, if you have any idea about modeling this, please tell me:

I don't like that either. While `Image` is derived from `ImageView` in c++, this is not the case (or intransparent) in the C-API. While it would make sense to model it like that in k/n as well to support passing the result of a barcode creation directly back to the reader, this is not really useful in non-testing use cases, so not a priority in my eyes. What I definitively not want to merge is a rename of the public facing `ImageView` type to `ImageViewImplNoCopy`. If we deem this subclassing worthwhile then we need to introduce something like `ImageViewBase` or `IImageView` and then have `Image` and `ImageView` implement that.

I did not have much time to really go through the PR in detail, though.

ISNing:
I think the current version makes more sens.

wrappers/kn/src/nativeMain/kotlin/zxingcpp/Barcode.kt
```diff
@@ -53,6 +53,23 @@ fun ZXing_Position.toKObject(): Position = Position(
 
 @OptIn(ExperimentalForeignApi::class)
 class Barcode(val cValue: CValuesRef<ZXing_Barcode>) {
+	companion object {
+		@ExperimentalWriterApi
+		fun fromText(text: String, opts: CreatorOptions): Barcode =
```
axxel, line 58: I'd like to question the use of terms `fromText` and `fromBytes` in this context. Those are required if you have a language which either has no function overloading or you don't have a canonical type for those two options to overload on (that is true for C++, C and Rust, but e.g. not for C#). Here I would suggest to either simply have two (secondary) constructors that overload on `String` vs. `ByteArray` or if that is weird/impossible then maybe just a `from` function. If the constructor approach works, we also don't need the companion object, right?

wrappers/kn/src/nativeMain/kotlin/zxingcpp/BarcodeWriter.kt
```diff
@@ -0,0 +1,93 @@
...
+
+// TODO: Remove this annotation when the API is stable
+@RequiresOptIn(level = RequiresOptIn.Level.ERROR, message = "The Writer API is experimental and may change in the future.")
+@Retention(AnnotationRetention.BINARY)
+annotation class ExperimentalWriterApi
+
+
+@ExperimentalWriterApi
+@OptIn(ExperimentalForeignApi::class)
+class BarcodeWriter : WriterOptions() {
```
axxel, line 23: Looking at this, I'd like to question the utility of the `BarcodeWriter` class. Why not implement the `toSVG`/`toImage` members of `Barcode` directly calling the c-API and completely remove this class? With the above constructor idea the usecase could then look like
```Kotlin
val svg = Barcode("some test", BarcodeFormat.QRCode).toSVG();
```

Could you add something like this to the README.md? I'd like to always start from the client side usability and strive for a native feeling API. See also my current .NET and Rust README.

wrappers/kn/src/nativeMain/kotlin/zxingcpp/ImageView.kt
```diff
@@ -5,42 +5,63 @@
...
 import kotlinx.cinterop.*
 import zxingcpp.cinterop.*
 import kotlin.experimental.ExperimentalNativeApi
+import kotlin.native.ref.Cleaner
 import kotlin.native.ref.createCleaner
 
-@OptIn(ExperimentalForeignApi::class)
+@OptIn(ExperimentalForeignApi::class, ExperimentalNativeApi::class)
 class ImageView(
-	val data: ByteArray,
```
axxel, line 16: I would prefer we left this class as it was before. The `asImageView()` function is only useful for the testing use case. And the test based on that does not really add too much value, since it tests the functionality of the underlying c++ code but that is tested already elsewhere. We need to test if the member accessors of `Image` produce something correct. For that it is more helpful to get that data out and construct a new `ImageView` and then pass that back in.

axxel (changes requested):
Thanks. I think this can be further simplified. See below. Sorry for being a bit slow nowerdays.

wrappers/kn/src/nativeMain/kotlin/zxingcpp/BarcodeWriter.kt
ISNing: I believe it's best to keep the BarcodeWriter class because it serves as a logical place to apply writer options. Moreover, this aligns with Java's style, even though it may not be the most convenient approach.

Regarding the toSVG and toImage methods, they have already been implemented in the code, as seen here:
https://github.com/zxing-cpp/zxing-cpp/pull/764/files#diff-108ceccf4e8d9a00451a5f5cca9c48695f005ddcc552c009530799a244f037a1R152-R157

By providing both methods for converting the barcode, users have the flexibility to choose the one that best suits their needs and preferences.

I will add the usage of `toSVG` and `toImage` later.

wrappers/kn/src/nativeMain/kotlin/zxingcpp/ImageView.kt
ISNing: done

ISNing:
I've checked all the comments and pushed related changes.

Sorry for late responding...

wrappers/kn/README.md
```diff
@@ -40,6 +42,48 @@ barcodeReader.read(image).joinToString("\n") { barcode: Barcode ->
...
+A trivial use case looks like this:
+
+```kotlin
+import zxingcpp.BarcodeFormat
+import zxingcpp.BarcodeWriter
+import zxingcpp.CreatorOptions
+import zxingcpp.ImageFormat
+import zxingcpp.ImageView
+import zxingcpp.toSVG
+import zxingcpp.toImage
```
axxel, line 56: Those symbols don't exist, do they? I believe a `import zxingcpp.Barcode` is missing, though. And that would also be missing in the reading use case above?

Question: did you check that this code compiles? (I did not).

wrappers/kn/README.md
```diff
@@ -40,6 +42,48 @@ barcodeReader.read(image).joinToString("\n") { barcode: Barcode ->
...
+
+val text: String = "Hello, World!"
+val format = BarcodeFormat.QRCode
+val width: Int = 256
+val height: Int = 256
```
axxel, line 61: Those variables are never used...

wrappers/kn/README.md
```diff
@@ -40,6 +42,48 @@ barcodeReader.read(image).joinToString("\n") { barcode: Barcode ->
...
+
+val opts = CreatorOptions().apply {
```
axxel, line 63: Creating a `CreatorOptions` object without a format does not make any sens. I would like to see a constructor for `CreatorOptions` that forces to supply a format.

wrappers/kn/README.md
```diff
@@ -40,6 +42,48 @@ barcodeReader.read(image).joinToString("\n") { barcode: Barcode ->
...
+    format = BarcodeFormat.QRCode
+   // more options, see documentation
+}
+
+val barcode = Barcode(text, opts)
+// or
+val barcode2 = Barcode(text, format)
+
+val barcodeWriter = BarcodeWriter().apply {
```
axxel, line 72: I'm still not convinced that `BarcodeWriter` brings any value here. `WriterOptions` that can optionally be passed to `Barcode.toImage()` seems pretty adequate to me.

wrappers/kn/README.md
```diff
@@ -40,6 +42,48 @@ barcodeReader.read(image).joinToString("\n") { barcode: Barcode ->
...
+   rotate = 10
```
axxel, line 73: rotate can only be multiples of `90`. As a demo property, I'd rather use `sizeHint = 400` as an example, which is also going to be way more common.

wrappers/kn/src/nativeMain/kotlin/zxingcpp/BarcodeWriter.kt
axxel: The logical place to apply the options to is `WriterOptions`. Given my stance that `BarcodeWriter` is useless here, I thought about the reason for the existence of `BarcodeReader`. Why not have a `Barcode` constructor that takes an image? The answer is, the reader returns multiple objects, which a constructor can't.

What do you mean by "Java's style"?

wrappers/kn/src/nativeMain/kotlin/zxingcpp/ImageView.kt
```diff
@@ -5,13 +5,14 @@
...
 
+import cnames.structs.ZXing_Image
 import cnames.structs.ZXing_ImageView
 import kotlinx.cinterop.*
 import zxingcpp.cinterop.*
 import kotlin.experimental.ExperimentalNativeApi
 import kotlin.native.ref.createCleaner
 
-@OptIn(ExperimentalForeignApi::class)
+@OptIn(ExperimentalForeignApi::class, ExperimentalNativeApi::class)
```
axxel, line 15: Why would `ImageView` be suddenly "experimental"?

wrappers/kn/src/nativeMain/kotlin/zxingcpp/ImageView.kt
```diff
@@ -57,3 +59,32 @@ enum class ImageFormat(internal val cValue: ZXing_ImageFormat) {
...
+class Image(val cValue: CValuesRef<ZXing_Image>) {
+	val data: ByteArray
+		get() = ZXing_Image_data(cValue)?.run {
+			readBytes(width * height).also { ZXing_free(this) }
+		}?.takeUnless { it.isEmpty() } ?: throw OutOfMemoryError()
+	val width: Int get() = ZXing_Image_width(cValue)
+	val height: Int get() = ZXing_Image_height(cValue)
+	val format: ImageFormat
+		get() = ZXing_Image_format(cValue).parseIntoImageFormat() ?: error(
+			"Unknown format ${
```
axxel, line 74: This is actually an "internal error" I believe, because there is no user in put that can ever result in this circustance, only a bug in the library implementation.

axxel (changes requested):

wrappers/kn/src/nativeMain/kotlin/zxingcpp/BarcodeWriter.kt
ISNing: Because I understand that kotlin is more suited to java styles (a bunch of controllers, managers, blabla) because it influences the style of code that comes from java, and of course, in my experience, kotlin is more suited to functional programming, so I think both options are fine.

ISNing: Hmm, so why don't we change BarcodeReader to something like `ImageView.readBarcodes(opts)` or sth?
Dont you think that it's wierd there's `Reader` but not a `Writer`?

ISNing: I think maybe we're influenced by the design of original ZXing java

wrappers/kn/src/nativeMain/kotlin/zxingcpp/ImageView.kt
ISNing: fixed in https://github.com/zxing-cpp/zxing-cpp/pull/764/commits/dbc3617c9f82ac53e94a7e2b2a79a32311dbc542

wrappers/kn/README.md
ISNing: Rewritten example code and compiled: https://github.com/zxing-cpp/zxing-cpp/pull/764/commits/a2cea2de74e1e8dd6b9d29812b3f7851e393f710

wrappers/kn/README.md
ISNing: Rewritten example code: https://github.com/zxing-cpp/zxing-cpp/pull/764/commits/a2cea2de74e1e8dd6b9d29812b3f7851e393f710

wrappers/kn/README.md
ISNing: Rewritten example code, the former one has syntax error: https://github.com/zxing-cpp/zxing-cpp/pull/764/commits/a2cea2de74e1e8dd6b9d29812b3f7851e393f710

wrappers/kn/README.md
ISNing: done:

wrappers/kn/README.md
ISNing: done

wrappers/kn/src/nativeMain/kotlin/zxingcpp/ImageView.kt
ISNing: I marked in up rather than remove for it reduce users' effort on debug if there's really something wrong with the implementation: https://github.com/zxing-cpp/zxing-cpp/pull/764/commits/7c4afeafab364428061fc8c0d9b0c61b7c18bfc9

axxel:
Thanks for your continued dedication to maintain your contributed code base. Much appreciated.

PR #776, opened 2024-04-25
fix help message

urishab:
"-formats" should be "-format" in the help message


axxel:
Why do you think so? Note that the option accepts a list of accepted formats.

urishab:
I just tried it in the command line and "-formats" doesn't work.
It has to be "-format" or the reader tries to look for a file called "-formats"

axxel:
Ups. Now I see. I chose to fix the parsing of the argument instead of the help text. Note that `-format` still works as well as any substring between `fo` and `formats`. Thanks for your input.

Issue #772, opened 2024-04-24
demo_cam_reader.html "front"/"back" camera selection does not work


axxel:
<sup>Originally posted by **549393092** April 24, 2024</sup>
Dear,

Could your please check the demo code
https://github.com/zxing-cpp/zxing-cpp/blob/master/wrappers/wasm/demo_cam_reader.html#L135
which will receive parameter "front" and "back" to trigger camera.
However, it seems not work. Instead it works with "user" and "environment". see [MDN doc](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia).

axxel:
@549393092 Would you care to provide a PR with the fix?

549393092:
@axxel I tried to fix this issue, but seems not that simple.
When I select the camera, the updateVideoStream event is triggered, however, it will not switch the camera from front to back.
Meanwhile, when the page is loaded, error occurs in the console when initialize Zxing. The error is 
```
wasm streaming compile failed: TypeError: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.
falling back to ArrayBuffer instantiation
```
Regards

axxel:
Thanks for trying. I currently don't have time to look into this. Maybe @Sec-ant can provide some insights?

549393092:
@Sec-ant Thanks! 
I tried the demo, but still can not switch the camera from front to back. What browser did you use, I am using Edge.

Yes, the initialization error is gone now. Where did you download the "zxing_reader.js" and "zxing_reader.wasm", could your please try the file in this [link](https://github.com/zxing-cpp/zxing-cpp/releases/download/v2.2.0/wasm-artifacts.zip) .

Regards

Sec-ant:
> What browser did you use, I am using Edge.

I'm using the [Kiwi browser](https://github.com/kiwibrowser/src.next/releases/tag/8822657649) on an Android 12 phone, it's a browser based on chromium.

> Where did you download the "zxing_reader.js" and "zxing_reader.wasm"

I didn't download them somewhere, I built the files locally according to [the instructions](https://github.com/zxing-cpp/zxing-cpp/tree/master/wrappers/wasm#build).

> could your please try the file in this [link](https://github.com/zxing-cpp/zxing-cpp/releases/download/v2.2.0/wasm-artifacts.zip) .

[...]

PR #777, opened 2024-04-29
Bugfix: it's not working when switch camera from front to back

549393092:
For more detail, please check #772.


axxel:
Thanks! Also to @Sec-ant!

Issue #612, opened 2023-09-05
MSVC build is broken, no countl_zero and popcount


EchterAgo:
With Visual Studio 2022 I get these errors when building zxing-cpp 2.1.0:

```
  zxing-cpp\core\src\BitHacks.h(41,14): error C2039: 'countl_zero': is not a member of 'std' [zxing-cpp\build\temp.win32-cpython-38\Release\zxingcpp.vcxproj]
  C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.37.32822\include\vector(26,1): message : see declaration of 'std' [zxing-cpp\build\temp.win32-cpython-38\Release\zxingcpp.vcxproj]
  zxing-cpp\core\src\BitHacks.h(84,14): error C2039: 'countr_zero': is not a member of 'std' [zxing-cpp\build\temp.win32-cpython-38\Release\zxingcpp.vcxproj]
  C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.37.32822\include\vector(26,1): message : see declaration of 'std' [zxing-cpp\build\temp.win32-cpython-38\Release\zxingcpp.vcxproj]
  zxing-cpp\core\src\BitHacks.h(153,14): error C2039: 'popcount': is not a member of 'std' [zxing-cpp\build\temp.win32-cpython-38\Release\zxingcpp.vcxproj]
  C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.37.32822\include\vector(26,1): message : see declaration of 'std' [zxing-cpp\build\temp.win32-cpython-38\Release\zxingcpp.vcxproj]
  zxing-cpp\core\src\BitHacks.h(153,14): error C3861: 'popcount': identifier not found [zxing-cpp\build\temp.win32-cpython-38\Release\zxingcpp.vcxproj]
```

This seems to be caused by https://github.com/zxing-cpp/zxing-cpp/commit/9980b2da410c4194578d942cd1d6d4d619a05722 and building version 2.0.0 works.

axxel:
Hallo Axel (du bist erst der 3. andere Axel der mir in meinem Leben über den Weg lief ;)).

1) Since PyPi has pre-build binary packages for windows available, why would you need to build them yourself?
[...]

PR #787, opened 2024-05-18
Fix '/Zc:__cplusplus' leaking into GNU-based compilers on Windows

bindreams:
The change in #612 has introduced a bug when the library is used as a dependency on Windows with the Clang compiler.

For example, when zxing is installed via vcpkg, it is automatically compiled using MSVC and all is well. However, when a user (me) attempts to build a dependent program using a MSVC-compatible compiler with a GNU frontend, such as Clang (not mingw-based clang, the official one), the program fails to compile. This happens because zxing exports `/Zc:__cplusplus` publicly, even though the compiler might not support it. Example compilation error on a dependent project:
```
C:\PROGRA~1\LLVM\bin\CLANG_~1.EXE [...] -Wall -Wextra -pedantic -Werror /Zc:__cplusplus -MD -MT [...]
CLANG_~1: error: no such file or directory: '/Zc:__cplusplus'
```

I was able to fix this by using a generator expression around the flag. This change propagates to `ZXingTargets.cmake` and patching vcpkg to use my fork also seems to fix the issue.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -52,7 +52,7 @@ if (WINRT)
 endif()
 if (MSVC)
     set (ZXING_PUBLIC_FLAGS ${ZXING_PUBLIC_FLAGS}
-        /Zc:__cplusplus
+    $<$<CXX_COMPILER_ID:MSVC>:/Zc:__cplusplus>
     )
 endif()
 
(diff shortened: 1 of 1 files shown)


axxel:
You mean the generator expression ends up in `ZXingTargets.cmake` and hence still gets evaluated from inside your client code cmake project?

bindreams:
Yes, exactly.

bindreams:
From an idealist point of view, a better solution would be to move the code affected by the flag into a cpp file, and make the flag private. Then the issue would go away altogether.

Current approach works only because MSVC does not provide a GNU-style frontend (yet), and Clang's MSVC-style frontend does not need this flag.

axxel:
Thanks for the effort and insight. Removing the public dependency on `BitHacks.h` would work but maybe there is going to be another situation where we need to determine the c++ version in the future and we'd end up with the same issue.

Issue #783, opened 2024-05-12
AndroidManifest.xml:2 Android Build Error


ffreality:
Master branch gives this error when I try to build for Android from both Windows 11 and WSL2 Ubuntu 22.04

**Installed Environment**
Android Studio 2023.3.1
openjdk 21.0.2 2024-01-16
Gradle 8.7
NDK 25.1
Android Studio SDK from Android 8.1 to Android 14

All paths set.

`zxing-cpp/wrappers/android/zxingcpp/build/intermediates/aapt_friendly_merged_manifests/release/aapt/AndroidManifest.xml:2: AAPT: error: attribute 'package' in <manifest> tag is not a valid Android package name: 'zxingcpp'.`

ffreality:
Any suggestion ?

axxel:
I had a look at this and to my surprise, I could indeed reproduce the issue when building a release build via `./gradlew :zxingcpp:assembleRelease` as supposed to a debug build (which is what the CI system does). After some random experiments to determine where exactly the string `zxingcpp` in the autogenerated AndroidManifest.xml comes from, it turns out that is from here:
https://github.com/zxing-cpp/zxing-cpp/blob/d0c1f343e2b17847ec7d26642ed09af643492eaa/wrappers/android/zxingcpp/build.gradle.kts#L9
If you change that string to e.g. `zxingcpp.lib` it builds again. This used to work in the past and I believe some tool chain update broke this. I currently don't have a clear understanding of the implications of changing this `namespace` property with respect to the published android package and any client side code. There is some documentation to be found [here](https://developer.android.com/build/configure-app-module)

@markusfisch do you have an insights in this regard?

markusfisch:
I wouldn't have known this before I checked the docs, but I'm surprised it ever worked without a dot, because according to the documention you mentioned, the `applicationId` "must have at least two segments (one or more dots)". And since `namespace` and `applicationId` [should be the same](https://developer.android.com/build/configure-app-module#change_the_package_name), this also applies to `namespace`. Maybe past build tools didn't check the application ID, but I don't know.

Now, since the `namespace` is only used as package name for the generated `R` and `BuildConfig` classes of a module, and because the wrapper doesn't have any resources, changing it _should_ have no effect for the published packages. And indeed, I can confirm that the `aar` does still have the same package name ("zxingcpp") for its classes even with a different `namespace` ("zxingcpp.lib").

[...]

PR #792, opened 2024-05-30
android: zxingcpp -> io.github.zxingcpp

luk1337:
Fixes the following aapt2 error:
attribute 'package' in <manifest> tag is not a valid Android package name: 'zxingcpp'.


axxel:
As recently discussed [here](https://github.com/zxing-cpp/zxing-cpp/issues/783#issuecomment-2118726507), only the `namespace` property needs to change to fix the issue you seem to have run into as well. I currently don't want to apply your PR as I view as it too invasive/disruptive.

luk1337:
ok, I'll just track the linked discussion.

PR #798, opened 2024-06-18
added CoreZint pod subspec to use Zint with zxing-cpp

farfromrefug:


axxel:
Thanks for your contribution but can you please describe how this helps you in using the new `zint` backend? Without properly setup pre-processor macros `-DZXING_WRITERS=NEW -DZXING_EXPERIMENTAL_API=ON` this has no effect other than bloating up your binary.

farfromrefug:
@axxel without this zint sources are simply not added to the project thus you wont ever be able to build. However you are right you still need to add the correct flags. The flags you mention wont work either are those are cmake flags which are not used with xcode.
The question is then do we add the correct flags automatically or do we let the user choose the ones he wants

Issue #807, opened 2024-07-12
Address Sanitizer (ASAN) reports alloc-dealloc-mismatch with a simple sample application


johngladp:
Firstly, thankyou for the excellent project!

I'm seeing an alloc-dealloc-mismatch reported when running zxing with ASAN enabled.  We're using a cmake toolchain file to inject the -fsanitizer=address and -stdlib=libc++ -nostdlib++ flags.  This simple test application is being built on Linux with clang++ v17.0.6.

```
==16691==ERROR: AddressSanitizer: alloc-dealloc-mismatch (malloc vs operator delete) on 0x52c000000200
    #0 0x55d6880a5add in operator delete(void*) (/--omitted--/out/build/release-linux-asan/built/test_bin/zxing_validation_tests+0x254add) (BuildId: 0027247aec8726a3)
    #1 0x55d6880a8e5c in std::__1::__shared_count::__release_shared[abi:ue170006]() /usr/lib/llvm-17/bin/../include/c++/v1/__memory/shared_ptr.h:172:9
    #2 0x55d6880a8e5c in std::__1::__shared_weak_count::__release_shared[abi:ue170006]() /usr/lib/llvm-17/bin/../include/c++/v1/__memory/shared_ptr.h:213:27
    #3 0x55d6880a8e5c in std::__1::shared_ptr<zint_symbol>::~shared_ptr[abi:ue170006]() /usr/lib/llvm-17/bin/../include/c++/v1/__memory/shared_ptr.h:772:23
    #4 0x55d6880a8e5c in ZXing::Result::~Result() /--omitted--/out/build/release-linux-asan/temp_pkgs/assets/zxing-Linux-x64-RelWithDebInfo-asan-1.0.0/package/include/ZXing/Barcode.h:45:7

0x52c000000200 is located 0 bytes inside of 30512-byte region [0x52c000000200,0x52c000007930)
allocated by thread T0 here:
    #0 0x55d688068978 in calloc (/--omitted--/out/build/release-linux-asan/built/test_bin/zxing_validation_tests+0x217978) (BuildId: 0027247aec8726a3)
    #1 0x55d6882c85bf in ZBarcode_Create packages/zint/src/zint-build/./packages/zint/src/zint/backend/library.c:85:37
[...]

PR #808, opened 2024-07-12
Fix to ensure the custom deallocator for zint_symbol is used for std::shared_ptr's too

johngladp:
Fixes to address issue: #807 

`std::unique_ptr<zint_symbol>` usage is replaced with:
`using unique_zint_symbol = std::unique_ptr<zint_symbol, zint_symbol_deleter>;`

This permits the std::move cast used in 
void Result::zint(unique_zint_symbol&& z)
```
{
	_zint = std::shared_ptr(std::move(z));
}
```
to move the customer deleter usage into the shared pointer too, std::default_delete usage only works for unique_ptr's which is why this change is needed.  This prevents ASAN errors that report alloc/dealloc mismatch due to the shared_ptr usage:
image
(The image is from our internal test pipeline and is of the test that was failing when ASAN is enabled.)

### DIFF
--- a/core/src/Barcode.cpp
+++ b/core/src/Barcode.cpp
@@ -16,6 +16,10 @@
 
 #ifdef ZXING_USE_ZINT
 #include <zint.h>
+void zint_symbol_deleter::operator()(zint_symbol* p) const noexcept
+{
+	ZBarcode_Delete(p);
+}
 #else
 struct zint_symbol {};
 #endif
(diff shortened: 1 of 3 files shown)


axxel:
Thank you very much for debugging and reporting his. Could you do me a favor and drop the gitignore commit from this PR?

And the broken rust-build has nothing to do with your change but an updated rust compiler on the CI system.

johngladp:
gitignore commit removed as requested.  Thanks!

axxel:
Thanks!

Issue #822, opened 2024-08-31
Clang / libc++ build error: `implicit instantiation of undefined template 'std::char_traits<unsigned char>'`


LinuxUserGD:
OS: Gentoo Linux amd64 musl libc (LLVM/Clang profile)

```bash
$ clang --version
clang version 19.1.0+libcxx
Target: x86_64-pc-linux-musl
Thread model: posix
InstalledDir: /usr/lib/llvm/19/bin
Configuration file: /etc/clang/x86_64-pc-linux-musl-clang.cfg
```

Steps to reproduce:
```bash
git clone https://github.com/zxing-cpp/zxing-cpp.git
cmake -S zxing-cpp -B zxing-cpp.release -DCMAKE_BUILD_TYPE=Release
cmake --build zxing-cpp.release -j8 --config Release
```
Build fails with the following output:
```bash
[ 26%] Building CXX object core/CMakeFiles/ZXing.dir/src/ReadBarcode.cpp.o
[ 27%] Building CXX object core/CMakeFiles/ZXing.dir/src/Utf.cpp.o
[ 28%] Building CXX object core/CMakeFiles/ZXing.dir/src/BitArray.cpp.o
[ 29%] Building CXX object core/CMakeFiles/ZXing.dir/src/GenericGF.cpp.o
In file included from <built-in>:3:
In file included from /home/user/git/zxing-cpp.release/core/CMakeFiles/ZXing.dir/cmake_pch.hxx:5:
In file included from /home/user/git/zxing-cpp/core/src/Barcode.h:10:
In file included from /home/user/git/zxing-cpp/core/src/BarcodeFormat.h:9:
In file included from /home/user/git/zxing-cpp/core/src/Flags.h:8:
In file included from /home/user/git/zxing-cpp/core/src/BitHacks.h:12:
In file included from /usr/include/c++/v1/vector:324:
In file included from /usr/include/c++/v1/__format/formatter_bool.h:16:
In file included from /usr/include/c++/v1/__format/concepts.h:16:
In file included from /usr/include/c++/v1/__format/format_parse_context.h:16:
/usr/include/c++/v1/string_view:301:42: error: implicit instantiation of undefined template 'std::char_traits<unsigned char>'
  301 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
      |                                          ^
[...]

Issue #386, opened 2022-08-19
Not able to read multiple DataMatrix symbols in one image


SalmanKhaja:
I am using ZxingReader binary. I am able to reader Multiple QR codes in single image file but for multiple Data Matrix codes, it's reading only one Data Matrix in image. Is this feature is not available or I have to do any work around?

axxel:
Yes there is a workaround to enable that: compile the library in c++20 mode, see https://github.com/nu-book/zxing-cpp/commit/d6dc409ef15ff7171c7f9130dff7266a7c65f053.

SalmanKhaja:
Thanks for help. But I have got following error when trying to Make ZXing configured with Cmake in C++20

```
pi@raspberrypi:~/zxing-cpp.release $ make
[  2%] Built target zueci-static
[  3%] Building CXX object core/CMakeFiles/ZXing.dir/src/TextUtfEncoding.cpp.o
In file included from /home/pi/zxing-cpp/core/src/TextUtfEncoding.cpp:8:
/home/pi/zxing-cpp/core/src/Utf.h:16:28: error: ‘std::__cxx11::wstring ZXing::FromUtf8’ redeclared as different kind of symbol
 std::wstring FromUtf8(std::u8string_view utf8);
                            ^~~~~~~~~~~~~
/home/pi/zxing-cpp/core/src/Utf.h:14:14: note: previous declaration ‘std::__cxx11::wstring ZXing::FromUtf8(std::string_view)’
 std::wstring FromUtf8(std::string_view utf8);
              ^~~~~~~~
/home/pi/zxing-cpp/core/src/Utf.h:16:28: error: ‘u8string_view’ is not a member of ‘std’
 std::wstring FromUtf8(std::u8string_view utf8);
                            ^~~~~~~~~~~~~
/home/pi/zxing-cpp/core/src/Utf.h:16:28: note: suggested alternative: ‘u32string_view’
 std::wstring FromUtf8(std::u8string_view utf8);
                            ^~~~~~~~~~~~~
                            u32string_view
make[2]: *** [core/CMakeFiles/ZXing.dir/build.make:193: core/CMakeFiles/ZXing.dir/src/TextUtfEncoding.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:139: core/CMakeFiles/ZXing.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

axxel:
[...]

PR #836, opened 2024-09-21
Fix libcxx-19 char_traits<uint8_t> build failure

nvinson:
While char_traits<uint8_t> is not officially supported by any C++ standard, gcc and libcxx(<19) have extensions to support the type. The C++20 standard introduces support for char_traits<char8_t>.

Starting with libcxx-19, the extensions to support char_traits<T> where T is not a type specified by a C++ standard has been dropped. See https://reviews.llvm.org/D138307 for details.

This causes zxing-cpp to fail to build if the C++ compiler adheres to the C++17 (or older) and uses libcxx-19 as the C++ runtime library.

The fix is to use char_traits<char8_t> when possible, fallback to char_traits<uint8_t> when not possible, and emit a configure error if neither type is available.

Moreover, this patch changes the preferred standard from C++17 to C++20 as C++20 defines char_traits<char8_t>.

Fixes: #822

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,10 +71,26 @@ if(NOT ZXING_DEPENDENCIES IN_LIST ZXING_DEPENDENCIES_LIST)
 endif()
 
 if (NOT DEFINED CMAKE_CXX_STANDARD)
-    set (CMAKE_CXX_STANDARD 17)
+    set (CMAKE_CXX_STANDARD 20) # Use C++20 standard if available
 endif()
 if (NOT DEFINED CMAKE_CXX_EXTENSIONS)
     set (CMAKE_CXX_EXTENSIONS OFF)
+endif()
+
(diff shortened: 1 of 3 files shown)


axxel:
Thank you for digging into this and providing the background information. I'm not convinced of this 'solution', though. It is basically just a fancier error report compared to the `undefined template` one from clang, but does not fix the problem itself more than simply switching to c++20 does by itself.

I see two options:
1) just switch to c++-20 by default, which would solve @LinuxUserGD's and your issue.
2) change the code such that is actually compiles in c++-17 mode with libcxx-19.

I'm generally for using the 'latest and greatest' tools available and would have to switched to c++-20 by default a while back if it did not have significant performance implications related to [this commit](https://github.com/zxing-cpp/zxing-cpp/commit/d6dc409ef15ff7171c7f9130dff7266a7c65f053). But since this has now been enabled by default in pretty much all the wrappers for a while now and it would effectively close #386, I'll go ahead and do that now.

PR #837, opened 2024-09-22
Update Quadrilateral

MonkeybreadSoftware:
removed a and b being two variables.
a and b were used fro two different variables.

### DIFF
--- a/core/src/Quadrilateral.h
+++ b/core/src/Quadrilateral.h
@@ -151,15 +151,15 @@ bool HaveIntersectingBoundingBoxes(const Quadrilateral<PointT>& a, const Quadril
 }
 
 template <typename PointT>
-Quadrilateral<PointT> Blend(const Quadrilateral<PointT>& a, const Quadrilateral<PointT>& b)
+Quadrilateral<PointT> Blend(const Quadrilateral<PointT>& aa, const Quadrilateral<PointT>& bb)
 {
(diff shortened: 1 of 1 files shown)


axxel:
May I ask what the purpose / intent of this PR is?

MonkeybreadSoftware:
Fixes a compiler warning about two variables a and b with same name and one shadows the other.

axxel:
I see. I personally don't consider variable shadowing to be generally bad. Also the library compiles without any warnings in the default cmake config on my system. Did you explicitly enable that warning in yours? If not: what is your compiler/system you use? I assume this warning is showing up while compiling your client code?

MonkeybreadSoftware:
Yes, we compile code with -Weverything using clang and then disable the warnings we don't like.
And if you include zxing in several files and build for multiple platforms, you see these warnings a lot of times.

axxel:
Fair enough, I see your point. Is this the only annoying warning you get from including zxing-cpp?

MonkeybreadSoftware:
Yes. That is the only warnings we saw. Otherwise it builds fine here.
I used to patch this myself and thought I would contribute this to make it permanent.

axxel:
Thanks for your effort. As you can see, I chose a one-liner to achieve the same effect.

MonkeybreadSoftware:
Thank you.

PR #842, opened 2024-10-04
cmake: allow overriding python install directories

jameshilliard:
We may need to override these variables otherwise the python library may not get installed to the right location.


axxel:
Question: why not simply use `cmake -DCMAKE_INSTALL_PREFIX=...` to direct your installation to where you want it?

jameshilliard:
> Question: why not simply use `cmake -DCMAKE_INSTALL_PREFIX=...` to direct your installation to where you want it?

AFAIU using that would change the install location of the non-python targets which would not be desired.

axxel:
> AFAIU using that would change the install location of the non-python targets which would not be desired.

That is correct. So are you actually installing not only the python module but also the c++-lib / example binaries? I'm curious: what is your use case? Packaging for some distribution?

jameshilliard:
> So are you actually installing not only the python module but also the c++-lib / example binaries?

Yes.

> Packaging for some distribution?

For [buildroot](https://buildroot.org/).

PR #849, opened 2024-10-14
QRCode: add QR data mask to Barcode result

markusfisch:
I forgot to add `dataMask` to `Barcode` as well, so clients can access this member.

### DIFF
--- a/core/src/Barcode.cpp
+++ b/core/src/Barcode.cpp
@@ -47,6 +47,7 @@ Result::Result(DecoderResult&& decodeResult, DetectorResult&& detectorResult, Ba
 	  _position(std::move(detectorResult).position()),
 	  _sai(decodeResult.structuredAppend()),
 	  _format(format),
+	  _dataMask(decodeResult.dataMask()),
 	  _lineCount(decodeResult.lineCount()),
 	  _isMirrored(decodeResult.isMirrored()),
 	  _readerInit(decodeResult.readerInit())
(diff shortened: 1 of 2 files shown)


axxel:
Ahh, I was already wondering how your last PR was helping you... The "problem" is that I was fine with adding this info to this "internal" data structure, precisely because it was not affecting the public API. But the bar for me to add a new member to `Barcode` is _substantially_ higher. I'm afraid, the disussed niche application is not something that I care enough about to justify the maintenance cost across all the wrappers.

This thing brought back the old "MetaData" thingy that the upstream code base had. Maybe I would be more convinced if we added something that allowed to easily add "random" stuff that is only meaningful for certain symbologies, later without changing the API. But it would need to be very simple and generic, like a string containing a JSON object.

What do you think? Are there any other meta-data that you know of which might be interesting to add?

markusfisch:
Good question. I can't think of anything specific at the moment, but there are certainly properties of other symbologies that could be interesting for similar reasons. So I think it's a good idea to transfer such metadata via a generic field so that the maintenance effort remains as low as possible.

Also, the data mask is a very QR Code specific thing, so I can well understand that you would rather not have this element in `Barcode`. I just did it this way because it was the most simplest solution. But a generic string holding a JSON would also be fine!

axxel:
I consider the JSON idea to be the way to go (in one form or another). Update, see [here](https://github.com/zxing-cpp/zxing-cpp/discussions/862#discussioncomment-12765746).

PR #858, opened 2024-10-22
README: fix typos

spageektti:
Hi. This PR fixes typos in the `README.md` file.


axxel:
I believe the additional comma is wrong but overall it is an improvement. :). Thanks.

axxel:
I double checked, and the extra comma is fine but so is the original version (in contrast to the comma in this sentence, which is apparently mandatory ;)).

PR #861, opened 2024-11-01
QR Code: add the possibility to set a fixed version

betarho:
There are use cases where the generated QR Code must have fixed modules, no matter what the text length is.
These patches add the possibility to explicitly set the QR Code version.
It is the user responsibility to guarantee that the text to be encoded has a length that fits into the selected version.

The patches touch both the `MultiFormatWriter` and the `WriteBarcode` "CreateBarcode" methods defined under the ZXing experimental API.
Also the `ZXingWriter` CLI example program has been updated, adding a new option.

### DIFF
--- a/core/src/MultiFormatWriter.cpp
+++ b/core/src/MultiFormatWriter.cpp
@@ -55,11 +55,17 @@ MultiFormatWriter::encode(const std::wstring& contents, int width, int height) c
 		return exec0(std::move(writer));
 	};
 
+	auto exec3 = [&](auto&& writer, auto setEccLevel) {
+		if (_qrVersion > 0 && _qrVersion <= 40)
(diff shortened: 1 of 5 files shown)


axxel:
Thanks for your effort. As laid out in #862, I'm considering an alternative approach for adding symbology specific configuration and result meta-information. You are welcome to join the discussion.

axxel:
There is [progress](https://github.com/zxing-cpp/zxing-cpp/discussions/862#discussioncomment-12765746) on the string-based symbology specific parameter front. If you have `ZXING_EXPERIMENTAL_API=ON` configured, you can call e.g.
```bash
ZXingWriter -size 500 microqrcode "1234" -options "version:2,datamask:0" out.png -v
```

PR #864, opened 2024-11-06
Missing include (for std::nullptr_t)

stbergmann:


axxel:
Thanks. May I ask which compiler complained about that?

stbergmann:
> May I ask which compiler complained about that?

oh, recent LLVM trunk Clang/libc++ (on macOS)

Issue #871, opened 2024-11-25
Android 15 16KB Page Size Support


ccrowell-kr:
[Android 15 adds support for devices with 16KB page sizes](https://developer.android.com/about/versions/15/behavior-changes-all#16-kb)

Without support, devices with a 16KB page size will not be able to run apps using the Android wrapper. [This can be verified locally by creating a 16KB emulator](https://developer.android.com/guide/practices/page-sizes#test) and running the demo app

[To support this, there are only 2 requirements:](https://developer.android.com/about/versions/15/behavior-changes-all#16-kb)
1. AGP version 8.5.1 or higher
    - The project is currently on 8.9
2. NDK 28 or NDK 27 with the linker flag `APP_SUPPORT_FLEXIBLE_PAGE_SIZES` enabled
    - This is the only thing that needs to change. I verified this locally by uncommenting the NDK version in the Android wrapper `build.gradle.kts` and updating it to the [latest release](https://github.com/android/ndk/releases/tag/r28-rc1)
    - It's worth noting that NDK 28 is only in RC
    
    
I'm happy to open a PR that does the same. Is there a reason NDK version is commented out?

axxel:
I added to NDK version line to force a specific version with sufficient c++20 support in the past. when this or a newer version became the default, I removed the line. The idea being that people might want to use a specific version for whatever reason. So using the latest release (27?) should be fine if that is required (I trust your analysis at this point ;)). PR is welcome.

axxel:
Can this now be closed with the merging of your PR?

ccrowell-kr:
Yes, thank you! I expected it to be closed by referencing it in the PR, but looks like that didn't happen.

axxel:
If you would have written "fixed" or "fixing" or "fix for" instead of "addresses", it would have automatically closed the issue.

PR #872, opened 2024-11-26
android: add linker flag to support flexible page sizes in Android 15

ccrowell-kr:
This enables the Android wrapper to be used in apps running Android 15 on devices with a 16KB page size.

I have tested this locally with a 16KB emulator. Without the flag, the demo app crashes immediately, with the new flag, the demo app works as expected.

Addresses #871

### DIFF
--- a/wrappers/android/zxingcpp/build.gradle.kts
+++ b/wrappers/android/zxingcpp/build.gradle.kts
@@ -20,7 +20,12 @@ android {
         }
         externalNativeBuild {
             cmake {
-                arguments("-DCMAKE_BUILD_TYPE=RelWithDebInfo", "-DANDROID_ARM_NEON=ON", "-DZXING_WRITERS=OFF")
+                arguments(
+                    "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
+                    "-DANDROID_ARM_NEON=ON",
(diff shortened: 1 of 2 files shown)


wrappers/android/zxingcpp/build.gradle.kts
```diff
@@ -116,6 +121,10 @@ publishing {
 }
 
 signing {
+    setRequired {
```
ccrowell-kr, line 124: Happy to remove this or change the implementation. I added it because it allows us to run `./gradlew publishToMavenLocal` and use the `mavenLocal()` repository rather than manually including the aar for testing.

I borrowed the implementation from this [discussion on the gradle forums](https://discuss.gradle.org/t/how-to-only-enable-the-signing-plugin-if-were-about-to-release/39555)

wrappers/android/zxingcpp/src/main/AndroidManifest.xml
```diff
@@ -1,2 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
ccrowell-kr, line 1: The empty manifest is not required now that namespace is declared in the `build.gradle.kts` but happy to restore this if desired.

axxel:
Thanks a lot and sorry for the delayed response.

PR #877, opened 2024-12-23
Handle ZXingError class in iOS wrapper

GUIEEN:
The `ZXing::Error` class does not inherit from `std::exception`, causing the app to crash when this error is thrown while using the iOS wrapper. I’m not sure why it doesn’t inherit from `std::exception`, but other wrappers in other languages handle these exceptions.

- android
  - https://github.com/zxing-cpp/zxing-cpp/blob/master/wrappers/android/zxingcpp/src/main/cpp/ZXingCpp.cpp#L259-L261
- python
  - https://github.com/zxing-cpp/zxing-cpp/blob/master/wrappers/python/zxing.cpp#L117-L119

### DIFF
--- a/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
+++ b/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
@@ -143,6 +143,9 @@ ZXIGTIN *getGTIN(const Result &result) {
              ];
         }
         return zxiResults;
+    } catch (const ZXing::Error& e) {
+        SetNSError(error, ZXIReaderError, ZXing::ToString(e).c_str());
+        return nil;
     } catch(std::exception &e) {
         SetNSError(error, ZXIReaderError, e.what());
         return nil;
--- a/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
+++ b/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
@@ -143,11 +143,11 @@ ZXIGTIN *getGTIN(const Result &result) {
              ];
         }
         return zxiResults;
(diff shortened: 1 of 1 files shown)


axxel:
> I’m not sure why it doesn’t inherit from std::exception, but other wrappers in other languages handle these exceptions.

This question got me baffled for a moment. I thought, this is an excellent question and why the hell did I not inherit `Error` from `std::exception`?!? After looking through the use cases of `Error` and my memory, I remembered what was behind this decision. To not forget again and for others to be detectable, I decided to add a comment to the `Error.h` file.

Now to your specific change here: As described in my mentioned comment, `Error` objects are not meant to be thrown from inside `ReadBarcodes`, if they are, then this should be considered a bug in the library. Did you encounter such a case prior to introducing #878 ?

Note that the two other implementations you linked to are not catching `Error` objects but simply anything.

GUIEEN:
> Note that the two other implementations you linked to are not catching Error objects but simply anything.

You're right. Those are catching everything, but basically, they prevent the app from crashing in the case of a `ZXing::Error` being thrown.

> if they are, then this should be considered a bug in the library. Did you encounter such a case prior to introducing https://github.com/zxing-cpp/zxing-cpp/pull/878 ?

Yes, it’s difficult to reproduce. It happens randomly, but the iOS app I'm working on crashes due to a `ZXing::Error` and an assertion failure.

axxel:
As described in the comment, the library is not supposed to throw `Error` objects up the stack into user/wrapper code. Therefore, I'd consider it to be wrong/misleading if we merged this PR as is. I see a point in adding a catch-all clause, to prevent any crash due to unhandled exceptions. Would you be interested in modifying your PR accordingly?

As for the assertions, they (should) disappear in a release built.

GUIEEN:
Sure. I updated the code to catch any thrown objects.

PR #878, opened 2024-12-23
Convert assertion to throwing error

GUIEEN:
```
logged res value at crash
res = 10449134607534
```

This assertion failure caused rare crashes, so I’m replacing it with error throwing instead.

### DIFF
--- a/core/src/oned/ODDataBarReader.cpp
+++ b/core/src/oned/ODDataBarReader.cpp
@@ -133,7 +133,7 @@ static std::string ConstructText(Pair leftPair, Pair rightPair)
 	auto res = 4537077LL * value(leftPair) + value(rightPair);
 	if (res >= 10000000000000LL) { // Strip 2D linkage flag (GS1 Composite) if any (ISO/IEC 24724:2011 Section 5.2.3)
 		res -= 10000000000000LL;
-		assert(res <= 9999999999999LL); // 13 digits
+		throw FormatError("res is over than 9999999999999LL"); // 13 digits
 	}
 	auto txt = ToString(res, 13);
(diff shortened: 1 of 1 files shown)


axxel:
Would you be able to provide a sample image that causes this assert to fail for me to understand what is going on? I'd like to determine where the problem originates and then potentially fix it at the root instead of where it shows.

axxel:
To motivate my request: this assert assumes that there is a bug in the library if it evaluates to false, not an issue with the input the library is processing. I'd like to first find out whether or not this assumption is correct and then apply the fix at the correct position.

GUIEEN:
> Would you be able to provide a sample image that causes this assert to fail for me to understand what is going on?

Since it happens very randomly, I'm not sure I could capture the image and supply it. I rarely encounter this issue while running the app for a long time (~20 minutes). The code I'm testing is basically the same as the demo app. Test device is iPad Pro (4th generation), no position change.

axxel:
If you could add a means to save the `imageView` data as an image in your new `catch` block inside `ZXIBarcodeReader`, then simply let your app run until you hit the issue. Then send me that image data, if possible. That would be helpful.

GUIEEN:
Sure, I will try to capture the camera image at the crash point when I have the time to test and reproduce this problem.

axxel:
Since I am about to make a new release, I had another look at the code and came to the gut feeling that there might be some input where this assert would have triggered without there being a bug in the code. So I decided to do something about the crash you saw without a full understanding of what caused it.

I'd still be interested in the sample mentioned above but don't spend more time on it than you can comfortably spare.

PR #881, opened 2024-12-31
Add version functions

MonkeybreadSoftware:
We like to have version functions, so we can query at runtime the version of zxing and display in GUI.

I thought I could contribute this for both C and C++ API.
For C++ I put it into the Reader class.

### DIFF
--- a/core/src/ReadBarcode.cpp
+++ b/core/src/ReadBarcode.cpp
@@ -15,6 +15,8 @@
 #include "MultiFormatReader.h"
 #include "Pattern.h"
 #include "ThresholdBinarizer.h"
+#include "Reader.h"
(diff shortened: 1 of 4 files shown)


axxel:
Thanks for your contribution. I have a couple of issues with your solution, though (e.g. `ZXVersion.h` is deprecated, `Reader` is not the right place to put the function, etc). And since it is more effort to communicate my preferred solution than to simply implement it, I'll do that instead. I hope you don't mind.

MonkeybreadSoftware:
That's fine. So I triggered it at least :-)

Thank you.

axxel:
Please let me know if this satisfies your requirements.

PR #889, opened 2025-01-10
Changed thread_local to ZX_THREAD_LOCAL

MonkeybreadSoftware:
We got a define for ZX_THREAD_LOCAL to turn it on/off and we should use it.

### DIFF
--- a/core/src/GlobalHistogramBinarizer.cpp
+++ b/core/src/GlobalHistogramBinarizer.cpp
@@ -8,6 +8,7 @@
 
 #include "BitMatrix.h"
 #include "Pattern.h"
+#include "ZXConfig.h"
 
 #include <algorithm>
 #include <array>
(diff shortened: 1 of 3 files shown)


axxel:
Question: what is your motivation for this change? Do you need to "disable" this, meaning switching to `static` in your build? If so, why?

Background: this is mostly a historic relic, introduced because the macOS llvm from a couple of years ago did not support `thread_local`. From a thread-safety and usability point of view, `thread_local` is mandatory at the places your just changed. In fact, it would probably be best to remove all current occurrences of `ZX_THREAD_LOCAL` with `thread_local`.

MonkeybreadSoftware:
Currently we have thread_local causing a dependency on the VCRUNTIME140_1.dll due to the tls implementation in the VS 2019 runtime.
So I set ZX_THREAD_LOCAL to empty for Windows 64-bit Intel only to prevent this.
And of course then I need to use ZX_THREAD_LOCAL everywhere instead of  thread_local directly.

axxel:
I see. Thanks for the info. I'll merge this then. Could you please fix the compile (see below) and also add a line to the comment in `ZXConfig.h` outlining your use case for this (so I don't forget)?

The build is failing as there seems to be a `#include "ZXConfig.h"` missing in `ZXingC.cpp`.

MonkeybreadSoftware:
Thanks for considering adding it. I added the include and a comment. Let's see if the CI works.

axxel:
I squashed the merge, since all those changes are semantically one unit from my point of view. Thanks for sharing.

PR #902, opened 2025-02-11
[Python] Support using system CMake

mgorny:
Add `cmake` PyPI dependency only if the program is not found, and use
the system version instead.  This avoids unnecessary dependencies
on third-party binary packages, and improves portability by using
downstream-patched CMake version.

### DIFF
--- a/wrappers/python/pyproject.toml
+++ b/wrappers/python/pyproject.toml
@@ -3,7 +3,6 @@ requires = [
     "setuptools>=42",
     "setuptools_scm",
     "wheel",
-    "cmake>=3.15",
     "pybind11[global]",
 ]
 build-backend = "setuptools.build_meta"
--- a/wrappers/python/setup.py
+++ b/wrappers/python/setup.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import shutil
 import subprocess
 import sys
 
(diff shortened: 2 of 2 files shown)


axxel:
So you are implying that an existing (system) cmake installation can't possibly fulfill the current `requires` requirements?

mgorny:
> So you are implying that an existing (system) cmake installation can't possibly fulfill the current `requires` requirements?

Yes, and also yes. Firstly, all the common builders (such as `pip` or `uv`) default to using build isolation, so it won't even consider the system CMake installation and instead install another copy of CMake locally. Secondly, even if you forced `--no-build-isolation`, then the regular CMake installation does not provide the recognized `cmake` Python package, so it would be determined to be missing, e.g.:

```
$ python -m build -w
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - cmake>=3.15
  - pybind11[global]
  - setuptools>=42
  - setuptools_scm
  - wheel
[…]
$ python -m build -wn
* Getting build dependencies for wheel...
running egg_info
writing zxing_cpp.egg-info/PKG-INFO
writing dependency_links to zxing_cpp.egg-info/dependency_links.txt
writing top-level names to zxing_cpp.egg-info/top_level.txt
reading manifest template 'MANIFEST.in'
writing manifest file 'zxing_cpp.egg-info/SOURCES.txt'

ERROR Missing dependencies:
	pybind11[global]
	pybind11-global==2.13.6
	cmake>=3.15
```

axxel:
Thanks for the info. Then I see one potential regression of your approach: if the installed cmake version on your system is too old, the build will fail. Can you maybe add a quick check to your `shutil` line to make sure the available cmake is also sufficiently new?

mgorny:
Okay, it should be good now. I'm calling `cmake -E capabilities` to get the version number in machine-readable form, and the code adds `cmake` dependency if anything looks amiss: `cmake` is missing, doesn't return correct capabilities or the version is too old.

axxel:
Thanks!

mgorny:
Thanks!

PR #917, opened 2025-03-24
AZDecoder: add flag `haveFNC1` and only erase `<GS>` if set.

### LINKED ISSUE #916: The first byte is lost when recognizing the Aztec code
When decoding some Aztec codes, the first byte is lost. The sample image contains an array of eight bytes (HEX): 
`1D 55 10 94 E2 E7 14 FF`
However, the decoding result is seven bytes: 
`55 10 94 E2 E7 14 FF`

Image

gitlost:
Fixes #916 The first byte is lost when recognizing the Aztec code

### DIFF
--- a/core/src/aztec/AZDecoder.cpp
+++ b/core/src/aztec/AZDecoder.cpp
@@ -249,12 +249,13 @@ static StructuredAppendInfo ParseStructuredAppend(ByteArray& bytes)
 	return sai;
 }
 
-static void DecodeContent(const BitArray& bits, Content& res)
+static void DecodeContent(const BitArray& bits, Content& res, bool &haveFNC1)
 {
 	Table latchTable = Table::UPPER; // table most recently latched to
 	Table shiftTable = Table::UPPER; // table to use for the next read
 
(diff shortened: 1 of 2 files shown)


core/src/aztec/AZDecoder.cpp
```diff
@@ -321,24 +324,24 @@ DecoderResult Decode(const BitArray& bits)
 
 	StructuredAppendInfo sai = haveStructuredAppend ? ParseStructuredAppend(res.bytes) : StructuredAppendInfo();
 
-	// As converting character set ECIs ourselves and ignoring/skipping non-character ECIs, not using
-	// modifiers that indicate ECI protocol (ISO/IEC 24778:2008 Annex F Table F.1)
-	if (res.bytes.size() > 1 && res.bytes[0] == 29) {
```
axxel, line 326: Any particular reason why you dropped that comment. Was there anything wrong with it?

gitlost:
It was made out-of-date by the new `Content` and `SymbologyIdentifier` processing which do add ECI protocol modifiers.

PR #918, opened 2025-03-25
AZDecoder: fix `ParseStructuredAppend()` when have ECIs

gitlost:
As noted in `AZDecoderTest::SymbologyIdentifier()` maybe makes sense if `Result::symbologyIdentifier()` were changed to account for ECI, i.e. `toString()` -> `toString(hasECI())`.

### DIFF
--- a/core/src/Content.cpp
+++ b/core/src/Content.cpp
@@ -81,7 +81,7 @@ void Content::erase(int pos, int n)
 	bytes.erase(bytes.begin() + pos, bytes.begin() + pos + n);
 	for (auto& e : encodings)
 		if (e.pos > pos)
-			pos -= n;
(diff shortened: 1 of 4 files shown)


core/src/Content.cpp
```diff
@@ -81,15 +81,15 @@ void Content::erase(int pos, int n)
 	bytes.erase(bytes.begin() + pos, bytes.begin() + pos + n);
 	for (auto& e : encodings)
 		if (e.pos > pos)
-			pos -= n;
+			e.pos -= n;
```
axxel, line 84: Wow, that is stunning. 3 years and this has not surfaced, yet. Thanks for fixing this total fail.

PR #923, opened 2025-03-29
Refactor include path

fnadeau:
Follow up on discussion #921

Include path now have ZXing prefix. Order of include was not touch, so it is a bit messy in some files. Running clang-format yield more changes to the code than to the header sorting. If it's required, I'll do it, but I believe a full clang-format should be done has there are a lots of changes unrelated to header sorting.  This could be done in a separate commit as part of v3.0 release.


axxel:
Thanks for your efforts but maybe there was a misunderstanding. My suggestion was to simply remove the `-I${includedir}/ZXing` in `zxing.pc.in`. The proposed change here is not at all what I'd like to move forward with.

PR #924, opened 2025-03-29
Remove ZXing path from include path

fnadeau:
As per discussion #921 

Tested with sample from README and other project.

### DIFF
--- a/core/zxing.pc.in
+++ b/core/zxing.pc.in
@@ -5,4 +5,4 @@ Name: ZXing
 Description: ZXing-C++ library
 Version: @PROJECT_VERSION@
 Libs: -L${libdir} -lZXing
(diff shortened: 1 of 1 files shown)


axxel:
Thanks for your work on this. I'll let the PR sit here for the time being and merge it prior to the 3.0 release, which will likely be the next, but I'm not sure, yet.

morenohot18 (commented):
Usa pkg-config para validar que el archivo .pc funcione correctamente: pkg-config --cflags zxingEsto debería devolver las banderas de compilación correctas. Por ejemplo: -I/usr/include

axxel:
@morenohot18 please follow the implicit policy of posting only in the English language. Otherwise your contribution is likely going to be ignored.

PR #933, opened 2025-04-15
Update gitignore

Xie-Long-0:
Ignore .vs and .vscode files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,10 @@ CMakeLists.txt.user
 
 # JetBrain (AndroidStudio, clion)
 .idea
+
+# Visual Studio
+/.vs
+/CMakeUserPresets.json
+
+# Visual Studio Code
+/.vscode


.gitignore
```diff
@@ -17,3 +17,15 @@ CMakeLists.txt.user
...
+
+# Visual Studio
+/.vs
+
+# Visual Studio Code
+/.vscode
+
+# CMake
+/build
+/out
```
axxel, line 29: Where would that directory come from? What is it supposed to contain?

.gitignore
```diff
@@ -17,3 +17,15 @@ CMakeLists.txt.user
...
+/CMakePresets.json
```
axxel, line 30: Please remove that file. If present, this is expressly meant to be shared via git.

axxel (changes requested):
It seems you did not follow the line endings style. Please make sure your lines end the way as all others in this file (unix style line endings).

.gitignore
Xie-Long-0: I'm so sorry, this `/build` directory is the output directory I specified, it shouldn't appear here, I will delete it.

.gitignore
Xie-Long-0: These files `out/` `CMakePresets.json` `CMakeUserPresets.json` are automatically generated by opening the project in CMake mode in VS.
The `out/` is the compilation output directory, I will change it to somewhere else. But `CMakePresets.json` bothers me, I can't delete it, and it seems to be automatically generated by the Qt plugin.
```json
// CMakePresets.json
{
  "version": 3,
  "configurePresets": [
    {
      "hidden": true,
      "name": "Qt",
      "cacheVariables": {
        "CMAKE_PREFIX_PATH": "$env{QTDIR}"
      },
      "vendor": {
        "qt-project.org/Qt": {
          "checksum": "wVa86FgEkvdCTVp1/nxvrkaemJc="
        }
      }
    }
  ],
  "vendor": {
    "qt-project.org/Presets": {
      "checksum": "67SmY24ZeVbebyKD0fGfIzb/bGI="
    }
  }
}
```

Issue #743, opened 2024-03-08
Qt 6.6.1 Android has no Video but some output errors


LeonnardoVerol:
Hi, we meet again

Error Print (even with permission granted)
![image](https://github.com/zxing-cpp/zxing-cpp/assets/6124410/6d044449-72b0-4a1a-b9b9-ccdc56ff8995)

Version Used 1: Master Branch as of 2023y/10m/27d
Version Used 2: Master Branch as of 2024y/03m/07d

QT Kit: Android QT 6.6.1 Clang x86_64 Debug
Emulator Device: Galaxy Nexus API 31 x86_64
JDK: 17
Other Settings:
![image](https://github.com/zxing-cpp/zxing-cpp/assets/6124410/425bebcc-9aa2-4e7b-a1c0-985d5811e8a8)
 
---

**My Guess: QT 6.5+ Bug. Because they changed how permissions work**

![image](https://github.com/zxing-cpp/zxing-cpp/assets/6124410/625f3642-5341-44c4-a836-d412148bc29d)

- https://www.qt.io/blog/qt-6.5-lts-released
- https://doc-snapshots.qt.io/qt6-6.5/permissions.html
- [Since Qt 6.6.1] https://doc.qt.io/qt-6/qml-application-permissions.html

Extra Code: 

```
import QtCore
```

```
            CameraPermission {
                id: cameraPermission
            }

            ColumnLayout {
                anchors.left: parent.left
                anchors.bottom: parent.bottom

                Text {
                    text: qsTr("Status: %1").arg(cameraPermission.status === Qt.PermissionStatus.Undetermined ? qsTr("Undetermined")
                                       : cameraPermission.status === Qt.PermissionStatus.Denied ? qsTr("Denied")
                                       : qsTr("Granted"))
                }

                Button {
                    visible: cameraPermission.status !== Qt.PermissionStatus.Granted
                    text: qsTr("Authorize")

                    onClicked: cameraPermission.request()
                }
            }
```

Example:

![image](https://github.com/zxing-cpp/zxing-cpp/assets/6124410/d40a25b7-de88-44bb-8ce3-1386782d87fa)

---

Error Output In Details:

```
W ImageReader_JNI: Unable to acquire a buffer item, very likely client tried to acquire more than maxImages buffers
W ImageReader_JNI: Unable to acquire a buffer item, very likely client tried to acquire more than maxImages buffers
[...]

Issue #75, opened 2019-06-28
Qt wrapper


swex:
Hi there!
Thank you for great project!
I've made wrapper for Qt based applications: https://github.com/swex/QZXingNu
If somebody interested feel free to use it and improve.

axxel:
+1 :-)

May I ask you the following questions (out of curiosity)?:
 * I guess you are aware of the "officially referenced" Qt-Wrapper based on the original/old c++ port (https://github.com/ftylitak/qzxing)?
 * How does your wrapper compare theirs?
 * What made you invest in this work instead of using their existing one?
 * Have you approached them whether they would be interested in joining forces?

swex:
1. Yes I know about qzxing, I've made my own, based on your zxing rewrite when I failed to get qzxing to work it was about a year time ago [see](https://github.com/ftylitak/qzxing/issues/86). And about the same time I discovered your rewrite and implemented new wrapper. btw upstream fix merged 7 days ago.
2. No I don't think so. I'd be happy to Qt community to have one barcode encode/decode library to rule them all. I'm just not happy with amount of legacy code in original zxing
3. See 1
4. Kinda, https://github.com/ftylitak/qzxing/issues/81#issuecomment-504464955 let me make it here: @ftylitak please replace old zxing with one from this repo. I've already made it, feel free to take necessary code parts to your repo

ftylitak:
Hello @swex  and @axxel 

I too believe that there should be a unified solution with the C++ part of ZXing, no argue. 

Though I have got some concerns on which I am open for discussion. 

There are a couple of implementations of C++ ZXing. To name a few:
* [nu-book/zxing-cpp](https://github.com/nu-book/zxing-cpp)
* [glassechidna/zxing-cpp](https://github.com/glassechidna/zxing-cpp)
* [ftylitak/qzxing](https://github.com/ftylitak/qzxing) (this is added taking into consideration the ZXing C++ part leaving aside the Qt specific parts)

To my understanding, each of the implementations/approaches started with a specific goal from the creators which I appreciate because they are all valid. For example, it is very nice to see the fresh and state of the art implementation of **nu-book/zxing-cpp** with C++14.  Meanwhile, it is very acceptable to see **glassechidna/zxing-cpp** being a direct continuation of ZXing C++. This was the goal of QZXing as well, even though it remained more targeted to Qt. 

[...]

PR #938, opened 2025-04-21
Prevent image buffer exhaustion on Android using Qt

m7913d:
This commit fixes the following exception:
java.lang.IllegalStateException: maxImages (10) has already been acquired, call #close before acquiring more.

This exception is thrown when the device cannot process the images fast enough.

Note that Qt captures the images in a separate (Java) thread (on Android) and sends them asynchronously to the main (GUI) thread.

This fix moves the processing to a separate thread and discards all new images while another image (frame) is still being processed. This allows Qt to close the Android resources before the buffer is full.

In recent versions of Qt, Qt catches this exception itself (see https:// bugreports.qt.io/browse/QTBUG-116526) by restarting the camera, but frequent restarts may degrade performance.

On non-Android platforms, this fix has the advantage that the GUI remains responsive even if the barcode processing is slow, and that always the most recently captured frame is processed.

Related to #743

### DIFF
--- a/example/ZXingQtReader.h
+++ b/example/ZXingQtReader.h
@@ -16,8 +16,11 @@
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
 #include <QAbstractVideoFilter>
 #else
+#include <QMutex>
+#include <QThread>
 #include <QVideoFrame>
 #include <QVideoSink>
+#include <QWaitCondition>
 #endif
 #include <QElapsedTimer>
 #endif
@@ -307,6 +310,47 @@ inline Barcode ReadBarcode(const QVideoFrame& frame, const ReaderOptions& opts =
 	return !res.isEmpty() ? res.takeFirst() : Barcode();
 }
 
+class BarcodeReader;
+
+class BarcodeProcessor : public QThread
+{
+	BarcodeReader& _reader;
+	QVideoFrame _nextFrame;
+	QMutex _mutex;
+	QWaitCondition _wait;
+public:
+	BarcodeProcessor(BarcodeReader& reader) : _reader(reader) {}
+
+	void start() {QThread::start();}
+	void stop()
+	{
+		requestInterruption();
+		_wait.wakeOne();
+		wait();
+	}
+
+	void setNextFrame(const QVideoFrame& frame)
+	{
+		QMutexLocker l(&_mutex);
+		_nextFrame = frame;
+		_wait.wakeOne();
+	}
+
+private:
+	virtual void run() override;
+
+	QVideoFrame takeNextFrame ()
+	{
+		QMutexLocker l(&_mutex);
+		if (!_nextFrame.isValid())
+			_wait.wait(&_mutex);
+
+		QVideoFrame f = _nextFrame;
+		_nextFrame = QVideoFrame();
+		return f;
+	}
+};
+
 #define ZQ_PROPERTY(Type, name, setter) \
 public: \
 	Q_PROPERTY(Type name READ name WRITE setter NOTIFY name##Changed) \
@@ -333,7 +377,15 @@ public:
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
 	BarcodeReader(QObject* parent = nullptr) : QAbstractVideoFilter(parent) {}
 #else
-	BarcodeReader(QObject* parent = nullptr) : QObject(parent) {}
+	BarcodeReader(QObject* parent = nullptr) : QObject(parent), _processor(*this)
+	{
+		_processor.start();
+	}
+	~BarcodeReader()
+	{
+		_processor.stop();
+	}
+
 #endif
 
 	// TODO: find out how to properly expose QFlags to QML
@@ -373,10 +425,11 @@ public:
 	ZQ_PROPERTY(bool, isPure, setIsPure)
 
 	// For debugging/development
-	int runTime = 0;
+	QAtomicInt runTime = 0;
 	Q_PROPERTY(int runTime MEMBER runTime)
 
 public slots:
+	// Function should be thread safe, as it may be called from a separate thread.
 	ZXingQt::Barcode process(const QVideoFrame& image)
 	{
 		QElapsedTimer t;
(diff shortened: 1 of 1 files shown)


m7913d:
Note that this PR only fixes the Qt6 implementation. However, I assume Qt5 is affected too by this bug. I'm not sure whether we should still put some effort in Qt5 as Qt5 will be EoS very soon.

axxel:
Thanks for your contribution. I'd like to suggest though, to not 'force' the use of a multi-threaded approach on everyone using that header file. What about adding some pre-processor macro like `ZXING_QT_READER_MULTITHREADING` that needs to be defined before including the header to enable this feature?

m7913d:
I'm not really against such an approach, but

 - on Android, one should (almost) always use the multi-threaded approach, unless one can ensure the software will only run on hardware that is able to process the barcodes as fast as the frame rate. This is almost impossible to guarantee, as other processes may be running on the device and the barcode processing time itself is not constant (I think).
 - in general, I think it is good practice to avoid running CPU intensive tasks on the GUI thread to keep the GUI responsive at all times. An unresponsive GUI results in a bad user experience. If the user device is busy (with another task), I'm ok with a slower barcode processing, but I would expect the GUI itself being responsive.
 - I wonder why one would want to disable the multi-threading approach. Do you think they are using a device with only a single core? Even cheap smartphones has multi-core processors. Do you want to save battery? I think the frame-skipping feature of the multi-threaded implementation may save more battery than the small overhead of another thread. I wonder if anyone is using Qt / `QVideoSink` on a platform for which the single threaded approach is desirable.

Therefore, I would rather add an option to disable the multi-threaded approach than expecting the user to enable the multi-threaded approach (especially on Android) or just enable the multi-threaded approach always. Note that for rare cases, one can also call `BarcodeReader::Process` himself.

What do you prefer?

axxel:
Alright. First up: it is actually nice to see that someone takes any interest in this Qt wrapper stuff. Until now, it seemed hardly anyone is using that at all.

Second: I agree with all your points above. My hesitation was mainly caused by a lack of active memory what this code I wrote years ago actually does, plus I didn't properly look at what your change was doing. I somehow thought you made the `BarcodeReader::process()` slot mandatorily switching the thread context.

After freshening up my memory, I see the situation as follows. `BarcodeReader` currently serves 3 use cases:
1. provide the ReaderOptions API for initial configuation
2. the `process(QVideoFrame)` slot for synchronously scanning one frame for barcodes
3. the `setVideoSink()` function to setup contiguous scanning

I agree that the 3rd use-case should be asynchronous by default, and also be limited to processing only the most recent frame(s) by default. This gives the best out-of-the-box user experience. With that established, I wonder whether there is room for improving your PR by maybe making use of a `QThreadPool`? Ether the "global" application wide one or an instance owned by the `BarcodeReader`. This should then allow to be more flexible with respect to how many frames might be processed in parallel. This might be overkill, but I wanted to float the idea. Either way, I'd like to propose to look into doing away with the `BarcodeProcessor` class, at least not make it a publicly visible entity, and merge it's functionality with the `BarcodeReader`.

If you would look into how this could be implemented neatly, that would be much appreciated.

m7913d:
Concerning the interest in this repo by the Qt community, I think a lot of them are still using `QZXing` (based on the original zxing repo). It may be interesting to better highlight that zxing-cpp supports Qt too (and probably has a more performant barcode parser under the hood).

Using `QThreadPool` is an option too, although I think that (by default) a single processing thread is a good trade off between performance and not exhausting all resources (memory / cpu / battery) or maybe at most two by default to ensure at least one frame is being processed all the time in case the processing of a frame is slightly slower than the frame rate.

Concerning the hiding of implementation details, is it your intention to keep the Qt wrapper header-only?

axxel:
> I think a lot of them are still using `QZXing`

In case you are interested in ancient history, you might want to have a look at #75. I had another quick glance at the repo and it seems the project can safely be considered as abandoned.

> It may be interesting to better highlight that zxing-cpp supports Qt too

That is true. I was thinking about mentioning Qt in the "wrappers" section of the readme for quite some time. But with the feeling that this is not really used, I did not want to set something in stone. The example code I can change however I see fit without worrying about backward compatibility.

> (and probably has a more performant barcode parser under the hood).

Oh, it definitively does ;).

> Using `QThreadPool` is an option too, although I think that (by default) a single processing thread is a good trade off

Sure, the default should be 1. My point was to make it flexible.

> Concerning the hiding of implementation details, is it your intention to keep the Qt wrapper header-only?

Not necessarily, if there is a clear advantage of doing something else. The header-only approach makes it trivial to build and distribute the wrapper. If we'd consider the `ZXingQtReader.h` to be 'releasable', I could simply add it to the core package and everyone could use it out of the box without any build dependency on Qt for the library itself.

m7913d:
@axxel I changed the implementation to use a `QThreadPool` with 1 thread by default and removed the `BarcodeProcessor` class.

axxel:
That looks really neat. One thing though: I fear `~BarcodeReader()` could dead-lock if `maxThreadCount() >= 2` and the VideoSink is feeding new frames faster than they can be processed. I believe simply calling `setMaxThreadCount(0)` before calling `waitForDone()` should do the trick?

m7913d:
@axxel I added the requested changes, although I think a dead-lock could not happen as the destruction is called in the same thread as the `onVideoFrameChanged` method. Nevertheless, setting `maxThreadCount` to 0 doesn't harm. Another alternative may be the clear the sink before calling `waitForDone`.

axxel:
Thanks a lot for your time. If you have more suggestions on how to improve this qt-thingy or its visibility inside the qt-zxing community, I am all ears.

m7913d:
Concerning the Qt visibility, I would:
 - create a new Qt wrapper folder based on `ZXingQtReader.h` (mark it as experimental if you are not yet happy with the interface) and add relevant "Getting Started" information.
 - mention Qt / QML support in your README
 - create a PR to the original [ZXing](https://github.com/zxing/zxing) library to include Qt / QML in the description. It may be useful to also include that QZXing is based on the deprecated C++ port and that SCodes is an alternative Qt/QML wrapper that uses zxing-cpp

However, note that I'm not an expert in managing visibility of open source libraries.

PR #941, opened 2025-04-22
Added missing public header file Range.h

arnowillig:
The header "Range.h" was missing after a "make install".
The header is required by another header "ByteArray.h".
This fixes the issue.


axxel:
Thanks for pointing that out. I wanted to add another line to the patch to prevent a duplication and thought it is just quicker to do that myself than asking you to change your PR.

Issue #936, opened 2025-04-17
Change the behaviour of `bytesECI()`?


axxel:
This is about whether or not the behavior of `bytesECI()` is optimal with respect to the handling of the symbology identifier. My claim and mental model was that this property is exactly what a standard (hardware) barcode reader would produce. Looks like I was wrong.

@oehhar wrote [in another discussion](https://github.com/zxing-cpp/zxing-cpp/discussions/862#discussioncomment-12865746):

-----------------
Lets look to two examples:

The first code is a system login code with non ISO-8859-1 characters, e.g. containing an ECI:

![code_eci](https://github.com/user-attachments/assets/b4328fcc-ed81-4885-a758-3544b926724b)
The code contains an ECI 2 for ISO-Latin-2 and the data "1HJ" then a byte with decimal value 234 then "drek".
The byte with decimal 234 in ISO8859-2 is unicode codepoint x0119 "ę"
The decode result is as follows. Remark that byte 234 is shown by its Unicode representation  "ê".
- text 1HJędrek
- format DataMatrix
- bytes 1HJêdrek
- bytesECI ]d4\0000041HJêdrek
- content Text
- symbologyIdentifier \]d1
- hasECI 1

Here, bytesECI is, what a normal scanner would output. The item "text" with the decoded ECI is great (remark the change of "ê" to "ę" due to the ISO-8859-2 decoding).

The output of a standard scanner is exactly what "bytesECI" contains.

Now try a HIBC code without any ECI which only contains printable ASCII:

![code_no_eci](https://github.com/user-attachments/assets/c808b048-c469-486f-8b93-7b6d656a5691)

The decode result is:

- text +E4947611201/$LI
- format DataMatrix
- bytes +E4947611201/$LI
- bytesECI ]d4+E4947611201/$LI
- content Text
- symbologyIdentifier \]d1
- hasECI 0

The output of a standard scanner is what the concatenation of "symbologyIdentifier" and "bytes" contains.

To get the standard output in both cases, I do:

if (hasECI) return bytesECI else return symbologyIdentifier+bytes

-----------------

Only now did I realize the issue: `]d4` vs. `]d1` in the case where there is actually no eci designator contained in the message. I would like to change the behavior, such that it fits my mental model, namely exactly what Harald wrote in his last line above. I wonder whether a standard conforming reader is actually required to _not_ advertise the (potential) presence of ECI designators if there aren't any?

[...]

PR #942, opened 2025-04-22
Content: modify `bytesECI()` to behave like a standard reader w.r.t. ECIs

axxel:
The new code only advertises ECI data in the symbology identifier, if there were actually ECI designators present in the symbols bit stream.

For a detailed discussion and motivation, see
https://github.com/zxing-cpp/zxing-cpp/issues/936

### DIFF
--- a/core/src/ByteArray.h
+++ b/core/src/ByteArray.h
@@ -28,6 +28,7 @@ public:
 	explicit ByteArray(const std::string& str) : std::vector<uint8_t>(str.begin(), str.end()) {}
 
 	void append(ByteView other) { insert(end(), other.begin(), other.end()); }
+	void append(std::string_view other) { insert(end(), other.begin(), other.end()); }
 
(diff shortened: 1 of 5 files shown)


oehhar:
Thanks, Axel, I appreciate.
It will take me some time to test the revised implementation.

Please allow me to provide two test codes, in case you want to test it:

Code_1_backslash_2

   *   Text: 1\2
   *   Current bytesECI: ]d41\\2
   *   New bytesECI: ]d11\2

Code_ECI4_1_backslash_234

   *   Content: ECI4 1 \ 0d234
   *   text: 1\ę
   *   bytes: 1\ê 
   *   bytesECI: {]d4\0000041\\ê

Here, the "\" is correctly doubled. "text" translates the ECI correctly.

There is value in the old "bytesECI". It is just not following the standard.
The advantage is, that a client application always uses the ECI protocol and does not have to support the standard protocol.

Thanks for all,
Harald

PR #945, opened 2025-05-04
Improvements to Qt/QML interface

m7913d:
ZXingQt6CamReader.qml: Starting from Qt 6.6, camera permission should be requested manually. Only start the camera after the permission is granted by the user, otherwise starting the camera will fail and starting will not be retried after the permission is granted.

ZXingQtReader.h: always use BarcodeFormat/ContentType/TextMode enum from main ZXing namespace to allow calling
`ZXing::ReaderOptions().setFormats(ZXing::BarcodeFormat::MatrixCodes)` even if compiled with QML support (especially useful in case QML / QWidgets are mixed). Moved redefinition of those enum for QML to separate namespace.

BarcodeReader: expose flags as QFlags to qml

registerQmlAndMetaTypes: use qmlRegisterType to register the metatype to be able to store the position / barcode in a QML property, i.e. `property barcode myBarcode: ...`

registerQmlAndMetaTypes: expose ZXing as ZXingQt to QML, as the Qt interface of ZXing is also exposed using the ZXingQt namespace in C++.

Moved helper functions to Detail namespace

ZXingQtInitializer: automatically register qml types, similar to how Qt does it internally.

### DIFF
--- a/core/src/Flags.h
+++ b/core/src/Flags.h
@@ -67,6 +67,7 @@ public:
 	int count() const noexcept { return BitHacks::CountBitsSet(i); }
 
 	constexpr inline bool operator==(Flags other) const noexcept { return i == other.i; }
+	constexpr inline bool operator!=(Flags other) const noexcept { return i != other.i; }
 
 	inline Flags& operator&=(Flags mask) noexcept { return i &= mask.i, *this; }
 	inline Flags& operator&=(Enum mask) noexcept { return i &= Int(mask), *this; }
--- a/example/ZXingQt6CamReader.qml
+++ b/example/ZXingQt6CamReader.qml
@@ -9,7 +9,8 @@ import QtQuick.Controls
 import QtQuick.Layouts
 import QtQuick.Shapes
 import QtMultimedia
-import ZXing
+import ZXingQt
+import QtCore // required for CameraPermission (Qt 6.6 and above)
 
 Window {
 	visible: true
@@ -29,12 +30,12 @@ Window {
 		id: barcodeReader
 		videoSink: videoOutput.videoSink
 
-		formats: (linearSwitch.checked ? (ZXing.LinearCodes) : ZXing.None) | (matrixSwitch.checked ? (ZXing.MatrixCodes) : ZXing.None)
+		formats: (linearSwitch.checked ? (ZXingQt.LinearCodes) : ZXingQt.None) | (matrixSwitch.checked ? (ZXingQt.MatrixCodes) : ZXingQt.None)
 		tryRotate: tryRotateSwitch.checked
 		tryHarder: tryHarderSwitch.checked
 		tryInvert: tryInvertSwitch.checked
 		tryDownscale: tryDownscaleSwitch.checked
-		textMode: ZXing.TextMode.HRI
+		textMode: ZXingQt.TextMode.HRI
 
 		// callback with parameter 'barcode', called for every successfully processed frame
 		onFoundBarcode: (barcode)=> {
@@ -58,12 +59,23 @@ Window {
 		id: devices
 	}
 
+	// Starting from Qt 6.6, camera permission should be requested manually. For older Qt versions, CameraPermission object can be simply removed as Qt automatically requests the permission.
+	CameraPermission {
+		id: cameraPermission
+		Component.onCompleted: {
+			if (status !== Qt.PermissionStatus.Granted)
+				request();
+		}
+	}
+	// End of Qt 6.6+ section
+
 	Camera {
 		id: camera
 		cameraDevice: devices.videoInputs[camerasComboBox.currentIndex] ? devices.videoInputs[camerasComboBox.currentIndex] : devices.defaultVideoInput
 		focusMode: Camera.FocusModeAutoNear
 		onErrorOccurred: console.log("camera error:" + errorString)
-		active: true
+		active: cameraPermission.status === Qt.PermissionStatus.Granted // for Qt 6.6 and above
+		// active: true // pre Qt 6.6
 	}
 
 	CaptureSession {
--- a/example/ZXingQtReader.h
+++ b/example/ZXingQtReader.h
@@ -27,11 +27,12 @@
 
 namespace ZXingQt {
 
-Q_NAMESPACE
 
 //TODO: find a better way to export these enums to QML than to duplicate their definition
 // #ifdef Q_MOC_RUN produces meta information in the moc output but it does end up working in qml
 #ifdef QT_QML_LIB
+namespace QML {
+Q_NAMESPACE
 enum class BarcodeFormat
 {
 	None            = 0,         ///< Used as a return value if no valid barcode has been detected
@@ -64,19 +65,23 @@ enum class ContentType { Text, Binary, Mixed, GS1, ISO15434, UnknownECI };
 
 enum class TextMode { Plain, ECI, HRI, Hex, Escaped };
 
-#else
+Q_DECLARE_FLAGS(BarcodeFormats, BarcodeFormat)
+Q_DECLARE_OPERATORS_FOR_FLAGS(BarcodeFormats)
+Q_FLAG_NS(BarcodeFormats)
+Q_ENUM_NS(BarcodeFormat)
+
+Q_ENUM_NS(ContentType)
+Q_ENUM_NS(TextMode)
+} // namespace QML
+#endif
+
 using ZXing::BarcodeFormat;
 using ZXing::ContentType;
 using ZXing::TextMode;
-#endif
 
 using ZXing::ReaderOptions;
 using ZXing::Binarizer;
 using ZXing::BarcodeFormats;
-
-Q_ENUM_NS(BarcodeFormat)
-Q_ENUM_NS(ContentType)
-Q_ENUM_NS(TextMode)
 
 template<typename T, typename = decltype(ZXing::ToString(T()))>
 QDebug operator<<(QDebug dbg, const T& v)
@@ -129,8 +134,8 @@ public:
 
 	using ZXing::Barcode::isValid;
 
-	BarcodeFormat format() const { return static_cast<BarcodeFormat>(ZXing::Barcode::format()); }
-	ContentType contentType() const { return static_cast<ContentType>(ZXing::Barcode::contentType()); }
+	BarcodeFormat format() const { return ZXing::Barcode::format(); }
+	ContentType contentType() const { return ZXing::Barcode::contentType(); }
 	QString formatName() const { return QString::fromStdString(ZXing::ToString(ZXing::Barcode::format())); }
 	QString contentTypeName() const { return QString::fromStdString(ZXing::ToString(ZXing::Barcode::contentType())); }
 	const QString& text() const { return _text; }
@@ -138,6 +143,7 @@ public:
 	const Position& position() const { return _position; }
 };
 
+namespace Detail {
 inline QList<Barcode> ZXBarcodesToQBarcodes(ZXing::Barcodes&& zxres)
 {
 	QList<Barcode> res;
@@ -149,6 +155,7 @@ inline QList<Barcode> ZXBarcodesToQBarcodes(ZXing::Barcodes&& zxres)
 #endif
 	return res;
 }
+} // namespace Detail
 
 inline QList<Barcode> ReadBarcodes(const QImage& img, const ReaderOptions& opts = {})
 {
@@ -172,7 +179,7 @@ inline QList<Barcode> ReadBarcodes(const QImage& img, const ReaderOptions& opts 
 	};
 
 	auto exec = [&](const QImage& img) {
-		return ZXBarcodesToQBarcodes(ZXing::ReadBarcodes(
+		return Detail::ZXBarcodesToQBarcodes(ZXing::ReadBarcodes(
 			{img.bits(), img.width(), img.height(), ImgFmtFromQImg(img), static_cast<int>(img.bytesPerLine())}, opts));
 	};
 
@@ -282,7 +289,7 @@ inline QList<Barcode> ReadBarcodes(const QVideoFrame& frame, const ReaderOptions
 		}
 		QScopeGuard unmap([&] { img.unmap(); });
 
-		return ZXBarcodesToQBarcodes(ZXing::ReadBarcodes(
+		return Detail::ZXBarcodesToQBarcodes(ZXing::ReadBarcodes(
 			{img.bits(FIRST_PLANE) + pixOffset, img.width(), img.height(), fmt, img.bytesPerLine(FIRST_PLANE), pixStride}, opts));
 	}
 	else {
@@ -330,6 +337,9 @@ class BarcodeReader : public QObject, private ReaderOptions
 {
 	Q_OBJECT
 
+	Q_PROPERTY(QML::BarcodeFormats formats READ formatsQML WRITE setFormatsQML NOTIFY formatsChanged)
+	Q_PROPERTY(QML::TextMode textMode READ textModeQML WRITE setTextModeQML NOTIFY textModeChanged)
+
 public:
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
 	BarcodeReader(QObject* parent = nullptr) : QAbstractVideoFilter(parent) {}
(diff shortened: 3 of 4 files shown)


m7913d:
@axxel This PR contains some ideas to improve the Qt / QML interface. Feel free to cherry pick from the proposed changes if desired.

I'm not completely happy with the need for the enum duplication in the QML namespace. Code duplication could be avoided if we add the Q_NAMESPACE declaration to the original enum declarations (guarded by a QT_QML_LIB define). Disadvantage is that Qt-specific code will exist in the main library code. What do you think?

core/src/Flags.h
```diff
@@ -67,6 +67,7 @@ class Flags
 	int count() const noexcept { return BitHacks::CountBitsSet(i); }
 
 	constexpr inline bool operator==(Flags other) const noexcept { return i == other.i; }
+	constexpr inline bool operator!=(Flags other) const noexcept { return i != other.i; }
```
axxel, line 70: What made this addition mandatory? The `Q_DECLARE_FLAGS(BarcodeFormats, BarcodeFormat)` below is completely independent of my `Flags` class, right?

example/ZXingQt6CamReader.qml
```diff
@@ -9,7 +9,8 @@ import QtQuick.Controls
 import QtQuick.Layouts
 import QtQuick.Shapes
 import QtMultimedia
-import ZXing
+import ZXingQt
```
axxel, line 12: What was the motivation behind this? From a QML users point of view, `ZXing` seemed to be the canonical name for that.

example/ZXingQtReader.h
```diff
@@ -408,6 +406,20 @@ public slots:
 	void failedRead();
 	void foundBarcode(ZXingQt::Barcode barcode);
 
+	void formatsChanged();
```
axxel, line 409: Is this implicitly a `Q_SIGNAL` when mentioned in a `Q_PROPERTY` call?

example/ZXingQtReader.h
```diff
@@ -330,6 +337,9 @@ class BarcodeReader : public QObject, private ReaderOptions
 {
 	Q_OBJECT
 
+	Q_PROPERTY(QML::BarcodeFormats formats READ formatsQML WRITE setFormatsQML NOTIFY formatsChanged)
```
axxel, line 340: Maybe I'm overlooking something but does this mean the `formats` property of the (non-QML) class `ZXingQt::BarcodeReader` is of type `QML::BarcodeFormats`?

example/ZXingQtReader.h
```diff
@@ -497,14 +511,24 @@ inline void registerQmlAndMetaTypes()
 
 	// supposedly the Q_DECLARE_METATYPE should be used with the overload without a custom name
 	// but then the qml side complains about "unregistered type"
-	qRegisterMetaType<ZXingQt::Position>("Position");
-	qRegisterMetaType<ZXingQt::Barcode>("Barcode");
+	qmlRegisterType<ZXingQt::Position>("ZXingQt", 1, 0, "position");
+	qmlRegisterType<ZXingQt::Barcode>("ZXingQt", 1, 0, "barcode");
```
axxel, line 515: Why switch to `qmlRegisterType` here and why use a lower case type name (e.g. `position`)?

example/ZXingQtReader.h
```diff
@@ -346,35 +356,23 @@ class BarcodeReader : public QObject, private ReaderOptions
...
-	// results in the runtime error "can't assign int to formats"
-	Q_PROPERTY(int formats READ formats WRITE setFormats NOTIFY formatsChanged)
-	int formats() const noexcept
-	{
-		auto fmts = ReaderOptions::formats();
-		return *reinterpret_cast<int*>(&fmts);
-	}
-	Q_SLOT void setFormats(int newVal)
+	BarcodeFormats formats() const noexcept { return ReaderOptions::formats(); }
+	void setFormats(BarcodeFormats newVal)
```
axxel, line 360: So we have a "non-slot" `setFormats` and a slot `setFormatsQML`?

example/ZXingQtReader.h
```diff
@@ -27,11 +27,12 @@
 
 namespace ZXingQt {
 
-Q_NAMESPACE
 
 //TODO: find a better way to export these enums to QML than to duplicate their definition
 // #ifdef Q_MOC_RUN produces meta information in the moc output but it does end up working in qml
 #ifdef QT_QML_LIB
+namespace QML {
```
axxel, line 34: Why a sub-namespace `QML` here`? See also below.

axxel (commented):
Thanks a bunch for your effort in helping make this Qt stuff move forward! I have a couple questions...

core/src/Flags.h
m7913d: This operator is used at line 362 of ZXingQtReader.h.

`Q_DECLARE_FLAGS(BarcodeFormats, BarcodeFormat)` is indeed independent of `ZXing::Flags`, but `BarcodeReader` uses `ZXing::BarcodeFormats` and only converts to `ZXingQt::QML::BarcodeFormats` (= `QFlags`) when exposing it to QML.

example/ZXingQt6CamReader.qml
m7913d: You expose the Qt wrapper inside the ZXingQt namespace. I thought it would be more consistent if the QML wrapper is exposed with the same name. So, if someone uses your library from Qt, he should always use the name `ZXingQt` and never `ZXing` directly.

Another (consistent) option may be to drop the ZXingQt namespace and just use ZXing both in QML and as namespace in C++.

As said earlier, feel free to cherry pick from my proposed changes.

example/ZXingQtReader.h
m7913d: Advantage of moving the redefinition of the enums in order to expose it to QML, to a separate namespace is that the enums in ZXingQt can be a typedef to their ZXing versions. This allows your `ZXingQtReader.cpp` example to compile even if linked with the QML (Qt::Quick) library, i.e. `ZXingQt::ReaderOptions().setFormats(ZXingQt::BarcodeFormat::MatrixCodes)` is valid in all cases (which was previously not the case if linked with Qt::Quick).

Nobody should uses this QML namespace directly, as from C++, one should use `ZXingQt::BarcodeFormat` and in QML it is exposed as `ZXingQt.BarcodeFormat`. So it could be considered as an implementation detail.

example/ZXingQtReader.h
m7913d: No, `ZXingQt::BarcodeReader` does not store the `formats` property itself. The configured formats are only stored by the base class `ZXingQt::ReaderOptions` (=`ZXing::ReaderOptions`). So, `formats` is stored as `ZXing::BarcodeFormats`.

The `Q_PROPERTY` macro does not add a member to the class itself, it only tells to QML (i.e. Qt's meta-system) that a property `formats` is available for this class, which can be accessed using `formatsQML` and changed using the `setFormatsQML`, which returns/accepts a `QML::BarcodeFormats` value.

Technically, you can access the properties exposed by the `Q_PROPERTY` macro from C++ using Qt's meta system. If one would do so, he would indeed retrieve a `QML::BarcodeFormats` value, but I guess nobody will do this as just using `BarcodeReader::formats` and `BarcodeReader::setFormats` members is much easier from C++.

example/ZXingQtReader.h
m7913d: None of both member functions are marked as a slot. Starting from Qt 5, declaring member functions as slots becomes less important, as Qt signals and slots system does now allow connecting a signal to a slot using function pointers to any member (even those not declared as a slot).

Most of the time, I do not declare configuration (setter) functions as a slot. I reserve the slots mostly for the actions, such as "start" / "stop", "capture" or "hide"/"show". To be able to use a slot, one also needs a signal with the same arguments. For the setter functions, such a signal is most of the time not present. If desired, you can mark `setFormats` as a slot, it doesn't harm much either.

Concerning `setFormatsQML`, I wouldn't mark it as a slot as it is only used to expose it to QML, nobody should use it from C++ directly.

example/ZXingQtReader.h
m7913d: This is a signal because it is in a `signals` section (starts at line 405), not because it is mentioned in `Q_PROPERTY`.

You can mark a function as a signal by either
 - placing it in a signals sections (starts with `signals:` or `Q_SIGNALS:` (not the S at the end)
 - writing `Q_SIGNAL` in front of the member declaration.

Qt itself mostly uses the first option.

example/ZXingQtReader.h
m7913d: I'm not sure why you used `qRegisterMetaType` in the first place, as based on the comment it seems that your intention was to expose it to QML.

Exposing types to QML should be done using `qmlRegisterType`. This allow the type to be used for a QML property, i.e.

    // QML file
    ...
    property barcode myBarcode: ...
    ...

Value types (i.e. classes using the `Q_GADGET` macro instead of the `Q_OBJECT` macro), should be exposed to QML using a lower case name according to the [docs](https://doc.qt.io/qt-6/qtqml-cppintegration-definetypes.html#value-types-with-enumerations):

> In contrast to object types, value types require lower case names.

example/ZXingQtReader.h
axxel: I see, thanks for the explanation. I'm wondering whether there would be any benefit to be had on the QWidgets side if we simply used those `Q_...` augmented redefinitions in `ZXingQt` instead of the typedefs?

example/ZXingQtReader.h
axxel: > Technically, you can access the properties exposed by the `Q_PROPERTY` macro from C++ using Qt's meta system.

That is what I was talking about. My point is, that the meta-object-system from Qt which is used for signal dispatching and also cross-thread-boundary message passing and I forgot for what else more, has an application outside (and predates) QML. Advertising the `formats` property as being of type `QML::BarcodeFormats` even without that type even being defined, looks like a source of trouble to me. This would obviously also just go away, if we did away with this extra namespace (see above).

example/ZXingQtReader.h
axxel: > Qt signals and slots system does now allow connecting a signal to a slot using function pointers to any member (even those not declared as a slot)

But without moc-support you can't dynamically attach a signal to a slot with the `SIGNAL()` and `SLOT()` macros, right? While that might not be the most favorable approach nowerdays, I'd argue that it should still be supported for consistencies, sake.

example/ZXingQtReader.h
axxel: > I'm not sure why you used `qRegisterMetaType` in the first place

I don't remember either. It was most likely derived from some sample/tutorial code. If `QML_VALUE_TYPE` is the way to go for QT6, then I'd prefer that. Given the imminent deprecation of Qt5, I'd prefer to completely drop QT5 support before 'releasing' this as an 'official' wrapper anyway.

Regarding the lower-case naming, I'll need to read some docs to get the reasoning and implication, it seems.

core/src/Flags.h
axxel: > This operator is used at line 362 of ZXingQtReader.h.

Ah, and that used to be an integer comparison...

example/ZXingQt6CamReader.qml
m7913d: - Combining QWidgets and QML in a single project is rather rare, but may be done: see [`QQuickWidget`](https://doc.qt.io/qt-6/qquickwidget.html).
 - The same developer may integrate ZXing in multiple applications, which may be a combination of QML and QWidgets based application depending on the needs of the customers of the developer.
 - In a single project, one may use ZXing from C++ and QML together.  Before this PR, that was even required as one had to call `ZXingQt::registerQmlAndMetaTypes()` and import `ZXing` in QML.

Nevertheless, if you prefer to keep using `ZXing` in QML, I can revert

example/ZXingQtReader.h
m7913d: `Q_PROPERTY` is not needed/used for signal dispatching or cross-thread-boundary message passing. As far as I know, `Q_PROPERTY` indeed existed before QML but was mainly for QtDesigner integration. One could also query the value using `QObject::property`, but I'm not sure why one would do so (maybe for integration with a scripting environment?).

To be honest, I only use `Q_PROPERTY` to expose properties to QML.

example/ZXingQtReader.h
m7913d: In my opinion, not much. Almost the same as using `Q_PROPERTY` outside a QML environment. If I'm not mistaken, one advantage may be that `qDebug() << ZXingQt::BarcodeFormat::MatrixCodes` prints the name instead of the integer value.

example/ZXingQt6CamReader.qml
axxel: Yes, I'd prefer sticking with the `ZXing` name in QML.

example/ZXingQtReader.h
axxel: Not all functions in general but those that might be conceivable be hooked up with a slot. That is admittedly unlikely for this particular function (`setFormats`) but all setters should behave the same in this regards, I believe.

example/ZXingQtReader.h
axxel: That would be one advantage. Another would be that offering a `QFlag` based API is obviously more "native" to Qt developers and offering the most natural API in the different wrappers is definitively a design goal. And if we need to duplicate the enums for QML anyway, then why not use them consistently throughout the Qt wrapper, remove the need for the `QML` namespace as well as the 'duplication' of the `format` property.

m7913d: Does that imply that you also want to create a Qt wrapper for `ZXing::ReaderOptions` instead of using a simple typedef as `ZXing::ReaderOptions` expects `ZXing::BarcodeFormat` instead of `ZXingQt::BarcodeFormat`?

example/ZXingQt6CamReader.qml
m7913d: done

example/ZXingQtReader.h
m7913d: I tried using `QML_VALUE_TYPE`, but that also requires changes to the CMake project (see https://doc.qt.io/qt-6/qqmlintegration-h-qtqml-proxy.html#QML_ELEMENT), which seems undesirable to me if you want to expose it as a header only wrapper. So, I would propose to keep using `qmlRegisterType` for now (this function is not deprecated by Qt as far as I know).

example/ZXingQtReader.h
m7913d: done

example/ZXingQtReader.h
axxel: I agree. Thanks for looking into it.

example/ZXingQtReader.h
axxel: That is an excellent point, which I've overlooked. While this obviously worsens the cost-benefit ratio, I'd argue that it would still be worth it. First I thought: simply subclass `ReaderOptions` inside `ZXingQt` and add an appropriate set of `QFlag<>` overloads. But then I realized I also need a `TextMode` overload set. Further, it would make sens to have a signal/slot-ified object for convenient parameter configuration. Now I'm thinking: maybe better go directly the way I outlined in #724 (especially [here](https://github.com/zxing-cpp/zxing-cpp/discussions/724#discussioncomment-8567239)). To make the Qt wrapper not unnecessarily complex, I'd probably only implement the `BarcodeReader` class and not the `ReadBarcodes + ReaderOptions` alternative. This is not obviously not completely thought out but I'm running out of time for now.

If you have general comments about the fluid API idea (which is still in 'flow'...), please add them below the above linked comment for better discoverability.

m7913d: > I'd probably only implement the `BarcodeReader` class and not the `ReadBarcodes + ReaderOptions` alternative.

So, you would like to implement all use cases inside `BarcodeReader`, i.e. `inline QList<Barcode> ReadBarcodes(const QImage& img, const ReaderOptions& opts = {})` should become a member of `BarcodeReader`?

Concerning the fluent approach, I'm not sure how well it works in case of a `QObject` base class as it doesn't provide a copy/move constructor.

LeonnardoVerol:
Thats a really nice PR. I hope it gets merged

PR #947, opened 2025-05-07
Specified type for std::max and std::min to resolve windows.h issues.

puz-ya:
Because of windows.h header, the integrated solution fails to compile as they (MS) have min() & max() macros there with the same name. Specifying type(s) for std::min and std::max in angle brackets solves this issue.

Checked in VS 2022, C++20, platform toolset v143.

### DIFF
--- a/core/src/ImageView.h
+++ b/core/src/ImageView.h
@@ -112,8 +112,8 @@ public:
 	{
 		left   = std::clamp(left, 0, _width - 1);
 		top    = std::clamp(top, 0, _height - 1);
-		width  = width <= 0 ? (_width - left) : std::min(_width - left, width);
-		height = height <= 0 ? (_height - top) : std::min(_height - top, height);
+		width  = width <= 0 ? (_width - left) : std::min<int>(_width - left, width);
+		height = height <= 0 ? (_height - top) : std::min<int>(_height - top, height);
 		return {data(left, top), width, height, _format, _rowStride, _pixStride};
 	}
 
--- a/core/src/Point.h
+++ b/core/src/Point.h
@@ -111,7 +111,7 @@ auto length(PointT<T> p) -> decltype(std::sqrt(dot(p, p)))
 template <typename T>
 T maxAbsComponent(PointT<T> p)
 {
-	return std::max(std::abs(p.x), std::abs(p.y));
+	return std::max<T>(std::abs(p.x), std::abs(p.y));
 }
 
 template <typename T>
--- a/core/src/Range.h
+++ b/core/src/Range.h
@@ -103,7 +103,7 @@ public:
 	{
 		if (pos > _size)
 			return {};
-		return {_data + pos, std::min(len, _size - pos)};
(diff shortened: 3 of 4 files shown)


axxel:
Thanks for your input. This issue has come up before. In my opinion, the correct approach to fix this is the official [`NOMINMAX`](https://stackoverflow.com/questions/11544073/how-do-i-deal-with-the-max-macro-in-windows-h-colliding-with-max-in-std/11544154#11544154) solution.

From my point of view, `zxing-cpp` is a c++20 code base and if the library user sets up an environment that violates the c++20 specification, it is their responsibility to fix that, not "mine". That said, there might be room for improvement inside `zxing-cpp` that is non-invasive. Could you do the the favor and see if the following patch would fix your issue:
```patch
diff --git a/core/CMakeLists.txt b/core/CMakeLists.txt
index d1531057..974fc939 100644
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -53,6 +57,7 @@ endif()
 if (MSVC)
     set (ZXING_PUBLIC_FLAGS ${ZXING_PUBLIC_FLAGS}
         $<$<CXX_COMPILER_ID:MSVC>:/Zc:__cplusplus>
+        -DNOMINMAX
     )
 endif()
```

Side note: if I would want to work around the windows.h problem, I'd rather go with the `(std::min)(a, b)` solution, as this prevents duplicating the type.

puz-ya:
Thanks for reply! 
Alright, I've got your point to following standards and in that case "clean" std::min / std::max should be kept as it is.

> it is their responsibility to fix that, not "mine". 

Agreed. Thus I made this PR :) We have a legacy project and it's other way around for us: easier to modify this part. 
My thought was: if std::min<type> fix this issue and doesn't brake anything else (still following standards, right?), then why not make life a little bit easier?

> Could you do the the favor and see if the following patch would fix your issue

This will not work for our case, as NOMINMAX should be put somewhere we don't have access to (legacy).

> I'd rather go with the (std::min)(a, b), as this prevents duplicating the type.

Noted, with your explanation it make more sense!
I guess then this PR is useless.

axxel:
> This will not work for our case, as NOMINMAX should be put somewhere we don't have access to (legacy).

That part I don't quite understand. The patch I proposed would add the `NOMINMAX` define everywhere where you include something from `zxing-cpp`. As long as you use cmake of course. So the required access for this is the same you already need when adding a `zxing-cpp` dependency in the first place. The only issue that I could foresee is that some client code that includes `zxing-cpp` is also using the global `min/max` macros from `Windows.h`. 

> My thought was: if std::min fix this issue and doesn't brake anything else (still following standards, right?), then why not make life a little bit easier?

I understand your position as well. My main issue with this is mostly that I don't want to be responsible for making sure another `std::min()` does not show up at a later point. So the idea with the public `NOMINMAX` could potentially be a catch-all solution that also works for all future changes. And it would obviously keep the code cleaner.

PR #957, opened 2025-05-20
uprev cargo dependencies

samuelcolvin:


axxel:
So you are basically increasing the minimum required version of the image crate without changing anything that would depend on any new features of that crate. What is your motivation / gain?

samuelcolvin:
The motivation/gain is that I'm using other dependencies that rely on images "0.25.*", so I'm generating an image using that version. Without this change I can't use those images with this crate.

I've relaxed the constraint to avoid an upper bound, which is generally considered good practice, although seems somewhat less common in Rust than it is in Python.

axxel:
Thanks. I'll uptick the version of the package itself the next time I push a release. What is "the auto-formater" you are referring to here?

samuelcolvin:
> The white space changes were made by the auto-formatter, I'd recommend you made them anyway, but I've reverted them to get this merged.

many IDEs, including zed, format files (e.g. using prettier) automatically on save, which is why the changed happened.

axxel:
> many IDEs, including zed, format files (e.g. using prettier) automatically on save, which is why the changed happened.

Sure but that then becomes rather user specific. My first interpretation of your statement was that this layout is "the standard". Anyway, thanks for the contribution.

samuelcolvin:
well, prettier is:
1. widely used
2. following the toml standard - https://github.com/toml-lang/toml/issues/608

This doesn't matter much, but it's ambiguous what "right" is.

samuelcolvin:
reading QR codes, see https://github.com/samuelcolvin/qrcam/blob/main/src/decode.rs

Currently a WIP, but the idea is to have an app for reading QR codes using the webcam on macos.

PR #962, opened 2025-06-11
Update README.md with OpenHarmony build project

kuniqi:
We support build project of zxing-cpp in OpenHarmony, so we'd like to contribute to community and share the project.


axxel:
Given that your build files are based on version 2.0 which is completely outdated, I would suggest you update that first to the current release, which has considerable improvements compared to the 2.5 year old version you are using.

kuniqi:
yes，we are preparing to update it

Issue #961, opened 2025-06-10
QR Code Binary Encoding 1 Byte shorter


HoratiuMarginean:
Model 2 QR Codes in binary mode, error correction level L, should be able to hold 17 bytes for version 1, 32 for version 2, 53 for version 3...
Currently, it seems that they can hold 1 less byte than they should (16 for v1, 31 for v2...).

Numeric and Alphanumeric modes (I have not tested Kanji yet) hold the correct amount as specified in the ISO table.

I am using a MultiFormatWriter with QRCode format, margin of 4, ecc level L (or 2) and ISO8859_1 character set.

janpatera:
We don't use the writers, but my educated guess is that zxing-cpp encodes the terminator under all circumstances, while it could encode one extra codeword instead.

HoratiuMarginean:
Alright. I managed to look into this and found something weird.

From QREncoder.cpp:
```
EncodeResult Encode(const std::wstring& content, ErrorCorrectionLevel ecLevel, CharacterSet charset, int versionNumber, bool useGs1Format, int maskPattern)
{
	bool charsetWasUnknown = charset == CharacterSet::Unknown;
	if (charsetWasUnknown) {
		charset = DEFAULT_BYTE_MODE_ENCODING;
	}

	// Pick an encoding mode appropriate for the content. Note that this will not attempt to use
	// multiple modes / segments even if that were more efficient. Twould be nice.
	CodecMode mode = ChooseMode(content, charset);

	// This will store the header information, like mode and
	// length, as well as "header" segments like an ECI segment.
	BitArray headerBits;

	// Append ECI segment if applicable
	if (mode == CodecMode::BYTE && !charsetWasUnknown) {
		AppendECI(charset, headerBits);
	}

	// ... rest of the code
}
```

When setting the encoding for ZXing::MultiFormatWriter (or ZXing::QRCode::Writer) to ZXing::CharacterSet::ISO8859_1 explicitly before using their encode method, QR codes created have an additional 12 bits for data added onto them:
- 4 for the ECI mode marker
- 8 for the ECI bits
```
static void AppendECI(CharacterSet eci, BitArray& bits)
{
	int eciValue = ToInt(ToECI(eci));
	if (eciValue >= 0 && eciValue <= 999999) {
		bits.appendBits(static_cast<int>(CodecMode::ECI), 4);
		if (eciValue <= 127) {
			bits.appendBits(eciValue, 8);
		}
[...]

PR #967, opened 2025-06-13
Minimal fix for issue #961

HoratiuMarginean:
From my understanding, ECI mode is useful for charactersets other than ISO8859_1.

Given this, I updated the method to work even if this characterset was set manually.

(I have never opened a pull request before, so I apologize for any possible mistakes)

### DIFF
--- a/core/src/qrcode/QREncoder.cpp
+++ b/core/src/qrcode/QREncoder.cpp
@@ -483,10 +483,11 @@ static const Version& RecommendVersion(ErrorCorrectionLevel ecLevel, CodecMode m
 EncodeResult Encode(const std::wstring& content, ErrorCorrectionLevel ecLevel, CharacterSet charset, int versionNumber,
 					bool useGs1Format, int maskPattern)
 {
-	bool charsetWasUnknown = charset == CharacterSet::Unknown;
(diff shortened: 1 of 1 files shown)


HoratiuMarginean:
> Hi Horatiu, bar code symbologies have a default encoding. Older ones (Data Matrix, QR Code) use ISO/IEC 8859-1 (it is a bit more complicated, as 8859-1 does not define control characters in the range 80-8F). Newer ones use UTF-8.
> 
> Character sets may be changed by ECI, but that is not the only usage of ECIs.
> 
> ECI mode simulates the output of a standard scanner:
> 
> * no ECI present: Symbology ID flags that, for example ]d1 for Data Matrix
> * any ECI present: ECI protocol applies:
>   
>   * the symbnology ID flags ECI mode (]d4 for Data Matrix)
>   * any present ECI is represented by "\dddddd", where dddddd is the 6 digit ECI number.
>   * any data "" is replaced by "\".
> 
> I hope this helps, typed to much, soory, Harald

Hello oehhar,

I apologize for not specifying, but I was referring strictly to QR Codes in this case.

Thank you for the clarification however.

PR #972, opened 2025-06-30
Remove legacy `OneToFourBitPattern` function from Code93 decoder

wooyechan:
This PR removes the `OneToFourBitPattern` function, which was part of the legacy Code93 decoding logic.

Code93 now consistently uses `NormalizedE2EPattern`, instead of `OneToFourBitPattern`.
Removing this helps simplify the code and reduce technical debt.

**Key changes:**
- Replaced all usage of `OneToFourBitPattern` with the standardized `NormalizedE2EPattern` approach, already used across other symbologies.
- Introduced new files to isolate and manage Code93 pattern data:
  - `ODCode93Patterns.cpp`: Contains the `CODE_PATTERNS` definition for Code93.
  - `ODCode93Patterns.h`: Corresponding header for pattern definitions.

**Modified files:**
- `ODCode93Reader.cpp`: Removed `OneToFourBitPattern`, updated decoding logic to use `NormalizedE2EPattern`.
- `ODMultiUPCEANReader.cpp`: Removed residual mentions of `OneToFourBitPattern` from an unused/dead code path in `DecodeDigit`. This was done for cleanup and consistency.
- `ODRowReader.h`: Removed `OneToFourBitPattern`, which is no more needed.
- `core/CMakeLists.txt`: Added the new `ODCode93Patterns` files to the build.

Please review the changes and let me know if further adjustments are needed.

Thank you.


core/src/oned/ODMultiUPCEANReader.cpp
```diff
@@ -62,46 +61,6 @@ static bool DecodeDigit(const PatternView& view, std::string& txt, int* lgPatter
 		AppendBit(*lgPattern, bestMatch >= 10);
 
 	return true;
-#else
```
axxel, line 65: Please leave this disabled code alone. It might be useful as documentation in the future.

axxel (changes requested):
Thank you very much. If you feel like improving this PR, please see my comments. I thought about changing the `constexpr` part inside the Code128 decoder as a reference first, but I don't have the time right now.

wooyechan:
Thank you for the feedback. I’ve applied all your suggestions except for making E2E_PATTERNS a constexpr. The reason is that ToInt is currently used in the computation, which makes it difficult to convert directly under the current structure.

While it's possible to work around this by rewriting the logic differently, or by manually declaring all patterns as done in CODE_PATTERNS, I wasn't sure which approach you'd prefer. Also, I’m currently tied up with other tasks, so I couldn’t find the time to implement either option at the moment.

Apologies for postponing that part, and please feel free to share any further thoughts on the current changes.

core/src/oned/ODCode93Patterns.cpp
```diff
@@ -0,0 +1,62 @@
...
+* Copyright 2016 Nu-book Inc.
+* Copyright 2016 ZXing authors
+*/
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ODCode93Patterns.h"
+
+namespace ZXing::OneD::Code93 {
+
+constexpr std::array<std::array<int, 6>, 48> CODE_PATTERNS = { {
```
axxel, line 11: Here I meant move the definition of the variable into the header and drop the cpp-file completely. I'm not sure that adding `constexpr` at this location changes anything at all.

core/src/oned/ODCode93Patterns.h
```diff
@@ -0,0 +1,15 @@
+/*
```
axxel, line 1: Please update the copyright comment (replace the two lines with your name) and add a missing final new-line at the end of the file.

wooyechan:
Sorry about that. I accidentally clicked it.. :'( I’ll push the updated changes very soon.

core/src/oned/ODCode93Patterns.h
wooyechan: Thank you for the additional feedback.
I have removed the corresponding .cpp file from the CMakeLists.txt, and added a newline character at the end of the header file.

PR #975, opened 2025-07-08
Dantti/strict qt

dantti:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
+# See https://pre-commit.com for more information
```
axxel, line 1: I'll have a look at that when I have the time but will not merge this PR with this included right now. Please remove that for the time being (and submit as a separate PR, if you wish).

example/CMakeLists.txt
```diff
@@ -2,6 +2,19 @@ zxing_add_package_stb()
 
 include (GNUInstallDirs)
 
+add_definitions(
```
axxel, line 5: Please use `target_compile_definitions` for the qt-targets only. Also what is the motivation to define all of those? You mentioned `-DQT_NO_SIGNALS_SLOTS_KEYWORDS` and `-DQT_NO_KEYWORDS` but why all the others?

example/ZXingQtCamReader.cpp
```diff
@@ -1,14 +1,14 @@
...
 
+#include "ZXingQtReader.h"
+
 #include <QGuiApplication>
 #include <QQmlApplicationEngine>
 
-#include "ZXingQtReader.h"
-
-int main(int argc, char *argv[])
+int main(int argc, char* argv[])
```
axxel, line 11: Please remove all white-space changes / clang-format fixes from this PR.

axxel (changes requested):
Thanks for your effort and welcome to the project :).

example/CMakeLists.txt
dantti: The others are to ensure QString is efficiently created

PR #979, opened 2025-07-20
const up some symbols

caolanm:
and move them out of the .data section


axxel:
May I ask how you found those? Some automated tool, I assume?

PR #987, opened 2025-08-06
feature: Native ARM64 support for Python wheels

stumpylog:
This PR adds building aarch64 wheels using the somewhat new native runners Github provides.

Closes #536 for good.

### DIFF
--- a/.github/workflows/publish-python.yml
+++ b/.github/workflows/publish-python.yml
@@ -22,46 +22,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-13, windows-latest] # at least macos-13 is required to enable c++20 support
-
-    steps:
-    - uses: actions/checkout@v4
-      with:
-        submodules: true
-
-    - name: Set up Python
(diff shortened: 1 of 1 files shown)


axxel:
Thank you for looking into this. It turns out that there is no ubuntu-latest-arm64 runner. So I was looking into this and found that the most recent `cibuildwheel` v 3.x allows to simplify the configuration considerably.

Please have a look at [this publish-python run](https://github.com/axxel/zxing-cpp/actions/runs/16799704770) and please manually install one of the [arm wheels](https://github.com/axxel/zxing-cpp/actions/runs/16799704770/artifacts/3708564387) and let me know if that works.

stumpylog:
Oh nice. I was thinking to make the v3 upgrade later and maybe transition to the action as well.  But your changes look even simpler.  Installs fine on my Pi

stumpylog:
If you get a chance to merge that here, there's some follow on work I'd like to investigate.  Then it would be awesome to see a release including the ARM wheels.

axxel:
Thanks for confirming that it worked. I pushed the change to this repo now.

PR #997, opened 2025-09-30
added 'BarcodeReader::from_image' alongside 'from'

5GameMaker:
The reasoning is that `from` is also implemented on trait `From` and thus you get a "From<&Image> is not implemented" error.

(Unfortunately couldn't test for Android as zxing-cpp did not compile)


5GameMaker:
Nevermind, that's some weird `image` stuff apparently

PR #999, opened 2025-10-02
Barcode Scanner and Visualizer

HamzaYslmn:
This Python script provides a simple way to scan barcodes from images (local files or URLs) using the zxing-cpp library. It decodes barcodes, extracts their information, and visually highlights detected barcodes on the image. The script can be run directly to test barcode detection and display results with bounding boxes.

### DIFF
--- a/wrappers/python/demo_read_show.py
+++ b/wrappers/python/demo_read_show.py
@@ -1,68 +1,70 @@
-from typing import List, Dict, Any, Union  # built-in
-from PIL import Image, ImageDraw  # pip install pillow
-from io import BytesIO  # built-in
-import zxingcpp  # pip install zxing-cpp
-import httpx  # pip install httpx
+# pip install opencv-python numpy zxing-cpp httpx
+from typing import Union
+import sys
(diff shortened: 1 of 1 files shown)


axxel:
Thanks for sharing your work. I have a couple of issues with it that don't make this a good addition as is:
* it depends on both PIL and opencv that seems redundant to me
* it contains non-english language
* it does not allow the user to pass in a url or filename
* it constructs something like a copy of the `Barcode` object
* it calls `get_bytes` twice, which does not make sense to me

There might be more. So I'll not merge this as is. If at all, it seems to make more sens to me to extend the existing [demo_reader.py](https://github.com/zxing-cpp/zxing-cpp/blob/master/wrappers/python/demo_reader.py) code.

HamzaYslmn:
> * it calls `get_bytes` twice, which does not make sense to me

It's just a arbitrary PR, don't worry too much about it. I saw something missing while working on the project with the kids and wanted to add it.
If you want, you can pull it, it's up to you :D

PR #1001, opened 2025-10-08
Use 16 KB page size alignment on 64-bit Androids

markusfisch:
Because starting November 1st, 2025, all new apps and updates to existing apps submitted to Google Play and targeting Android 15+ devices must support 16 KB page sizes on 64-bit devices.


axxel:
Thanks. I removed the unrelated removal of a comment.

PR #1017, opened 2025-12-09
Add support to disable some algorithms

misery:
Each codec adds source files, which increases compile time and the final library size. If the project only needs QR codes, there’s no reason to include ONED or DataMatrix support.


misery:
Oh, sorry... I just checked the release branch. Nice!

PR #1039, opened 2026-01-30
kn: fix CI and add documents for building script

isning:
This pull request had done following things:

1. Switched cinterop to use built header files rather than source headers to avoid failure of finding building artifact `Version.h`.
2. Introduced zig for providing sysroot for building linux targets and build shared libraries for them. This allows us to reuse the generated header files and run tests against the resulting dynamic libraries. But it's notable that we still do NOT ship shared libraries for linux targets.
3. Moved content of `libZXing.def` to `build.gradle.kts`.
4. Fixed invalid `strictEnums` configuration in `libZXing.def` before.

> Although Zig is often described as “another language compatible with C,” in this context we only use it for the cross-compilation sysroot it provides, in order to build Linux targets. In practice, it is essentially a wrapper around Clang, combined with a set of tools that handle tasks such as downloading and managing sysroots. It also makes it easy to specify details like the glibc version used by the sysroot.
> 
> Otherwise, we would have to maintain the sysroot ourselves and handle compilation for architectures such as ARM64. Given the fragmented nature of the Linux ecosystem, this would likely be quite cumbersome.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,12 +23,12 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
       with:
         submodules: true
 
     - name: Setup Python 3
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: '3.x'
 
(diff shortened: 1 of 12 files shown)


axxel:
My bad, I sent you the wrong link. Here is the right one: https://github.com/zxing-cpp/zxing-cpp/actions/runs/21869699416

PR #1040, opened 2026-01-30
chore(gitignore): update build artifacts

CHC383:
- `build` is created by VS Code clangd
- `zxing-cpp.release` is the output folder name from README, and could be created inside sub-folders

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,12 @@
+# Build artifacts
 *.o
 *.so
 *.lib
 *.d
 *.a
 compile_commands.json
+/build
+/**/zxing-cpp.release
 
 # QtCreator
 CMakeLists.txt.user


axxel:
I'm ok with adding the /build entry (also ran into this default behavior of VS Code. But the `/**zxing-cpp.release` entry is not convincing. please remove that.

axxel:
> Do you think the Build Instructions should be updated to simply use `build` instead of `zxing-cpp.release`? Otherwise people following them will still get bunch of unstaged files.

Your assumption is wrong. the build instructions are explicitly designed to prevent altering the source tree at all. I can recognize that there are cmake tutorials advertising a /build subdirectory inside the project folder, which then basically requires to add that to .gitignore and it seems to be the vscode default behavior as well.

I'll change the instructions to be more 'standard' and modify the gitignore file accordingly. thanks for suggesting improvements.

PR #1047, opened 2026-02-08
Fix MSVC v143 (VS2022) compilation bug

Mq-b:
Change `using Data = struct BarcodeData;` to `using Data = BarcodeData;` and add forward declaration. Add VS2022 CI job for regression testing.

I honestly don't quite understand why it compiles in VS2026 yet. I don't have the time to create a minimal reproducible example for testing and research at the moment.


axxel:
Thanks for looking into this. I compared the Visual Studio version and the MSVC version between the windows-latest and your new windows-2022 runner and they are exactly the same. So I'm pretty sure the code as currently in HEAD would work in your new CI job even without your workaround.

That said, I was able to come up with a minimal godbold example showing the problem with MSVC 19.42: https://godbolt.org/z/Mzs9W458W. Starting with 19.43 it works: https://godbolt.org/z/W5n4rTK93.

I thought that my solution would actually hide the name `BarcodeData` from the `ZXing` namespace for the client side. But it looks like it does not. This makes that name part of the ABI, which I don't like. I might rather switch the roles and make `BarcodeData` an alias for `Barcode::Data` instead of the other way around.

Mq-b:
> Thanks for looking into this. I compared the Visual Studio version and the MSVC version between the windows-latest and your new windows-2022 runner and they are exactly the same. So I'm pretty sure the code as currently in HEAD would work in your new CI job even without your workaround.
> 
> That said, I was able to come up with a minimal godbold example showing the problem with MSVC 19.42: https://godbolt.org/z/Mzs9W458W. Starting with 19.43 it works: https://godbolt.org/z/W5n4rTK93.
> 
> I thought that my solution would actually hide the name `BarcodeData` from the `ZXing` namespace for the client side. But it looks like it does not. This makes that name part of the ABI, which I don't like. I might rather switch the roles and make `BarcodeData` an alias for `Barcode::Data` instead of the other way around.

Sorry about that—looks like I broke the CI. I think the CI script still needs some work. That said, I've clearly tested that after modifying Barcode.h, it builds just fine on my end.

axxel:
> Thanks for the quick fix and the shout-out! Though I have to admit, I was secretly hoping to 'sneak' into the contributor list with a PR myself. 😉 You were just too fast for me! Glad the issue is resolved though.

Oh, I'm really sorry for that and completely understand the thought. (Once I was able to get a single line patch in the Linux kernel, which is something I took pride in for years - even long after that line was gone...). the 2 line fix was just way faster than communicating and making you remove the CI part, which I currently believe would not have added anything useful. Next time! :)

Mq-b:
> > Thanks for the quick fix and the shout-out! Though I have to admit, I was secretly hoping to 'sneak' into the contributor list with a PR myself. 😉 You were just too fast for me! Glad the issue is resolved though.
> 
> Oh, I'm really sorry for that and completely understand the thought. (Once I was able to get a single line patch in the Linux kernel, which is something I took pride in for years - even long after that line was gone...). the 2 line fix was just way faster than communicating and making you remove the CI part, which I currently believe would not have added anything useful. Next time! :)

No problem at all! Thanks for the encouragement and for sharing that story. I’m looking forward to more opportunities to contribute in the future! :)

PR #1055, opened 2026-02-11
kn: fix maven central artifacts publishing

isning:
@axxel I think this should fix maven central publishing.

### DIFF
--- a/.github/workflows/publish-kn.yml
+++ b/.github/workflows/publish-kn.yml
@@ -85,10 +85,13 @@ jobs:
       - name: Publish Library
         if: ${{ github.event.inputs.publish == 'y' }}
         env:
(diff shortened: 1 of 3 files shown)


isning:
> I was excited to see that the publish-kn now went through [successfully](https://github.com/zxing-cpp/zxing-cpp/actions/runs/21912241954/job/63269242975) but after logging in to https://central.sonatype.com/publishing, where any publication is supposed to show up as a draft that manually needs to be made public. This has worked for the android publication but nothing showed up from the kn publication.

Wierd, I tested on my side today, it do works:
https://github.com/isning/zxing-cpp/actions/runs/22086125551
image

Maybe you can run it again or check whether the SNAPSHOT publishment enabled for the namespace?
image

PR #1061, opened 2026-02-24
Swift: new wrapper based on C-API instead of ObjectiveC++

axxel:


farfromrefug:
@axxel does that mean you are dropping the objc version? if yes then it would be nive for yhe swift version to support @objc. many projects still use objc and won't stop anytime soon (much better c++ interoperability).
we also use it in an environment where we need objc.

jknlsn:
Thanks for the reply and the feedback! Some of this may well be me lacking project/C++ context, so I very much appreciate the clarifications.

> I deliberately did not do that since I wanted to prevent 2 things: a) copying (meaning maintaining duplicates) or loosing default values and b) having to set all values, even when they are not changed. Loosing those features might be worth it if making them Sendable is outweighing their benefits. What is the use case that your approach supports over mine?

Those two concerns are still covered in my implementation, `cDefaults` preserves the C library's defaults, and the Swift initialiser only requires callers to set the values they want to change. The difference is just where the mutable C state lives, temporarily at the call boundary rather than permanently inside the Swift object, because I didn't see a need to keep that native mutable state around between calls. So I don't think those features need to be lost, and the main additional benefit is compiler-checked `Sendable` conformance. In my own projects I'm trying to use Swift 6 strict concurrency where possible, so I do see value in the compiler-enforced concurrency safety that comes from that.

> You realized that keeping the Barcode handle is required for the zint machinery. I have not understood what your approach brings to the table then.

The main thing the struct approach brings is a Swift value-type surface with `Sendable` behaviour at the wrapper boundary, rather than exposing a mutable reference type with native state directly to callers. The handle is still retained internally for rendering and metadata access, so nothing is lost on that front. On the equality question I'm happy to leave content-based equality out of the discussion for now. You're right that the use case is less clear-cut, and callers who need to de-duplicate can always compare fields directly rather than having that baked into the wrapper.

> Also, there is one expensive property (`symbol`)... Moving that over to a Swift value is therefore a pessimization.

Agreed, if `symbol` were exposed, it should definitely be lazy/on-demand rather than eagerly copied. That could be addressed by trimming the eager-copy set to only the commonly-accessed fields.

> The retained object is the same (e.g. `data`) so what is wrong with the old code?

The concern is less about whether it happens to work in practice, and more about what lifetime the API actually guarantees. `withUnsafeBytes` documents the buffer as only valid for the duration of the closure, so storing that pointer in the native `ImageView` means the pointer escapes the scope Swift guarantees. By contrast, `NSData.bytes` gives a pointer whose lifetime is tied to the retained `NSData` object. Since the wrapper already retains that object for the lifetime of the `ImageView`, using `.bytes` makes the pointer lifetime explicit and guaranteed by the API contract.

> Looks like I have missed that part, I'm running low on time right now, so I'm skipping that.

No problem. On another look I don't think this one is worth changing. The current behavior is fine as-is.

> Not quite clear where you see the source of those drifting apart.

The main thing I was thinking about is that the enums are manually mirrored between C++ and Swift with no codegen keeping them in sync. So if a case gets added on the C++ side and the Swift wrapper doesn't get updated at the same time, it would still compile fine and the crash would only show up at runtime when a user happens to hit the new value. It just feels more natural to me for the bridging to be defensive there, since a thrown error is a lot more actionable than a bare force-unwrap crash. And the code cost is slim, it's basically `try` instead of `!` at the call sites.

On your numbered points about the reference implementation (naming consistency, file structure, C API changes, etc.), those are all entirely fair, and I think separate focused PRs would be a better way to discuss the ones worth pursuing.

Overall, I agree that the best option is to land this PR and then look at further changes from there. The current implementation already looks entirely usable from Swift/iOS to me, so none of my suggestions are blockers :)

PR #1071, opened 2026-03-12
fix test/fuzz build

kcwu:


kcwu:
I'm interested in making the software I'm using more stable (away from crash and/or security issue).

PR #1082, opened 2026-03-13
pdf417: fix heap-buffer-overflow in ByteCompaction

### LINKED ISSUE #1076: Heap-buffer-overflow in `ZXing::Pdf417::ByteCompaction` due to incorrect handling of interleaved ECIs
## Steps to Reproduce
Run the fuzzer `fuzzDecodeMatrix` with the following input.

```bash
printf "\x40\xe7\xe7\x7e\x37\x37\x00\x00\x37\x37\x00\x00\xff\x7e\x62\xfa\x37\x00\x00\x00" > crash.bin
./build-fuzz/fuzzDecodeMatrix crash.bin
```

## Crash Message (ASAN)
```
=================================================================
==1864918==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x504000005bf8 at pc 0x560f1bfa79b0 bp 0x7ffcde0c1810 sp 0x7ffcde0c1808
READ of size 4 at 0x504000005bf8 thread T0
    #0 0x560f1bfa79af in ZXing::Pdf417::ByteCompaction(int, std::vector<int, std::allocator<int>> const&, int, ZXing::Content&) /usr/local/google/home/kcwu/fuzz/targets/zxing-cpp/zxing-cpp/core/src/pdf417/PDFDecoder.cpp:401:29
    #1 0x560f1bfa79af in ZXing::Pdf417::Decode(std::vector<int, std::allocator<int>> const&) /usr/local/google/home/kcwu/fuzz/targets/zxing-cpp/zxing-cpp/core/src/pdf417/PDFDecoder.cpp:655:51
```

## Root Cause Analysis
In PDF417, Byte Compaction mode can contain interleaved ECI (Extended Channel Interpretation) escape sequences.
The decoding logic in `ZXing::Pdf417::ByteCompaction` used a look-ahead function `CountByteBatches` to determine the number of 5-codeword batches and trailing codewords. `CountByteBatches` correctly identifies and skips ECIs, counting only the data codewords.

However, the actual decoding loops in `ByteCompaction` (both for 5-codeword batches and trailing codewords) did not skip ECIs. If an ECI occurred inside a 5-codeword batch, it was incorrectly treated as a data codeword. This led to a misalignment of the `codeIndex`.

When `ProcessByteECIs` was subsequently called between batches or in the trailing loop, it could consume ECIs that `CountByteBatches` had already accounted for as being *outside* the data count. This resulted in `codeIndex` advancing beyond the end of the `codewords` vector, causing a heap-buffer-overflow when the loop tried to read the next expected data codeword.

## Fix Proposal
The fix involves updating the decoding loops in `ByteCompaction` to call `ProcessByteECIs` before reading each data codeword. This ensures that ECIs are correctly processed and skipped, keeping the `codeIndex` synchronized with the data codeword count returned by `CountByteBatches`. Additionally, `ProcessByteECIs` was updated to throw a `FormatError` if an unknown codeword (>= 900 and not a latch or ECI) is encountered, for consistency with `CountByteBatches`.

kcwu:
ECIs can be interleaved in Byte Compaction mode. The decoding loops must process and skip them to keep synchronized with the data codeword count.

Fixes a crash found by fuzzer.

this fixes #1076

### DIFF
--- a/core/src/pdf417/PDFDecoder.cpp
+++ b/core/src/pdf417/PDFDecoder.cpp
@@ -358,6 +358,8 @@ static int ProcessByteECIs(const std::vector<int>& codewords, int codeIndex, Con
 		int code = codewords[codeIndex++];
 		if (IsECI(code))
 			codeIndex = ProcessECI(codewords, codeIndex, codewords[0], code, result);
+		else
+			throw FormatError();
 	}
 
 	return codeIndex;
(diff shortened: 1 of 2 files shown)


axxel:
Found a better solution, respecting the spec.

Issue #1072, opened 2026-03-12
Slightly deformed DataMatrix symbols fail to decode


d-buchmann:
Digikey (a vendor of electronics components) puts data matrix codes on their products and most of them work well, e.g.:

<img width="150" height="150" alt="Image" src="https://github.com/user-attachments/assets/9e5a3078-939a-42e6-9a90-10213ad796ec" />  <img width="150" height="150" alt="Image" src="https://github.com/user-attachments/assets/83fd3535-3d7c-438b-8a65-f128f5452fcd" />

However, some codes just don't decode even though there isn't any obvious quality issue, so that posed the question if the content is actually malformed. I put one of the original codes (left) besides a copy that I created (right) using Zint Barcode Studio 2.16 from the binary data that I decoded using an [online scanner](https://online-barcode-reader.inliteresearch.com/).

<img width="300" height="150" alt="Image" src="https://github.com/user-attachments/assets/b322f540-09ad-40b2-a67e-3361c9051221" />

Looking at the raw binary content of both codes, it seems to suggest that the original uses reader initialization, but does not have the initial code word `0xea`. But the first codes (which decode fine) are also lacking it, so I don't understand what's going on here.

@axxel, maybe you can shed some light on this? Thank you very much in advance.

References:
https://github.com/Part-DB/Part-DB-server/issues/1281
https://github.com/Sec-ant/zxing-wasm/issues/330

Issue #1063, opened 2026-02-25
Datamatrix not detected


Edouard-Tby:
Hi thanks a lot for this great library.

I’m having trouble detecting the attached barcode. The image quality seems reasonable to me, but detection fails.

<img width="187" height="190" alt="Image" src="https://github.com/user-attachments/assets/b4a5b348-5469-4339-906c-eca5ea39dd39" />

Do you see anything in this image that could explain why detection fails?

Best,

Edouard

Issue #794, opened 2024-06-02
2D barcode not decoding properly 


Digvijay548:
I have multiple imgs which are not decoding we have used DataMatrix detection for That some barcode get decoding perfectly but others are not and this is frequently done With same barcodes, can I know the reason ( this non decoding barcode decoded by mobile app without any issue )

axxel:
To be able to say anything about the reason why a particular symbol was not detected, I need to have the image itself for inspection.

Digvijay548:
[Images.zip](https://github.com/user-attachments/files/15528658/Images.zip)
 We have these images that we aren't able to decode but when we use any mobile app, they are decoded

Digvijay548:
we are using python zxing-cpp for this and we are unable to decode We have also done object detection for better output but this exceptional images are not decoding

axxel:
I had a quick look at the image set and a closer look at one of the samples and the reason `zxing-cpp` does not currently detect those samples is mainly because they are printed on a curved surface, which is captured from some angle. This results in a distortion within the symbol that can't be dealt with the current approach of using only a perspective transformation.

That said, I noticed that at least two forks exist, where someone experimented with other ideas regarding datamatrix detection, see [here](https://github.com/chorbier/zxing-cpp/tree/sampleGridStuff) and [here](https://github.com/epoch8/zxing-cpp). Maybe you find something that works for you. Please consider providing feedback to @chorbier and @epoch8.

axxel:
> but when we use any mobile app, they are decoded

I tried 2 different ones and none worked. Can you give me an example for one that works for you?

Digvijay548:
I have used barcode Scanner by cognix

Digvijay548:
is there any solution ?

axxel:
Well, I gave you 2 pointers to potential solutions. Have you tried them? I am currently not working on this problem myself.

Digvijay548:
Which solution you are asking for ?

axxel:
> Which solution you are asking for ?

https://github.com/zxing-cpp/zxing-cpp/issues/794#issuecomment-2144901272

Digvijay548:
We have decoded bot images in mobile app of cognex barcode decoder app

Sandeep418:
> > Which solution you are asking for ?
> 
> [#794 (comment)](https://github.com/zxing-cpp/zxing-cpp/issues/794#issuecomment-2144901272)

[...]

PR #1088, opened 2026-03-21
DMDetector: timing-pattern-based correction for deformed symbols

Sec-ant:
Related: #794, #1063, #1072

Hi, @axxel. **The code is generated by an AI coding agent (the primary model is Claude Opus 4.6)**. I was testing my agent tooling and picked these long-standing DM issues as a case study. I went through quite a few iterations with the model to match the code style conventions and verify correctness, but I'm not deeply familiar with the detector internals myself, so treat this more as a reference/proof-of-concept than a polished contribution. **Feel free to close if it's not helpful.**

There are 2 commits in this PR:

1. ~~`ReadBarcodes: let valid results supersede errors at the same position`~~

   ~~When `returnErrors=true`, an error result can shadow a later valid result at the same position because `Contains(res, r)` blocks the valid one from entering `res`. This replaces `Contains` with `Find` + an `else if` branch that overwrites an existing error when a valid result arrives.~~

   This is superseded by 3dd7f60325c7cd7629a5cf381cf2801339bb87e3.

2. `DMDetector: timing-pattern-based correction for deformed symbols`

   The idea is to use the gap structure in timing pattern edges to build piecewise-linear remap functions, then re-sample the grid with non-uniform module spacing when the standard perspective-based sampling fails. Corrected candidates are yielded after standard ones so well-formed symbols are not affected.

Tested on 5 real-world images from the issues above:
| Image | Size | Before | After |
|-------|------|--------|-------|
| issue1063 | 44×44 | ChecksumError | ✓ |
| issue1072-left | 48×48 | ChecksumError | ✓ |
| issue794-12-1 | 26×26 | ChecksumError | ✓ |
| issue794-13-4 | 26×26 | ChecksumError | ✓ |
| issue794-3-2 | 26×26 | NoBarcode | ✓ | 

You're free to edit this branch or close this one if it involves so much editing that it's easier to write the code in your own way.


core/src/datamatrix/DMDetector.cpp
```diff
@@ -718,7 +718,225 @@ class EdgeTracer : public BitMatrixCursorF
...
+	for (int anchor : anchors)
+		results.push_back(buildRemap(anchor));
+	return results;
+}
+
+// Sample the grid using timing-pattern-corrected non-uniform module spacing.
+// Returns an empty result if the correction is not applicable or insignificant.
+static DetectorResult SampleGridCorrected(const BitMatrix& image, int width, int height,
+	const PerspectiveTransform& mod2Pix, const TimingRemap& topRK, const TimingRemap& rightRK)
+{
```
d-buchmann, line 883: I have added this snippet to produce an additional grid view when the timing corrected sampling is applied.
````c++
#ifdef PRINT_DEBUG
	LogMatrix log;
	static int i = 0;
	LogMatrixWriter lmw(log, image, 5, "grid_c" + std::to_string(i++) + ".pnm");
#endif
````

axxel:
Interesting stuff. Thanks for putting in the effort and sharing your results (on such generous terms, too). I started looking into this and would want to make some changes before merging it but it obviously solves real problems, so I intend to add something along those lines.

I had something more general in mind that would also work for other symbologies but your approach is obviously better than the status quo.

d-buchmann:
Thanks from me as well =)

From my testing symbols, most pass now which previously didn't. But still one negative sample remains:
digikey_4
It works if I do a _manual_ perspective correction before trying to decode.

Sec-ant:
I noticed the 3dd7f60325c7cd7629a5cf381cf2801339bb87e3 commit supersede the ReadBarcodes workaround, so I dropped that one and rebased this branch. An additional logic is added (with the help of the model) to address the sample provided by @d-buchmann in https://github.com/zxing-cpp/zxing-cpp/pull/1088#issuecomment-4125125311.

axxel:
@Sec-ant thanks for further improving the PR. I was working on in yesterday as well and noticed the domain mixup as well (and fixing it turned out to fix the additional sample, too). My approach is different though: actually fixing the domain mixup instead of introducing yet another sampling strategy.

There are more simpliplifications and cleanup I want to apply. I might end up simply pushing a new PR.

axxel:
Superseded by https://github.com/axxel/zxing-cpp/commit/b304f6656b865be11b00741d430b7d0027bc0ab4

PR #1094, opened 2026-04-09
iOS: add formats getter/setter to ZXIReaderOptions

hrasnam:
## Summary

- The `formats` property on `ZXIReaderOptions` is declared in the header but has no getter/setter implementation, so reading or writing barcode format filters on the iOS wrapper is a no-op
- This PR adds the missing getter and setter implementations that bridge between `NSArray<NSNumber *>` and the C++ `BarcodeFormats`
- The getter filters out C++ sub-variant formats (e.g. `QRCodeModel1`, `AztecRune`) that have no `ZXIFormat` mapping, preventing spurious `NONE` entries in the returned array
- The `formats` property qualifier is changed from `strong` to `copy` to follow Objective-C conventions for collection properties

## Notes

- An empty formats array means "scan all formats", consistent with the C++ layer's behavior
- Setting a symbology-level format (e.g. `QR_CODE`) and reading it back may return expanded variants (e.g. `[QR_CODE, MICRO_QR_CODE, RMQR_CODE]`) — this is expected, as it reflects what the C++ layer will actually scan for

cc @axxel

### DIFF
--- a/wrappers/ios/Sources/Wrapper/Reader/ZXIReaderOptions.mm
+++ b/wrappers/ios/Sources/Wrapper/Reader/ZXIReaderOptions.mm
@@ -4,6 +4,7 @@
 
 #import "ZXIReaderOptions.h"
 #import "ReaderOptions.h"
+#import "ZXIFormatHelper.h"
 
 @interface ZXIReaderOptions()
 @property(nonatomic) ZXing::ReaderOptions cppOpts;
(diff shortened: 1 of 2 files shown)


wrappers/ios/Sources/Wrapper/Reader/ZXIReaderOptions.mm
```diff
@@ -17,6 +18,30 @@ -(instancetype)init {
...
+
+-(void)setFormats:(NSArray<NSNumber *> *)formats {
+    std::vector<ZXing::BarcodeFormat> nativeFormats;
+    nativeFormats.reserve(formats.count);
+
+    for (NSNumber *formatValue in formats) {
+        nativeFormats.push_back(BarcodeFormatFromZXIFormat((ZXIFormat)formatValue.integerValue));
+    }
+
+    self.cppOpts = self.cppOpts.setFormats(ZXing::BarcodeFormats(std::move(nativeFormats)));
```
axxel: I believe `self.cppOpts.setFormats(std::move(nativeFormats));` should also work?

wrappers/ios/Sources/Wrapper/Reader/ZXIReaderOptions.mm
```diff
@@ -17,6 +18,30 @@ -(instancetype)init {
     return self;
 }
 
+-(NSArray<NSNumber *> *)formats {
+    NSMutableArray<NSNumber *> *formats = [NSMutableArray array];
+    for (auto format : ZXing::BarcodeFormats::list()) {
```
axxel: why don't you iterate over `self.cppOpts.formats()` directly?

wrappers/ios/Sources/Wrapper/Reader/ZXIReaderOptions.mm
hrasnam: done 👍

wrappers/ios/Sources/Wrapper/Reader/ZXIReaderOptions.mm
hrasnam: done

axxel:
Thanks!

hrasnam:
> Thanks!

You're welcome 🙂. Do you have any time-estimate for the next release?

PR #1097, opened 2026-04-11
Fix constructor parameter names in PointT

MonkeybreadSoftware:
We build usually with warnings for shadowed variables. So the parameters and the properties of the class should be different to avoid any confusion.

### DIFF
--- a/core/src/Point.h
+++ b/core/src/Point.h
@@ -23,7 +23,7 @@ struct PointT
 	T x = 0, y = 0;
 
 	constexpr PointT() = default;
-	constexpr PointT(T x, T y) : x(x), y(y) {}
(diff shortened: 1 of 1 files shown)


axxel:
I assume you are talking about your own project where you'd like to get rid of the warning, right? Is that the only one in the public API of `zxing-cpp`?

There are certainly other places in the `zxing-cpp` code base where shadowing happens. I am personally not a fan of those "fixes", especially trivial ones like you posted, since there is no confusion in my eyes there at all. If anything, they make the code less clear.

That said, if you really care about, I can live with this very local and isolated "fix".

MonkeybreadSoftware:
Well, yes, I build my own apps using zxing. And I'd like to try to reduce the warnings. 
I hope to someday contribute a more meaningful thing. Thank you.

PR #1103, opened 2026-04-22
Fix #include bug

KangLin:
When not install qml, then the follow error:

/home/install/include/ZXing/ZXingQt.h:18:10: fatal error: QtQmlIntegration/qqmlintegration.h: No such file or directory
   18 | #include <QtQmlIntegration/qqmlintegration.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.


axxel:
Thanks!

Issue #1104, opened 2026-04-22
Build bug in windows


KangLin:
OS: windows

The files under `core/src/libzint` are symbolic links. However, on Windows, these links cannot be accessed. So there will be errors during compilation. Files cannot be found.

It is recommended to use the files under the `zint` directory directly in CMakeLists.txt, rather than linking them.

PR #1105, opened 2026-04-23
Fix build bug in windows #1104

KangLin:
https://github.com/zxing-cpp/zxing-cpp/discussions/1037


axxel:
As I already mentioned: this approach breaks the build for the release tarball (for all systems).

KangLin:
> As I already mentioned: this approach breaks the build for the release tarball (for all systems).

I haven't encountered it. How do you operate it?

axxel:
You could not possibly have encountered it (yet) because there was never a release tarball/zip created for this PR. But those tarballs do not contain the `zint` subdirectory, so the CMakeLists.txt changes will break the build.

KangLin:
Since it is not included, why does it still exist in the source code?

PR #1108, opened 2026-05-06
Add PEP 561 typing support via stub file

desaleo:
## Add PEP 561 typing support via stub file

This PR adds basic typing support for the Python bindings.

### Motivation

Type hints are widely used in modern Python (3.7+) for static analysis and improved IDE support. Providing type information can make `zxing-cpp` easier to integrate in typed codebases without affecting existing users.

### What this PR does

* Adds a `zxingcpp.pyi` stub file describing the public Python API
* Adds a `py.typed` marker to comply with PEP 561
* Includes docstrings in the stub to improve discoverability and editor help

This approach does not change the runtime behavior nor the current build system.

### Notes

* The stub file was written manually based on `zxingcpp.cpp`, as tools like `pybind11-stubgen` currently produce incomplete results for this project.
* I’m aware of the discussion in [issue#948](https://github.com/zxing-cpp/zxing-cpp/pull/948) about potentially moving from pybind11 to nanobind, which may eventually allow automatic stub generation. This PR is meant as an incremental improvement until such a migration happens.
* I took the liberty to add a type alias `BarcodeFormatsLike` as it's, to my knowledge, the best way to stub `py::implicitly_convertible`. This can be adjusted or removed if a different approach is preferred.

### Alternative if not desired upstream

If maintaining type information in this repository is not preferred, I can instead publish a separate stub package (`types-zxing-cpp`) if you're Ok with that.

---

Let me know if you’d like this scoped down further, or adjusted.

PR #948, opened 2025-05-08
python: switch binding library to `nanobind`

Doekin:
This PR migrates the Python bindings for `zxingcpp` from `pybind11` to `nanobind`.

**Key Benefits:**

1.  **Enhanced IDE Support (Autocompletion & Type Hinting):**
    The primary driver for this change is `nanobind`'s straightforward support for generating stub files (`.pyi`). These stubs enable autocompletion and type hinting in IDEs, significantly improving the development experience when using the `zxingcpp` library.

    439856664-5b6c6dfc-3e86-450a-b75d-fb6ca92cd863

2.  **Smaller Binary Sizes**

3.  **Faster Compilation Times**

**Important Structural Change:**

*   As a result of this migration and to better accommodate the stub files and `nanobind`'s typical project structure, the `zxingcpp` Python extension has been refactored from a single-file module (e.g., `zxingcpp.pyd` or `zxingcpp.so`) into a Python **package** (a directory named `zxingcpp` containing an `__init__.py` and the compiled extension module).

PR #1109, opened 2026-05-06
Python wrapper: migrate from pybind11 to nanobind + scikit-build-core

stumpylog:
## Summary

This replaces the Python wrapper's build system and binding layer entirely. The pybind11 + setuptools/`setup.py` stack is replaced with [nanobind](https://github.com/wjakob/nanobind) and [scikit-build-core](https://github.com/scikit-build/scikit-build-core), which are the current recommended tools.

- scikit-build-core + pyproject.toml replaces `setup.py` -- fully PEP 517/518 compliant, no more setuptools. `python -m build` and `pip install` both work correctly.
- nanobind replaces pybind11. Standalone CMake builds fetch it via `zxing_add_package`; scikit-build-core builds get it from the build environment declared in `build-system.requires`.
- `zint/backend` is now correctly included in the sdist via a `wrappers/python/zint` symlink (mirroring the existing `core` symlink pattern).

## Package structure

The module is now a proper Python package rather than a single flat `.so`:

```
zxingcpp/
  __init__.py          # re-exports everything from _zxingcpp_core
  _zxingcpp_core.so    # compiled extension (or .pyd on Windows)
  _zxingcpp_core.pyi   # checked-in type stub (PEP 484)
  py.typed             # PEP 561 marker (enables type checker support)
  stubs.pats           # nanobind stub pattern file
```

## Performance improvements

- **numpy fast path**: `read_barcode` and `read_barcodes` have a dedicated `nb::ndarray<nb::numpy, uint8_t>` overload that skips the Python buffer protocol and passes the array directly to the C++ core with zero extra copies.
- **GIL release**: The GIL is released around all `ReadBarcodes` and `WriteBarcode`/`WriteSVG` calls. The C++ core is CPU-bound and does not call back into Python, so this is safe and allows Python threads to run concurrently with decode/encode operations.
- **`nb::is_final()`** on `Barcode`, `Image`, and `ImageView`: enables CPython 3.11+ `LOAD_ATTR` specialization (~1.22x faster attribute/method access on those types).
- **`nb::rv_policy::move`** on all `Barcode`-returning functions: avoids an extra copy when returning decode results to Python.

## What did not change

The public Python API is fully backwards-compatible. All existing `import zxingcpp` code continues to work unchanged.  No tests were changed, only some additions


stumpylog:
I do apologize for the size, but swapping out one wrapping method for another isn't really something possible piecemeal.

It starts from the basis of https://github.com/zxing-cpp/zxing-cpp/pull/948#issuecomment-3801594248, updating the packaging structure then the conversion and a few nanobind niceties on top.  It also makes the sdist a complete package.  Untar it and it is buildable as is, since it now includes `zint`, using the same symlink model as `core`

This replaces #948 and #1108.

axxel:
Thanks for sharing your work on this. I had a closer look at your PR over last couple days but decided to largely stick with the approach in #948. I had issues with the following aspects of your approach:
* the ndarray "fast path" is a duplication of code that adds not benefit (according to my understanding)
* I did not like the raw CPython API usage
* I don't want to include the auto-generated stub file in the git repo
* the scikit_build_core setup can be simpler
* the python package layout in the repo did not convince me after all

If you find any issues or missing optimization opportunities with my pushed version, please let me know.

PR #1117, opened 2026-05-21
DMDecoder: restrict AIM id (2nd position FNC1) to alpha/2 digits

gitlost:
Although the DataMatrix spec doesn't specify the format for the AIM
id, all other barcodes follow the Code 128 spec, so following suite
seems the better interpretation.


oehhar:
It is not "AIM ID", it is ISO/IEC 15424 Symbology ID, which was just revised.
Document on the way...

gitlost:
Thanks @oehhar.

What I'm referring to with "AIM id" should be called "AIM Application Indicator" (or more fully "AIM Application Specification Indicator") according to Annex B of ISO/IEC 15424:2025.

Annex B does confirm that alpha and/or 2 digits is the only allowed format for "AIM AppInd" (as I'll now call it).

PR #1122, opened 2026-06-09
README: add link to Julia wrapper

rakeshksr:
Hi @axxel, Thanks for this great library!

I am currently maintaining the Julia wrapper for it, and this pull request adds a link to the wrapper in the README so users can easily find it.
Thanks for considering the PR!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ It was originally ported from the Java [ZXing library](https://github.com/zxing/
     | [iOS](wrappers/ios/README.md) | [Kotlin/Native](wrappers/kn/README.md) | [.NET](wrappers/dotnet/README.md) |
     | [Python](wrappers/python/README.md) | [Qt](wrappers/qt/README.md) | [Rust](wrappers/rust/README.md) |
     | [WebAssembly](wrappers/wasm/README.md) | [WinRT](wrappers/winrt/README.md) | [Flutter](https://pub.dev/packages/flutter_zxing) (external project) |
+    | [Julia](https://github.com/rakeshksr/ZXingCPP.jl) (external project) | | |
 
 ## Supported Formats
 


axxel:
Thanks for sharing your work. In principle, I'm interested but I'd change the layout in the readme a little.

But before I do that: would you be interested in discussing a few changes (starting with the name of the wrapper/project (see https://github.com/zxing-cpp/zxing-cpp/discussions/720)) and continuing with some API design decisions you took?

rakeshksr:
Thanks for the interest! I followed the official Julia [Package naming guidelines](https://pkgdocs.julialang.org/v1/creating-packages/#Package-naming-rules) when choosing the name.

The package is built on top of a C wrapper, adapting the underlying C functions for reading and writing barcodes into standard Julia conventions. I'm open to suggestions on the API design.

axxel:
**Naming:** I would like to suggest calling the package `ZXingCpp.jl`, which seems to perfectly fit the (old) guidelines. Nowadays one apparently needs to add a hint that this is a wrapper library, e.g. `ZXingCppWrapper.jl` or `LibZXingCpp.jl`.

**API:** I don't know Julia but looking at your code, it seems it supports named optional parameters like Python and Swift. In that case, I'd suggest to do away with the `CreatorOptions` and `WriterOptions` (and `ReaderOptions` most likely, too) and instead add those parameters to the functions directly. See also https://github.com/zxing-cpp/zxing-cpp/discussions/726#discussioncomment-16157621. A priority for me is make each wrapper feel as native as possible to the respective language / environment. Mostly to make sure there is less incentive to port the whole project (again...) just to get a more native interface.

rakeshksr:
**Naming:** The package is already registered in the Julia registry, so renaming it would be difficult at this stage.

**API:** I have extended the functions to support optional parameters directly, so `CreatorOptions`, `WriterOptions`, and `ReaderOptions` are no longer necessary. You can find examples of the updated API in the README.

axxel:
> **Naming:** The package is already registered in the Julia registry, so renaming it would be difficult at this stage.

I see. Just for the sake of asking: would it technically be possible to release a new wrapper with a consistent name?

> **API:** I have extended the functions to support optional parameters directly, so `CreatorOptions`, `WriterOptions`, and `ReaderOptions` are no longer necessary. You can find examples of the updated API in the README.

That looks much nicer from a hight level perspective. One last thing that seems questionable to be is the elaborate / foreign looking naming of the enum constants like `ZXing_BarcodeFomat_QRCode`. Is that the most Julia like approach to it?

axxel:
I added it with a little layout tweak.

PR #1134, opened 2026-07-14
MicroPDF417: fix out-of-bounds matrix access in ScanCandidate corner extrapolation

### LINKED ISSUE #1133: MicroPDF417: out-of-bounds codeword-matrix access in ScanCandidate corner extrapolation
## Summary

`MicroPdf417::ScanCandidate()` can index the codeword `Matrix` out of bounds in
its corner-extrapolation step. Debug builds abort on the `Matrix` bounds
assertion (`Matrix.h:66`, `SIGABRT`); release builds (`NDEBUG`) do a silent
out-of-bounds read.

Notably this is reachable without opting into MicroPDF417: requesting `PDF417`
is enough, since `MicroPDF417 <= PDF417` makes `MultiFormatReader` instantiate
`MicroPdf417::Reader` as well.

## Version

Current `master`; introduced by `2317e21f` (*"MicroPDF417: estimate symbol
position from detected codeword positions"*).

## When it happens

Live-camera scanning via the ObjC wrapper (`ZXIBarcodeReader`) with `PDF417`
enabled. No actual MicroPDF417 symbol needs to be present... the detector forms a
degenerate candidate from ordinary frames and the corner extrapolation then
reads outside the codeword matrix.

```
Assertion failed: (x >= 0 && x < _width && y >= 0 && y < _height),
function operator(), file Matrix.h, line 66.   // x = -1, y = 0
```
```
4  ZXing::Matrix<...Codeword>::operator()(int x, int y)   Matrix.h:66
5  ZXing::Matrix<...Codeword>::operator()(PointI p)
6  ZXing::MicroPdf417::ScanCandidate(...)                 MicroPDFReader.cpp
7  ZXing::MicroPdf417::Reader::read(...)
8  ZXing::MultiFormatReader::read(...)
9  ZXing::ReadBarcodes(...)
```

## Root cause

`closestCorner()` derives its scan range from the **module** dimensions
(`si.width()/2`, `si.height()/2`) but indexes `cwMat`, which is `nCols x 53`
**codewords**:

```cpp
for (int x = 0; x <= si.width() / 2; ++x)
    for (int y = 0; y < si.height() / 2; ++y) {
        auto offset = PointI{x, y} * dir;
        if (cwMat(corner + offset).count > 1)   // corner + offset may be out of range
            return offset;
    }
```

Deterministic case: `nCols == 1` ⇒ `si.width() == 38` ⇒ `x` runs `0..19`. For
`trI` (`corner = {nCols-1, startRow} = {0,0}`, `dir = {-1,1}`), `x == 1` gives
`cwMat(-1, 0)`.

## Suggested fix

A per-cell bounds check keeps in-range behaviour identical:

```cpp
auto p = corner + offset;
if (p.x < 0 || p.y < 0 || p.x >= cwMat.width() || p.y >= cwMat.height())
    continue;
if (cwMat(p).count > 1)
    return offset;
```

I went with a guard rather than fixing the loop bounds directly because the
correct per-corner extent depends on `startRow` and the ±`dir` of each of the
four corners; happy to do it whichever way you prefer (or add a `Matrix::isIn()`
mirroring `BitMatrix`). PR attached.

Verified against the real `Matrix` header: unguarded aborts as above for
`nCols == 1`, guarded returns cleanly, core still builds.

## Possibly related

The linear read `(&cwMat(0, si.startRow))[i]` for `i in [0, si.nCWs())` a few
lines further down can over-read the flat buffer for the same degenerate
candidates (silent, no assert).

alexmanzer:
Fixes #1133.

`MicroPdf417::ScanCandidate()`'s `closestCorner()` scans over the module
dimensions (`si.width()/2`, `si.height()/2`) but indexes `cwMat`, which is only
`nCols x 53` codewords, so `corner + offset` can leave the matrix. For a
degenerate candidate (`nCols == 1`) the first out-of-range step is `cwMat(-1, 0)`
— assert/`SIGABRT` in debug, silent out-of-bounds read in release. It is
reachable via the plain `PDF417` format, since `MicroPDF417 <= PDF417`.

This adds a per-cell bounds check. I chose a guard over fixing the loop bounds
directly because the correct extent differs per corner (`startRow` offset and
the ±`dir`); glad to switch to bounded loops or a `Matrix::isIn()` helper
mirroring `BitMatrix` if you prefer.

Verified against the real `Matrix` header (unguarded aborts for `nCols == 1`,
guarded returns cleanly) and the core library still builds.

### DIFF
--- a/core/src/pdf417/MicroPDFReader.cpp
+++ b/core/src/pdf417/MicroPDFReader.cpp
@@ -736,7 +736,12 @@ static BarcodeData ScanCandidate(const BitMatrix& image, const Cluster& lraps)
 		for (int x = 0; x <= si.width() / 2; ++x)
 			for (int y = 0; y < si.height() / 2; ++y) {
 				auto offset = PointI{x, y} * dir;
-				if (cwMat(corner + offset).count > 1)
+				auto p = corner + offset;
+				// The scan range is derived from the module dimensions (si.width()/height()) while cwMat is
+				// only nCols x 53 codewords, so p can fall outside the matrix for small/degenerate candidates.
+				if (p.x < 0 || p.y < 0 || p.x >= cwMat.width() || p.y >= cwMat.height())
+					continue;
+				if (cwMat(p).count > 1)
 					return offset;
 			}
 		return {};


axxel:
Thanks for providing a pin-point hint at the point of failure. The fix, suggested by the AI, gave away that it did not understand what the real issue actually was... Thank god, my expertise is not completely useless (yet). ;).

alexmanzer:
@axxel 
Unfortunately, both of the fixes on master still have issues that trigger assertions (now on other places).
It just happened again with commit `5442cd5746aa7152c15d058b70bad68ed6518502`.

Here’s the analysis of the backtrace from the crashed app.
(I wouldn't open another merge request right now, since they usually don't get implemented anyway.)

--- 

Analysis:
MicroPDF417: second out-of-bounds in ScanCandidate when si.nCols == 0 (follow-up to #1133)

Follow-up to #1133 — bounding the closestCorner scan range fixed the loop access, but there's a second, independent out-of-bounds in the same function that it doesn't cover.

Symptom: Matrix bounds assertion (Matrix.h:66, x = -1, y = 0) → SIGABRT in debug / OOB read in release. Same trigger as #1133 — live scanning with PDF417 enabled (so MicroPdf417::Reader runs) on a spurious candidate; no real MicroPDF417 symbol needed.

Stack:
```
Matrix<…Codeword>::operator()(x=-1, y=0)   Matrix.h:66
Matrix<…Codeword>::operator()(PointI)
MicroPdf417::ScanCandidate                 MicroPDFReader.cpp:751
```

Cause: SymbolInfo::nCols defaults to 0 and DetermineNumCols() can return 0 for a degenerate candidate. The corner extrapolation's dst uses si.nCols - 1 directly as a column index:
```
auto dst = QuadrilateralF(cwMat(PointI{0, si.startRow} + tlI).leftPos(),
                          cwMat(PointI{si.nCols - 1, si.startRow} + trI).rightPos(), …);
```

With si.nCols == 0 this becomes cwMat(-1, …). closestCorner itself doesn't fault in this case — with a 0-sized symbol its loops are empty, so it returns {} and the OOB surfaces at the dst line.

Suggested fix: reject a degenerate symbol before the extrapolation, e.g. right after DetermineSymbolInfo:
```
if (si.nCols <= 0 || si.nRows <= 0)
    return {};
```

That covers both closestCorner and the src/dst construction — a 0-column/row candidate can't be decoded anyway. (Wherever you prefer; the point is that a degenerate SymbolInfo currently reaches the corner extrapolation.)

axxel:
Thanks for reporting (and testing) this. Fix is in place.

PR #1135, opened 2026-07-22
rmQR: sample short symbols via scored fusion of all function patterns

coliniuliano:
Improved rmQR detection for a personal project since I saw it said "TODO proper" in the source.
This was entirely done and refined with Claude Opus and Fable.
I used the following test images and got the following results:

image
image
image

image
image

AI description:

The single perspective transform estimated from the top-left finder pattern extrapolates the module grid across the full symbol width, so a sub-pixel error at the finder drifts to a full module at the far edge of wide symbols (e.g. R7x43) and sampling fails. Small (~5px/module) blurry symbols were additionally discarded by hard reject-gates before sampling even began.

Replace the sampling for short (7/9 module) symbols with one scored fusion of every available function pattern, mirroring how SampleQR uses what it finds and rebuilds the rest, and make every pre-sampling stage measurement-tolerant:

Detection (finder pattern quad and format information):
- use whichever finder boundary ring line is measurable (5x5/7x7 blend, either alone, or the 3x3 core) - each maps to a known square in module coordinates
- tolerate one blurred module in the top-edge gate and sample the 18 format information modules under a grid of sub-module offsets (wider for single-ring quads), keeping the best BCH-scored read
- keep the best read per candidate version instead of trusting a single best-hamming word: the sampled image content (the fused edges' joint timing score) arbitrates, so a wrong version that decodes to valid-looking format info loses to content, and a near-valid read one bit past the BCH limit may prove itself against the image

Sampling (scored fusion of the module grid):
- gather edge anchors from the finder, traces along both edge timing patterns (with adaptive direction re-estimation so steep rotation is followed), and the finder sub pattern
- reject anchors that drifted off the symbol's row (perpendicular deviation check against the finder's prediction)
- robustly line-fit each edge, dropping outliers such as the discrete jumps a timing mis-count produces
- jointly refine the two edges against the image content: search a window around the extrapolated far corner (scale and tilt, seeded from both fitted edges) and score each candidate by how well BOTH rows then explain their known timing content. Scoring the rows together pins the shared geometry with twice the evidence and rejects decoys such as a neighbouring symbol's timing row
- only adopt a scored line over a fitted one when it wins by a clear margin: the score plateaus at low resolution and plateau noise must not nudge a good fit off its measurement
- keep an independently fitted bottom edge only if it agrees with the refined top about the symbol height and its row content scores no worse than the configuration the joint search validated
- sample the grid in tiles between the fitted edges

Unreliable features are simply discarded by the fit and scoring - a single pass, no fallback strategies, no rescaling. The decoder checksum remains the final arbiter. QR and MicroQR sampling are unchanged.

### DIFF
--- a/core/src/BinaryBitmap.h
+++ b/core/src/BinaryBitmap.h
@@ -48,6 +48,7 @@ public:
 
 	int width() const { return _buffer.width(); }
 	int height() const { return _buffer.height(); }
+	const ImageView& imageView() const { return _buffer; }
 
 	/**
 	* Converts one row of luminance data to a vector of ints denoting the widths of the bars and spaces.
@@ -55,6 +56,17 @@ public:
 	virtual bool getPatternRow(int row, int rotation, PatternRow& res) const = 0;
 
 	const BitMatrix* getBitMatrix(bool transposed = false) const;
+
+	/**
+	* Access to the underlying luminance for detectors that re-threshold locally
+	* (per symbol). Returns -1 when the buffer is not single-channel luminance.
+	*/
+	int luma(int x, int y) const
+	{
+		return _buffer.format() == ImageFormat::Lum && x >= 0 && x < width() && y >= 0 && y < height()
+				   ? *_buffer.data(x, y)
+				   : -1;
+	}
 
 	void invert();
 	bool inverted() const { return _inverted; }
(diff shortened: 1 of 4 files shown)


axxel:
This surely looks like an impressive improvement. I only had a brief look so far. May I ask why you decided to close the PR?

coliniuliano:
Closed because I was still making changes, opened a new one.

PR #1136, opened 2026-07-23
rMQR: sample short symbols via scored fusion of all function patterns

coliniuliano:
Improved rmQR detection for a personal project since I saw it said "TODO proper" in the source.

This was entirely done and refined with Claude Opus and Fable.
Each image was tested at various angles and all gave very good results.

I used the following test images and got the following results:

image
image
image

image
image

AI description:

The single perspective transform estimated from the top-left finder pattern extrapolates the module grid across the full symbol width, so a sub-pixel error at the finder drifts to a full module at the far edge of wide symbols (e.g. R7x43) and sampling fails. Small (~5px/module) soft symbols were additionally discarded by hard reject-gates before sampling began, and the global binarization thresholds them by the luck of its block-grid alignment relative to the symbol.

Replace the rMQR detection and sampling with one scored measurement pipeline (QR and MicroQR are untouched):

Detection:
- two binarization front-ends feed one pipeline: finder patterns are scanned in both the local-average and a global-histogram binarization of the same luminance (each symbol proceeds with the binarization that detected it)
- the finder quad uses whichever boundary ring line is measurable (5x5/7x7 blend, either alone, or the 3x3 core)
- format information is sampled under a grid of sub-module offsets from the binary image AND from grayscale thresholded on the finder's own modules; the best read per candidate version is kept and the sampled image content arbitrates between versions

Grid geometry (scored fusion):
- anchors from the finder, traces along both edge timing patterns, and the finder sub pattern; robust line fits with outlier rejection
- refinement objective: continuous luminance separation over ALL modules of known color (finder, both timing rows including alignment columns, corner and sub pattern) - the anchors break the even-module shift invariance that pure alternation scoring suffers from
- search: seeded far-corner grid, a coarse 4D sweep over both endpoints around pure finder geometry, then shrinking-step coordinate descent from multiple starts

Sampling and verification:
- the grid is sampled from the globally binarized image AND from luminance with per-column thresholds calibrated on the symbol's known modules; the error correction arbitrates (a decodable matrix wins outright), with known-module agreement as the tie-breaker
- if the primary configuration fails to decode, near-tied geometry hypotheses from the search are trial-decoded (decode-verified rescue, paid only by symbols that failed anyway)

### DIFF
--- a/core/src/BinaryBitmap.h
+++ b/core/src/BinaryBitmap.h
@@ -48,6 +48,7 @@ public:
 
 	int width() const { return _buffer.width(); }
 	int height() const { return _buffer.height(); }
+	const ImageView& imageView() const { return _buffer; }
 
 	/**
 	* Converts one row of luminance data to a vector of ints denoting the widths of the bars and spaces.
@@ -55,6 +56,17 @@ public:
 	virtual bool getPatternRow(int row, int rotation, PatternRow& res) const = 0;
 
 	const BitMatrix* getBitMatrix(bool transposed = false) const;
+
+	/**
+	* Access to the underlying luminance for detectors that re-threshold locally
+	* (per symbol). Returns -1 when the buffer is not single-channel luminance.
+	*/
+	int luma(int x, int y) const
+	{
+		return _buffer.format() == ImageFormat::Lum && x >= 0 && x < width() && y >= 0 && y < height()
+				   ? *_buffer.data(x, y)
+				   : -1;
+	}
 
 	void invert();
 	bool inverted() const { return _inverted; }
--- a/core/src/qrcode/QRDetector.cpp
+++ b/core/src/qrcode/QRDetector.cpp
@@ -8,6 +8,9 @@
 
 #include "QRDetector.h"
 
+#include "BinaryBitmap.h"
+#include "DecoderResult.h"
+#include "QRDecoder.h"
 #include "BitArray.h"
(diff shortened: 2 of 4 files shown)


coliniuliano:
With grayscale it can now detect 80 codes from the (full res) wide image in ~700ms.

image

axxel:
Thanks for putting your time into improving zxing-cpp. However, this PR is not in a state where I would merge it. The one commit it contains is way too big and touches too many things at once to properly review it. Also, just from the PR description there are things that I generally don't approve of, like using two different binarization algorithms in parallel. Another general worry I have is runtime performance, for which I care a lot. That means if your image contains say 100 symbols, of which your current approach detects 80 by using 700ms and a less effective approach finds only 70 say but does so in 100ms then that one would be preferred by me.

To make some progress here, could you please do the following:
1) supply the full resolution image that you used to generate the above results.
2) remove the parallel application of local/global binarization
3) instruct your AI agent to split the changeset into several ones that are coherent and somewhat minimal in what they are doing. I would guess an appropriate number might be 5 - 10 changesets (again, I have not properly looked at the code at all)

EDIT:
4) The FindFinderPatternsLuma idea that duplicates existing infrastructure and tries to fix directional blur (very general problem) in a single specific symbology (rmQR) will not be merged either. 

Please don't close this PR again just to open another one with the same objective. Any content of this PR can be replaced/amended.

Issue #145, opened 2020-07-16
PDF417 anyone?


axxel:
After I'm almost done with #89 the remaining main 'offender' in terms of performance is the PDF417 decoder. I started looking at the code and found it to be pretty hard to read (to put it politically correct :)). I would also question the general approach the authors took: the code basically performs individual horizontal line-scanning operations. This can generally only work as long as the rotation of the symbol is so small that at least one complete codeword is visible when scanning a horizontal line. I tested the code by starting with a [perfect](https://github.com/nu-book/zxing-cpp/blob/master/test/samples/pdf417-3/01.png) symbol upscaled it by a factor of 2, to make it even easier, and rotated it. Even just 3° rotation are already too much. On the other hand the current code deals happily with corrupt samples like [this](https://github.com/nu-book/zxing-cpp/blob/master/test/samples/pdf417-3/11.png).

I would argue that it is more important to handle real-world rotations caused by hand-held cameras than to be able to deal with failing dot-matrix printers or thick black lines half across your code. Meaning: if a (faster) replacement would deal with the first on the expense of the second, it would be a net win?

I doubt I have the time to re-implement the PDF417 decoder but still wanted to ask for opinions on the matter, just in case.

axxel:
Addendum: I just found a one-line change that immediately makes the code about twice as fast in the "falsepositive" use case: simply don't look for the stop-guard-pattern if no start-guard-pattern has been found. If I do that, there are 4 samples that fail as a result. 3 of those are not valid PDF417 symbols because their start-guard is missing. There is a "compact PDF417" version available but only the stop-guard can be missing, not the start-guard.

The 4th is one of the [multi-symbol examples](https://github.com/nu-book/zxing-cpp/blob/master/test/samples/pdf417-4/04-01.png). That one is actually also invalid because it does not have the required row height of at least 3X.

I propose to remove the invalid samples and take that 2x speed improvement we get 'for free'.

huycn:
In our use cases, the change you propose would be a net win.

axxel:
To summarize #152: aside from better support for rotated symbols, also support vertically flipped symbols.

tailsu:
[...]

PR #1138, opened 2026-07-26
PDF417: match codewords using integer arithmetic

nyluke:
`GetClosestDecodedValue()` brute-force searches all 2787 symbols for the one whose bar width ratios are closest to the measured module bit counts. On a 10 MP photo of a PDF417 it was the hottest function in the whole read — about 44% of the time inside `ReadBarcodes()` in a sampling profile.

The error being minimised is

```
argmin_j sum_k (barSize[j][k] / 17 - moduleBitCount[k] / sum)^2
```

Multiplying by the positive constant `(17 * sum)^2` and dropping the `sum_k moduleBitCount[k]^2` term, which is the same for every symbol, gives an equivalent all-integer expression:

```
argmin_j (sum * sumOfSquares[j] - 34 * dot(barSize[j], moduleBitCount))
```

Same argmin, no division, no float rounding. The table drops from 87 kB of `float` to 22 kB of `int8_t`, small enough to stay in L1 for the whole scan, and the loop is short enough to vectorise without the early-exit branch.

### Results

`ReaderTest` on `test/samples`, Apple M3, Release build:

| | before | after |
|---|---|---|
| `-tpdf417` decode | 183 ms | **137 ms** |
| full suite decode | 748 ms | **697 ms** |

Medians over 7 and 5 runs respectively. The full-suite figure is diluted by the ~40 non-PDF417 sample sets; PDF417 itself is the −25%.

### Accuracy

Rather than rely on the expected-count assertions, I diffed the complete `ReaderTest` output (timings stripped) before and after: **identical on every sample set and rotation, 0 misreads**, including the `falsepositives-*` sets. All 199 unit tests pass.

```bash
./build/test/blackbox/ReaderTest test/samples 2>&1 | grep -vE "time:|time\." > after.txt
# ... rebuild on master ...
diff before.txt after.txt
```

Worth noting the integer form is strictly more precise than the float one, so the only place the two could ever disagree is a near-tie the float version got wrong.

### Notes

1. **The `#if 1` is gone.** It documented a deliberate 87 kB-on-heap vs 87 kB-in-`.rodata` tradeoff, with a note about a 7% smaller binary. At 22 kB I think `.rodata` is unambiguously the right side of that, so I dropped the runtime-initialised variant rather than carry both — should be reviewed.

2. **How does this sit with b802e2b?** That added the E2E-based `GetCodeword(ne2ep)` matcher, currently used by the isPure path and MicroPDF417, and "Preparing code re-use across all PDF417 symbologies" suggests the ratio matcher may be on its way out. This patch speeds up the path `PDFScanningDecoder.cpp` still uses for every normal read, so it helps today, but might become irrelevant.

### DIFF
--- a/core/src/pdf417/PDFCodewordDecoder.cpp
+++ b/core/src/pdf417/PDFCodewordDecoder.cpp
@@ -11,8 +11,10 @@
 
 #include <algorithm>
 #include <array>
+#include <cassert>
 #include <cstdint>
 #include <limits>
+#include <memory>
 
 namespace ZXing {
 namespace Pdf417 {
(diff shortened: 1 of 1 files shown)


core/src/pdf417/PDFCodewordDecoder.cpp
```diff
@@ -454,52 +454,56 @@ static constexpr int getSymbol(int idx)
...
+	//   argmin_j (sum * sumOfSquares[j] - 34 * dot(barSize[j], moduleBitCount))
+	// that picks the exact same symbol without any division or rounding.
+	int sum = Reduce(moduleBitCount);
+	ModuleBitCountType bitCounts = moduleBitCount;
+	if (sum <= 1) // the original code compares against all-zero ratios in this case
+		bitCounts.fill(0), sum = 1;
+
+	// worst case is bounded by 65535 * 17^2, so this comfortably stays inside int
+	int bestScore = std::numeric_limits<int>::max();
 	int bestMatch = -1;
```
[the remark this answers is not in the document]
axxel: that is not correct in this case. ignore.

core/src/pdf417/PDFCodewordDecoder.cpp
```diff
@@ -454,52 +454,56 @@ static constexpr int getSymbol(int idx)
...
+{
+	// Find the symbol whose bar width ratios are closest to the measured ones, i.e.
+	//   argmin_j sum_k (barSize[j][k] / 17 - moduleBitCount[k] / sum)^2
+	// Scaling that by the positive constant (17 * sum)^2 and dropping the sum_k moduleBitCount[k]^2
+	// term (which is the same for every symbol) leaves an equivalent all-integer expression
+	//   argmin_j (sum * sumOfSquares[j] - 34 * dot(barSize[j], moduleBitCount))
+	// that picks the exact same symbol without any division or rounding.
+	int sum = Reduce(moduleBitCount);
+	ModuleBitCountType bitCounts = moduleBitCount;
+	if (sum <= 1) // the original code compares against all-zero ratios in this case
```
axxel: This check can safely turned into an `assert(sum>=8)` because each bar/space needs to be at least 1 pixel wide per definition.

core/src/pdf417/PDFCodewordDecoder.cpp
```diff
@@ -454,52 +454,56 @@ static constexpr int getSymbol(int idx)
...
-				}
-				currentBit = currentSymbol & 0x1;
-				table[i][CodewordDecoder::BARS_IN_MODULE - j - 1] = size / 17.f; // MODULES_IN_CODEWORD
-			}
-		}
-		return table;
-	}();
+	std::array<std::array<int8_t, CodewordDecoder::BARS_IN_MODULE>, SYMBOL_COUNT> sizes;
+	std::array<int16_t, SYMBOL_COUNT> sumOfSquares; // max is 17^2, so int16_t is plenty
+};
```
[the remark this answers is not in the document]
axxel: That is a valid point and it means that the size of `sumOfSquares` can actually be cut in half (`int_8` instead of `int_16`).

core/src/pdf417/PDFCodewordDecoder.cpp
```diff
@@ -454,52 +454,56 @@ static constexpr int getSymbol(int idx)
 	return SYMBOL_TABLE[idx] | 0x10000;
 }
 
-static int GetClosestDecodedValue(const ModuleBitCountType& moduleBitCount)
+// The bar widths (in modules, 1..6) of every symbol, plus the sum of their squares. 22kB of int8_t
```
axxel: That misses the `SYMBOL_COUNT * sizeof(int_16)` bytes in `sumOfSquares` (or half that, see below). So the total size is 27kB (with the below optimization 24kB).

core/src/pdf417/PDFCodewordDecoder.cpp
```diff
@@ -454,52 +454,56 @@ static constexpr int getSymbol(int idx)
...
+// in .rodata instead of 87kB of float on the heap, which keeps the whole table in L1 during the scan.
+struct BarSizeTable
 {
-#if 1 // put 87kB on heap and calculate per process on first use -> 7% smaller binary
```
axxel, line 459: I'd prefer if you kept the `#if` switch, if only for documentation at the moment. Makes it a lot easier to experiment with the two options.

axxel (changes requested):
Thanks a lot for this substantial improvement. It is also very refreshing to get a non-trivial PR that someone actually crafted by hand these days. ;).

Yes, this code path might go away completely, but who knows when (see the age of the #145).

I would appreciate if you could incorporate the mentioned changes. If you don't have the time, I'll merge it as is and fix them later myself.

core/src/pdf417/PDFCodewordDecoder.cpp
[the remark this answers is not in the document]
nyluke: confirmed

core/src/pdf417/PDFCodewordDecoder.cpp
nyluke: Restored.. I used make_unique rather than the original std::vector because the table is now a struct of two fixed arrays - a plain static const would get constant-initialised into .rodata and the branch would silently not do what the comment says. The heap variant is 32 kB / 2.2% smaller, which is where the comment's figure comes from.

nyluke:
Changes re-tested and pushed 🙌

axxel:
Awesome. Anything else you care about looking into? :)

May I ask what your use case is? Are you specifically interested in PDF417 detection on Apple silicon?

nyluke:
Specifically PDF417 yes, for identity documents. Not necessarily Apple Silicon but I've been targeting ARM64. I have an optimization for JPG decoding that I'll submit also. Which areas are critical to look into outside of this pipeline?

axxel:
I'll try to remember that you care about PDF417. My recent addition of the MicroPDF417 code is a basis for finally looking into a PDF417 reimplementation and when the time comes, I'll be interested in real-world feedback...

Looking forward to your JPEG optimization (since the library itself does not deal in files or image formats, I assume you are talking about the `ZXingReader` example code?).

> Which areas are critical to look into outside of this pipeline?

What do you mean with 'pipeline' in this context? The library is in a pretty good shape overall. I spent _a lot_ of time optimizing the 'detection' part (meaning finding a symbol quickly and especially quickly ruling out the presence of a symbol) because that is what you pay for in every single frame. The 'decoding' part (what you just improved can be put in that box) may have some low hanging fruit left but it is generally cheaper than the detection anyway.

Issue #1098, opened 2026-04-13
UPC-E/A: 8 digits vs 12 vs 13 digits


janpatera:
The version 2.3.0 of the library used to return 8 digits for UPC-E codes (e.g. 00223393) and 12 digits for UPC-A. However, 3.0.2 now returns 13 in both cases. Could this new behavior be made optional, please? Although stripping the leading 0 is trivial, it requires some work to convert back to 8 digits.
Rationale: to check that the artwork contains the code it should contain, the decoded value should match what the artwork creator used to enter to create the barcode as much as possible.

axxel:
This behavior will not be optional. I consider the change to be a fix, since the old behavior was violating the ISO specification. The 12 digit UPC-A behavior was never more than a crude heuristic simply dropping the leading zero. That can trivially be implemented in client code if so desired. The original UPC-E text is still available via the new `Barcode::extra("UPCE")` property.

PR #1148, opened 2026-08-10
iOS: expose Barcode::extra() metadata as a dictionary on ZXIResult

bscanlon-kr:
In this [issue](https://github.com/zxing-cpp/zxing-cpp/issues/1098) its called out that to get access to the raw scan data for the UPC-E scan the extras need to be referenced. The extras aren't exposed in the iOS wrapper. (I suspect this is the case for many of the wrappers) this PR adds the ability to pull from the extras map.

### DIFF
--- a/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
+++ b/wrappers/ios/Sources/Wrapper/Reader/ZXIBarcodeReader.mm
@@ -39,10 +39,14 @@ NSDictionary<NSString *, id> *getExtra(const Barcode &barcode) {
     if (json.empty()) {
         return @{};
     }
-    NSData *data = [stringToNSString(json) dataUsingEncoding:NSUTF8StringEncoding];
+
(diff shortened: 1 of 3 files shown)


axxel:
As with the Android version, it seems there is a more appropriate type available in ObjectiveC as well (says Copilot):

A better default would usually be: `NSDictionary<NSString *, id> *extra`

Why: NSJSONSerialization can produce values that are not just NSString: NSNumber, NSArray, NSDictionary, NSNull

bscanlon-kr:
Yeah was waiting for us to work out the details on the Android one. I'll get these changes going.

axxel:
Can you please drop the last commit (and force push the branch). I don't find this helpful at all. Since the client code is not supposed to directly instantiate a `ZXIResult`, there is not point in having this initializer overload. Sorry for not making that clear right after Copilot suggested it.

Other than that, the PR is good to merge, right? (@alexmanzer, @markusfisch)

Issue #1131, opened 2026-07-10
Reducing binary size


Bixilon:
Hi there,
just did a fresh build of the android wrapper and my stripped `libzxingcpp_android.so` (arm64-v8a) has about 1.8MB.  My total app has about 4.7MB now :). My use case is very simple: scan a qr code. I've seen https://github.com/zxing-cpp/zxing-cpp/discussions/715 and writing is already off (https://github.com/zxing-cpp/zxing-cpp/blob/master/wrappers/android/zxingcpp/build.gradle.kts#L31).

Disabling other qr code formats is probably not an option, as I do not want to build the library myself (in the app), so are there any other optimizations that can be done?

Thanks

axxel:
First question: was this substantially different with earlier versions, meaning this is a regression? 1.8MB sounds a lot to me. The plain (non android) so file on my M4 (arm) is around 1MB.

Bixilon:
No, not *substantially*, in v3.0.2 it still had 1.6MB. but in https://github.com/zxing-cpp/zxing-cpp/discussions/715#discussioncomment-8385804 you pointed out that it might be compiler related. I am not even sure what compiler it used for my local build.

Bixilon:
I did some tests, but not as promising yet.

- Disabling all other formats (except QR): 1.8M -> 1.3M
- Replacing `libzueci` with noop (just for testing): 1.3M -> 1.2M
- Changing build type from `RelWithDebInfo` to `Release`: Almost no difference

axxel:
I did some research:
* a standard machO dylib (macOS so file) in the equivalent setup is 1M (instead of 1.8M)
* the biggest difference is the statically linked libstdc++ (`"-DANDROID_STL=c++_shared"`): approx. 800k
* using `MinSizeRel` instead of `Release` can strip a few kB but makes it noticeably slower (like 20% on my Pixel 5)
* hiding all symbols by default helps and should have been the default anyway: `set(CMAKE_CXX_VISIBILITY_PRESET hidden)`

The absolute minimum `libzxingcpp_android.so` file size I was able to achieve was 260k (270k in the `Release` config). That was with only QRCode support and a disabled `libzueci`. But that requires the packaging of the libstdc++ in the APK, which would most likely only increase the total size. With a statically linked libstcd++ I got 1.1M as supposed to your 1.2M, which should be the hidden symbols.

Let me know if using `c++_shared` does actually help your use case (in contrast to my assumption).

[...]

PR #1151, opened 2026-08-14
Reduce the Android binary size by 43% (1,747 KB → 988 KB)

tim4724:
Follow-up to #1131, which closed with `CMAKE_CXX_VISIBILITY_PRESET hidden` in place and the
question of "some more easy size optimizations that could be done without reducing functionality"
still open. There are two, and neither changes the API or any decode result.

All figures: `libzxingcpp_android.so`, arm64-v8a, stripped, built with the wrapper's own settings
(NDK 27.2.12479018, `RelWithDebInfo`, `ZXING_WRITERS=OFF`). The tables below are measured against
the v3.1.1 release; on current master (a17fd9d) the same change is 1,747,936 → 989,704, the same
−43.4%.

| | size | vs v3.1.1 |
|---|---|---|
| v3.1.1 | 1,747,000 | — |
| commit 1 alone (drop `std::locale`) | 1,191,376 | −555,624 |
| commit 2 alone (`--exclude-libs,ALL`) | 1,199,144 | −547,856 |
| **both (this PR)** | **988,280** | **−758,720 (−43.4%)** |

The two overlap by 344,760 bytes, because they remove much of the same libc++ code by different
routes — one stops referencing it, the other lets the linker collect it. Neither subsumes the
other: together they beat the better of the two by another 203 KB.

## 1. Three `ostringstream`s pin `std::locale`

`ToSVG()`, `EscapeNonGraphical()` and the PDF417 macro-block builder each use a
`std::ostringstream` purely to concatenate a few strings and integers. Instantiating any stream
pulls in the `std::locale`/`num_put` machinery, and it is genuinely reachable from the reader
entry points, so no linker flag can remove it — it has to go from the source. Replaced with plain
`std::string`/`std::wstring` appends.

It is all-or-nothing, which may be why it hasn't surfaced before. Removing the sites one at a
time (with commit 2 already applied):

| | size | Δ |
|---|---|---|
| `--exclude-libs,ALL` only | 1,199,144 | — |
| + `Utf.cpp` | 1,190,856 | −8,288 |
| + `BitMatrixIO.cpp` | 1,190,856 | 0 |
| + `pdf417/PDFDecoder.cpp` | 988,280 | −202,576 |

Anyone who converted one site, measured, and saw 8 KB would reasonably have stopped.

`AppendHex()` preserves the `std::setw()` semantics it replaces, including *not* truncating a
value needing more digits than the requested width — `<U+10FFFF>` still round-trips.

## 2. The visibility preset doesn't reach the statically linked libc++

`CMAKE_CXX_VISIBILITY_PRESET` applies to code compiled in this project, and it does its job: of
the 2031 symbols the published `.so` exports, **none** are ZXing's own C++ API. But symbols from
static archives keep default visibility and get re-exported:

```
$ llvm-nm --dynamic --defined-only libzxingcpp_android.so | wc -l
2031
```

which breaks down as 1812 `std::` symbols, 177 `typeinfo`/`vtable` for libc++ types, 32
`__cxa_*`/`__gxx_personality_v0`/`__dynamic_cast`, 6 `operator new`/`delete`, 2 `zueci_*` from
bundled libzueci — and 2 `Java_zxingcpp_*`, the only ones actually wanted.

The cost isn't the symbol table. The NDK links release configs with `--gc-sections`
(`build/cmake/android-legacy.toolchain.cmake:507` for `RelWithDebInfo`, which
`android.toolchain.cmake:55` includes by default), and **every exported symbol is a GC root** — so
the re-exported libc++ locale, iostream and demangler code can't be collected even though nothing
calls it. `--exclude-libs,ALL` hides it and lets the linker drop it. Section sizes, applying the
flag to v3.1.1 on its own:

```
.text     989,096 → 757,672
.dynstr   104,235 →   1,137
.dynsym    51,936 →   2,016
.rela.dyn  93,888 →  64,320
```

## Verification

**Correctness**

- **Blackbox** `ReaderTest` over `test/samples`: output byte-identical to v3.1.1 for every format
  (diffed with timings stripped).
- I ran this repo's own CI jobs locally on Ubuntu 24.04 in Docker, using each job's exact
  configure line: `build` (its C++ portions), `build-ubuntu-gcc-12`, `build-ubuntu-asan-ubsan`,
  `build-ubuntu-tsan` and `build-old-writers`. All five build with **no new warnings in the
  changed files**, and every `ctest` suite passes. The ASan/UBSan job is the relevant one for
  `AppendHex()`, which indexes a local buffer directly.
- Also builds clean on macOS/AppleClang across `ZXING_WRITERS` = `OFF`/`OLD`/`NEW`/`BOTH` and with
  `ZXING_READERS=OFF`.

  (Caveats on the local CI run: the containers are arm64 rather than the runners' x86_64, and I
  skipped the .NET/Go/Python wrapper tests, which need those toolchains and which this change
  doesn't touch.)
- `ToSVG()` has no unit test and `EscapeNonGraphical()` has three assertions, so I also built a
  differential harness linking the old and new implementations and compared them over **all
  1,112,064 non-surrogate Unicode code points**, plus 500 mixed strings and 40 matrices: identical
  SHA-256.

**Performance** — no regression.

- Decode throughput (blackbox `ReaderTest` over all samples, 7 runs each): median 800 ms → 791 ms,
  −1.1%, i.e. within noise. None of the three call sites is on the decode hot path.
- The rewritten functions are themselves faster: the differential harness above, whose dominant
  cost is 1.1M `EscapeNonGraphical()` calls, drops from 0.244 s to 0.168 s.
- Commit 2 cannot affect performance by construction: it is link-only, and all 79 ZXing object
  files are byte-identical before and after it (compared with debug info stripped, since
  `RelWithDebInfo` embeds the build path).

## Deliberately not included

Each of these was measured on top of this PR, and left out for the reason given. Happy to add any
of them if you'd prefer:

- **Thin LTO** — −25,840, but it lengthens link times.
- **`--icf=all`** — −21,696, but it can merge functions with identical bodies, which breaks
  function-pointer identity. `--icf=safe` is 0 bytes here, so there's no safe version of this one.
- **`--pack-dyn-relocs=android`** — −41,600, but APS2 needs API 23+ and this wrapper is
  `minSdk 21`.
- **Per-format `ZXING_ENABLE_*`** — by far the largest lever (QR-only is −391,528), but it can't
  apply to a prebuilt AAR that has to serve every consumer.

This is all specific to Android, where the NDK links libc++ statically into every shared object. A
`BUILD_SHARED_LIBS` build on Linux or macOS links libc++ dynamically and doesn't have the problem:
`libZXing.dylib` there exports 584 symbols, 480 of them ZXing's own.

## Disclosure

This was written with Claude (Anthropic's Claude Code). Every figure above is a measurement from a
local build rather than a model estimate, and the verification section lists exactly what was run.

### DIFF
--- a/core/src/BitMatrixIO.cpp
+++ b/core/src/BitMatrixIO.cpp
@@ -8,7 +8,6 @@
 
 #include <array>
 #include <fstream>
-#include <sstream>
 
 namespace ZXing {
 
@@ -54,20 +53,28 @@ std::string ToSVG(const BitMatrix& matrix)
 
 	const int width = matrix.width();
 	const int height = matrix.height();
-	std::ostringstream out;
+	std::string out;
 
-	out << R"(<?xml version="1.0" encoding="UTF-8"?>)" << "\n"
-		<< R"(<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 )" << width << ' ' << height << R"(" stroke="none">)"
-		<< "\n<path d=\"";
+	out += R"(<?xml version="1.0" encoding="UTF-8"?>)" "\n"
+		   R"(<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 )";
+	out += std::to_string(width);
+	out += ' ';
+	out += std::to_string(height);
+	out += R"(" stroke="none">)" "\n<path d=\"";
 
 	for (int y = 0; y < height; ++y)
 		for (int x = 0; x < width; ++x)
-			if (matrix.get(x, y))
-				out << "M" << x << "," << y << "h1v1h-1z";
+			if (matrix.get(x, y)) {
+				out += 'M';
+				out += std::to_string(x);
+				out += ',';
+				out += std::to_string(y);
+				out += "h1v1h-1z";
+			}
 
-	out << "\"/>\n</svg>";
+	out += "\"/>\n</svg>";
 
-	return out.str();
+	return out;
 }
 
 BitMatrix ParseBitMatrix(const std::string& str, char one, bool expectSpace)
--- a/core/src/BitMatrixIO.cpp
+++ b/core/src/BitMatrixIO.cpp
@@ -6,6 +6,8 @@
 
 #include "BitMatrixIO.h"
 
+#include "ZXAlgorithms.h"
+
 #include <array>
 #include <fstream>
 
@@ -53,14 +55,10 @@ std::string ToSVG(const BitMatrix& matrix)
 
 	const int width = matrix.width();
 	const int height = matrix.height();
-	std::string out;
-
-	out += R"(<?xml version="1.0" encoding="UTF-8"?>)" "\n"
-		   R"(<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 )";
-	out += std::to_string(width);
-	out += ' ';
-	out += std::to_string(height);
-	out += R"(" stroke="none">)" "\n<path d=\"";
+	std::string out = StrCat(R"(<?xml version="1.0" encoding="UTF-8"?>)" "\n"
+							 R"(<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 )",
+							 std::to_string(width), ' ', std::to_string(height),
+							 R"(" stroke="none">)" "\n<path d=\"");
 
 	for (int y = 0; y < height; ++y)
 		for (int x = 0; x < width; ++x)
--- a/core/src/Utf.cpp
+++ b/core/src/Utf.cpp
@@ -233,16 +233,10 @@ static bool iswgraph(wchar_t wc)
 	return true;
 }
 
-// Appends the ASCII string s. Only used with literals and the ascii_nongraphs table below.
-static void AppendAscii(std::wstring& out, const char* s)
-{
-	while (*s)
-		out += static_cast<wchar_t>(*s++);
-}
-
-// Appends val as uppercase hex, zero-padded to len digits. Like the std::setw() it replaces,
-// a value too large for len is not truncated but written in full.
-static void AppendHex(std::wstring& out, uint32_t val, int len)
+// Appends "<U+XXXX>", val in uppercase hex zero-padded to len digits. This is std::format("<U+{:0{}X}>")
+// spelled out by hand: on Android, instantiating std::format at all links the libc++ locale facets and
+// costs ~395 KB, which is most of what dropping the ostringstreams here bought (see #1151).
+static void AppendUnicodeEscape(std::wstring& out, uint32_t val, int len)
 {
 	wchar_t buf[8]; // enough for any uint32_t
 	int n = 0;
@@ -250,20 +244,23 @@ static void AppendHex(std::wstring& out, uint32_t val, int len)
 		buf[n++] = L"0123456789ABCDEF"[val & 0xf];
 		val >>= 4;
 	} while (val);
-	for (; n < len; ++n)
+	for (; n < len; ++n) // like std::format's width, a value needing more digits than len is not truncated
 		buf[n] = L'0';
+
+	out += L"<U+";
 	while (n--)
 		out += buf[n];
+	out += L'>';
 }
 
 std::wstring EscapeNonGraphical(std::wstring_view str)
 {
-	static const char* const ascii_nongraphs[33] = {
-		"NUL", "SOH", "STX", "ETX", "EOT", "ENQ", "ACK", "BEL",
-		"BS",  "HT",  "LF",  "VT",  "FF",  "CR",  "SO",  "SI",
-		"DLE", "DC1", "DC2", "DC3", "DC4", "NAK", "SYN", "ETB",
-		"CAN",  "EM", "SUB", "ESC",  "FS",  "GS",  "RS",  "US",
-		"DEL",
+	static const wchar_t* const ascii_nongraphs[33] = {
+		L"NUL", L"SOH", L"STX", L"ETX", L"EOT", L"ENQ", L"ACK", L"BEL",
+		L"BS",  L"HT",  L"LF",  L"VT",  L"FF",  L"CR",  L"SO",  L"SI",
+		L"DLE", L"DC1", L"DC2", L"DC3", L"DC4", L"NAK", L"SYN", L"ETB",
+		L"CAN",  L"EM", L"SUB", L"ESC",  L"FS",  L"GS",  L"RS",  L"US",
+		L"DEL",
 	};
 
 	std::wstring ws;
@@ -271,9 +268,9 @@ std::wstring EscapeNonGraphical(std::wstring_view str)
 	for (; str.size(); str.remove_prefix(1)) {
 		wchar_t wc = str.front();
 		if (wc < 32 || wc == 127) { // Non-graphical ASCII, excluding space
-			AppendAscii(ws, "<");
-			AppendAscii(ws, ascii_nongraphs[wc == 127 ? 32 : wc]);
-			AppendAscii(ws, ">");
+			ws += L'<';
+			ws += ascii_nongraphs[wc == 127 ? 32 : wc];
+			ws += L'>';
 		} else if (wc < 128) // ASCII
 			ws += wc;
 		else if (IsUtf16SurrogatePair(str)) {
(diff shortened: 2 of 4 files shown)


axxel:
Thanks a lot for this very insightful analysis. I am interested in merging this but would like to suggest an improvement: replace the raw string manipulation including the `AppendHex` and `AppendAscii` helper with the use of plain `std::format`, which is also locale independent and would have been the go-to solution if would have been available when the respective code was written. It might increase the total size again by a few kB but that would be acceptable to me given the net benefit would still be on the order of hundreds of kB.

tim4724:
Thanks. I tried it, and `std::format` turns out to be the one replacement that can't work here. Numbers first, `libzxingcpp_android.so` arm64-v8a stripped, same setup as the PR description:

| | size | vs this PR |
|---|---|---|
| this PR (before your review) | 989,248 | baseline |
| `std::format` for the hex escape only | 1,384,568 | **+395,320** |
| `std::format` for the `<{}>` escape only | 1,384,856 | **+395,608** |
| `std::format` for both | 1,384,120 | **+394,872** |

It's ~395 KB, not a few KB, and it's a flat one-time cost: one call site or two makes no difference, because it isn't the *formatter* that's expensive, it's `std::format` existing at all in reachable code.

You're right that `std::format` is locale-independent *semantically*. The problem is what libc++ has to instantiate to get there. `__handle_replacement_field` visits every alternative of `basic_format_arg`, so formatting a single `const wchar_t*` still instantiates the `long double` path, which on libc++ bottoms out in `__libcpp_snprintf_l`. Diffing the symbol tables, the `std::format` build gains 950 symbols the current one doesn't have, including:

```
std::locale::classic()::classic_locale
std::locale::__global()::g
std::__num_put<char>::__widen_and_group_float(...)
std::__num_get<wchar_t>::__stage2_float_loop(...)
std::__money_put<wchar_t>::__gather_info(...)
std::__libcpp_asprintf_l / __libcpp_snprintf_l / __libcpp_sscanf_l
```

That is precisely the `std::locale` machinery commit 1 exists to unlink, and once it's referenced again `--exclude-libs,ALL` can't collect it either, so commit 2's win partly goes with it. `.text` goes 615,568 → 814,328.

Worth knowing separately: NDK 27's libc++ (LLVM 18) doesn't define `__cpp_lib_format` at all, so `GTIN.cpp` and `JSON.h` currently compile to their `#else` branches on Android. And `EanAddOn`'s price formatting is unreachable from the JNI entry points, so it's dead-stripped anyway. I checked by making it unconditional and the size didn't move. So nothing in the tree pays this cost today; this rewrite would have been the first.

**What I changed instead.** I do take the readability point, so I pushed a commit that removes most of the hand-rolled feel without the size:

- `AppendAscii` is gone entirely. It only existed because `ascii_nongraphs` held narrow literals, and making the table `const wchar_t*` lets `operator+=` take them directly.
- `AppendHex` becomes `AppendUnicodeEscape`, emitting the whole `<U+XXXX>` token, so the call site is one line instead of three. Its comment names the `std::format("<U+{:0{}X}>")` it stands in for and why it isn't that.
- `ToSVG`'s header now uses the existing `StrCat` rather than an append chain.

Net −256 bytes (989,248 → 988,992), and eight fewer lines than the version you reviewed.

Verification for the new commit: differential tests against the pre-PR `ostringstream` implementations give byte-identical output over all 1,112,064 non-surrogate code points, 20,000 mixed strings and 302 random matrices including the degenerate sizes, clean under ASan/UBSan; the full `ctest` suite passes with `ZXING_WRITERS=BOTH`, no new warnings.

If you'd still rather have `std::format` and treat the 395 KB as acceptable, say so and I'll switch it. It would need a `__cpp_lib_format` guard with the manual path kept as the `#else` for gcc 12, though, which means both spellings stay in the file and Android silently gets the 395 KB the day an NDK defines the macro. That seemed like the worst of both, so I went the other way.

axxel:
Excellent analysis and justification. 395 is _a lot_ more than 'a few', so I totally agree with you on this.

Bixilon:
@tim4724 many thanks! While I still think 1MB is a lot for this library, its a lot better than before!

tim4724:
@Bixilon Agreed — and the lever that closes most of the rest of that gap is the per-format one you
measured in #1131. It can't ship in the AAR, which has to serve every consumer, but a consumer app
can take it without vendoring or forking anything: the wrapper's Kotlin and JNI compile straight
out of the repo as a source set. arm64-v8a, stripped:

| | size |
|---|---|
| 3.1.1 AAR | 1,747,000 |
| master (this PR) | ~989,000 |
| + QR-only, thin LTO, APS2 relocations | **540,928** |

armeabi-v7a is 405,192. That's −69% against the AAR, and it's what our app ships today.

All it needs is a checkout of this repo somewhere in your tree — a git submodule is one way, and
the one we use, since it pins the version and a bump is then just moving the pointer:

```sh
git submodule add https://github.com/zxing-cpp/zxing-cpp third_party/zxing-cpp
```

plus this in an app module's `build.gradle.kts`, with nothing copied into your sources:

```kotlin
val wrapper = rootProject.file("../third_party/zxing-cpp/wrappers/android/zxingcpp/src/main")

android {
    ndkVersion = "28.2.13676358"   // pin it: the NDK decides what lands in the .so
    defaultConfig.externalNativeBuild.cmake {
        arguments(
            "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
            "-DANDROID_ARM_NEON=ON",
            "-DZXING_ENABLE_1D=OFF", "-DZXING_ENABLE_AZTEC=OFF", "-DZXING_ENABLE_DATAMATRIX=OFF",
            "-DZXING_ENABLE_MAXICODE=OFF", "-DZXING_ENABLE_PDF417=OFF",
            "-DCMAKE_C_FLAGS=-flto=thin", "-DCMAKE_CXX_FLAGS=-flto=thin",
            // APS2 needs API 23+, which is why the wrapper (minSdk 21) can't do this itself
            "-DCMAKE_SHARED_LINKER_FLAGS=-flto=thin -Wl,--pack-dyn-relocs=android",
        )
    }
    sourceSets["main"].kotlin.srcDir("$wrapper/java")
    externalNativeBuild.cmake.path = file("$wrapper/cpp/CMakeLists.txt")
}
```

**One trap, because it fails silently.** The AAR carries `-keep class zxingcpp.** { *; }` in its
`consumer-rules.pro` and AGP merges that for you. Compiling the wrapper as a source set drops the
rule, and the JNI resolves the Kotlin side entirely by name. In our minified release R8 cut 11
classes to 2 and renamed every field the JNI looks up — the app built, the camera ran, and the
decoder returned nothing. Invisible in debug builds and in instrumented tests, since neither is
minified. Copy the rule into your own `proguard-rules.pro`.

Two smaller things the Maven artifact was also doing: `camera-core` arrives transitively through
its POM, and license/attribution tooling that reads the dependency graph stops seeing zxing-cpp
while its Apache-2.0 code keeps shipping.

Full working setup: tim4724/CouchPad-Controller@c7ab7c6 (its numbers predate this PR's last commit;
the submodule now tracks master).

Bixilon:
@tim4724 I am building a slightly modified version, but its still 463k (arm64-v8a) with only QR. I am fully aware that the the other formats can not be stripped away.

I am using camerax, its not that large (appcompat excluded). I am targetting api 26+.

The proguard rules don't bother me at all, its just 23k, and mostly required because of JNI. If you remove the rule it will break. I still think that ~500k is a lot, just for the QR detection part, the whole apk has barely 3.8M (with zxing-cpp), well that makes about 12%. See https://gitlab.bixilon.de/bixilon/unithen.


Anyways, I am more than happy with the reduction in size.

axxel:
@tim4724

1. Would you be interested in adding a "Show and tell" discussion entry, repeating what you described [here](https://github.com/zxing-cpp/zxing-cpp/pull/1151#issuecomment-5330519942)? I would then most likely add a reference to the wrapper README as well. This should substantially improve the discoverability of that most useful information.
2. Are you done with your campaign to reduce the binary size? There might be some more opportunities around the different pattern/character lookup tables with potentially the option to reduce binary size for the cost of runtime memory size + some CPU cost for the initialization. Asking to get a feel regarding the timing of the next release... ;)
3. Have you considered/experimented with the removal of the libzueci code (see [here](https://github.com/zxing-cpp/zxing-cpp/commit/b2ca864d5b5c9a116cceac6f109539c77eebd77f))? This would save you another ~100k if your QRCodes only contain simple ASCII content. If it is helpful, I can easily add a cmake build option to toggle this.
4. Maybe it would be worth it, to add a separate AAR target, that only decodes QRCodes? It seems quite likely that actually most Android consumer app use cases are solely interested in QRCode scanning. I expect to bifurcate the target anyway as soon as writer API is added to the wrapper as this would roughly double with size and have even less active users...

Bixilon:
@axxel 

> Have you considered/experimented with the removal of the libzueci code (see [here](https://github.com/zxing-cpp/zxing-cpp/commit/b2ca864d5b5c9a116cceac6f109539c77eebd77f))? This would save you another ~100k if your QRCodes only contain simple ASCII content. If it is helpful, I can easily add a cmake build option to toggle this.

I am building my custom library without it, are there any grave consequences? Like crashing with special qr codes or is just the data wrong (also not ideal), more like ignoring them or so?

> Maybe it would be worth it, to add a separate AAR target, that only decodes QRCodes? It seems quite likely that actually most Android consumer app use cases are solely interested in QRCode scanning. I expect to bifurcate the target anyway as soon as writer API is added to the wrapper as this would roughly double with size and have even less active users...

I'd be in favor of that solution. I am only using qr codes and probably most people too. For writing I still use original zxing, and on Android it is pretty small (probably due to tree shaking R8 optimization). The whole jar only has 596k (with all formats). But I don't have any exact numbers.

axxel:
@tim4724 @Bixilon: Just to make sure you have seen it... Please have a look at #1155.

PR #1158, opened 2026-09-07
Build docs only for a top-level build

Restia-Ashbell:
docs/ was added unconditionally, forcing projects that consume zxing-cpp via add_subdirectory()/FetchContent() to fetch doxygen-awesome-css at configure time. Only configure docs when PROJECT_IS_TOP_LEVEL, bumping the minimum CMake version to 3.21 and the README accordingly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.21)
 
 project(ZXing)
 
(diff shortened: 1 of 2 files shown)


axxel:
Thanks for the PR. I agree with Copilot here in the sense that I would not want to include this change in the imminent 3.1.2 patch release. I have two options for you: a) leave the PR as is and I will merge it sometime later or b) change it to use the suggested `STREQUAL` solution (adding a TODO comment to replace that with `PROJECT_IS_TOP_LEVEL` in cmake 3.21).

axxel:
I'm considering just going directly to 3.2, so probably not worth amending your PR.

